PR #3, opened 2015-11-26
Add tagging interface

tijmenb:
This PR is a first attempt at a tagging interface to replace the panopticon interface. Doing this will allow us to use the new tagging architecture.

We will be migrating applications one-by-one. The first one will be smart answers. This interface will not be used for publisher apps that have their own tagging interface.

## Happy path

User wants to tag a page on GOV.UK. This form accepts `/paths` and `https://www.gov.uk/full-gov.uk/urls`.

screen shot 2015-11-26 at 14 19 39

Because this is a smart answers, it can be tagged to mainstream browse pages, topics and organisations.

screen shot 2015-11-26 at 14 19 48

We use the default `select2` dropdowns for these.

screen shot 2015-11-26 at 14 20 03

Clicking "Update taggings" will save the links.

screen shot 2015-11-26 at 14 20 09

## URL doesn't exist

screen shot 2015-11-26 at 14 37 07

## Conflicting edit

When someone else makes a change after you've loaded the page, or the user makes an edit in an other browse tab, the change might conflict:

screen shot 2015-11-26 at 14 38 53

Trello: https://trello.com/c/XwVC0IM8


bishboria:
:+1:

PR #5, opened 2015-12-02
Get content-item basepath from url, not a form field

bishboria:
Save some users from having to enter a basepath via the form submission, and allow them to enter the basepath in the URL, eg `content-tagger.gov.uk/lookup/some/basepath`

### DIFF
--- a/app/controllers/lookup_controller.rb
+++ b/app/controllers/lookup_controller.rb
@@ -4,7 +4,7 @@ class LookupController < ApplicationController
   end
 
   def find_by_slug
-    content_lookup = ContentLookupForm.new(params[:content_lookup_form])
+    content_lookup = ContentLookupForm.new(lookup_params)
 
     if content_lookup.valid?
       redirect_to content_path(content_lookup.content_id)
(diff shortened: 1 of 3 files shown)


app/controllers/content_controller.rb
```diff
@@ -10,4 +10,9 @@ def update_links
   rescue GdsApi::HTTPConflict
     redirect_to :back, danger: "Somebody changed the tags before you could. Your changes have not been saved."
   end
+
+  def slug
```
tijmenb: Would this be better in the `LookupController`? We could re-use the `find_by_slug` method.

bishboria: probably would be better :)

app/controllers/lookup_controller.rb
```diff
@@ -13,4 +13,10 @@ def find_by_slug
       render 'new'
     end
   end
+
+private
+
+  def base_path
```
tijmenb: Would `lookup_params` be a better name?

tijmenb:
:+1:

PR #7, opened 2015-12-02
Rename initializer for admin template

tijmenb:
The file `config/initializers/govuk_admin_template.rb` is overwritten on deploy so display the environment indicators. In our case, it was overwriting our own initializer, causing the flash, app title and
logout-button to disappear.


jamiecobbett:
Would it be lots of effort to use environment variables? Then both files could live in this repo.

tijmenb:
Closing because it's better to rename the file that is being created by capistrano.

PR #8, opened 2015-12-03
Add download and parsing logic for the alpha taxonomy

mobaig:
The alpha taxonomy is a work-in-progress taxonomical structure currently
being put together by GOV.UK information architects. In order to make
the taxonomy data available for prototyping of the new tagging interface,
a data pipeline is required to get taxons into the content store and link
to them from specific piece of content.

These tasks handle the downloading and parsing of the remotely-stored
taxonomy mappings. These will be followed by additional tasks that will
handle interactions with the Publishing API and Content Store.

Trello: https://trello.com/c/bsZAwZ35/450-improve-alpha-taxonomy-tagging


lib/alpha_taxonomy/sheet_parser.rb
```diff
@@ -0,0 +1,51 @@
...
+        tsv_data.values_at(*desired_columns)
+      else
+        raise ArgumentError, "Column names did not match expected values #{desired_columns}"
+      end
+    end
+
+    # Standardise the appearance of taxonomy labels extracted from the spreadsheets.
+    # e.g. -
+    # "3 Early Years Foundation Stage (EYFS)" > "3-early-years-foundation-stage-eyfs"
+    def sluggify(taxonomy_label)
```
tijmenb: Do we still need to do this here?

mobaig: as discussed, no - updated to simply strip whitespace and expect pipe-separator.

PR #9, opened 2015-12-04
Fix kind of content available for 'parent'

tijmenb:
The "parent" tag is the primary mainstream browse pages, so the select box should be populated with those, not the topics.

https://trello.com/c/XwVC0IM8

### DIFF
--- a/app/views/content/_form_for_parent.html.erb
+++ b/app/views/content/_form_for_parent.html.erb
@@ -1,6 +1,6 @@
 <h3>Parent</h3>
 
 <%= f.select :parent,
-    options_for_select(TagFetcher.new.tags_of_type('topic'), @tagging_update.parent),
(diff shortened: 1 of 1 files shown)


app/views/content/_form_for_parent.html.erb
```diff
@@ -1,6 +1,6 @@
 <h3>Parent</h3>
 
 <%= f.select :parent,
-    options_for_select(TagFetcher.new.tags_of_type('topic'), @tagging_update.parent),
+    options_for_select(TagFetcher.new.tags_of_type('mainstream_browse_page'), @tagging_update.parent),
```
mobaig: this should go in a helper or an instance var in the controller.

tijmenb: Why?

mobaig: yeah cool, fair reasoning.

mobaig:
LGTM

PR #11, opened 2015-12-08
Extract linkset into object

tijmenb:
This means that even if the API returns nil (if the linkset doesn't exist) we return an empty LinkSet object that responds to the appropriate methods.


mobaig:
lgtm

PR #13, opened 2015-12-14
Add functionality to create taxons from an import spreadsheet

mobaig:
Add functionality to create taxons from an import spreadsheet.

Alpha-stage taxonomy mappings are currently being created by Information Architects on the Finding Things team. Ultimately, we want to be able to curate and manage these taxons and their mappings in Content Tagger. This work is an interim step to take a set of mapping spreadsheets on Google Drive, compile them, and create the taxons themselves in the Content Store. This PR will be followed by code that creates the taxon/link mappings as well.

Eventually this could all be replaced by a bulk-import tool built into Content Tagger itself. For now though, the sheet download and taxon creation can simply be run directly with `rails runner`.

https://trello.com/c/bsZAwZ35


lib/alpha_taxonomy/sheet_parser.rb
```diff
@@ -2,40 +2,50 @@
 
 module AlphaTaxonomy
   class SheetParser
-    def initialize(taxonomy_data)
-      @taxonomy_data = taxonomy_data
+    class BlankMappingField < StandardError; end
```
tijmenb: `BlankMappingFieldError`? :)

lib/alpha_taxonomy/taxon_presenter.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    def present
+      {
+        base_path: base_path,
+        format: "taxon",
+        title: @title,
+        publishing_app: 'content-tagger',
+        rendering_app: 'collections',
+        public_updated_at: DateTime.current.to_s,
```
tijmenb: In other apps we use `iso_8601`, is this the same?

lib/alpha_taxonomy/create_taxons.rb
```diff
@@ -0,0 +1,52 @@
...
+    def existing_taxons
+      @existing_taxons ||= Services.publishing_api.get_content_items(
+        content_format: 'taxon',
+        fields: %i(title base_path content_id details)
+      )
+    end
+
+    def create_new_taxon(presented_payload:)
+      content_id = SecureRandom.uuid
+      Services.publishing_api.put_content(
```
tijmenb: We need to `publish` too here, to allow them to be seen on the site.

lib/alpha_taxonomy/taxon_presenter.rb
mobaig: Yes.
```ruby
irb(main):004:0> DateTime.current.to_s
=> "2015-12-14T17:09:40+00:00"
irb(main):005:0> DateTime.current.iso8601
=> "2015-12-14T17:09:48+00:00"
```

lib/alpha_taxonomy/sheet_parser.rb
mobaig: :) - updated.

rboulton:
This looks like a good idea, but I'm a bit worried by alpha taxons being able to be created by two apps (collections-publisher and content-tagger).  In particular, what will happen if content-tagger creates a taxon, and then collections-publisher is used to edit it?  I think publishing-api only allows a single app to own a content item, so I think it may give a conflict error in that situation (not certain if this is still how it works, though).

I'm also a bit concerned that the details of what goes into the taxons will now need to be updated in two places when we want to change it - the schemas will help here, but things like optional fields, or default values for fields, will still need to be maintained carefully.  Also, we're not yet checking that everything going to the publishing api satisfies the schemas (but we will be soonish, hopefully).

Could you put a bit more detail in the pull request and commit descriptions, please? (links to trello are good, but tend to break after a while).  It would be good to rebase away the accidentally committed merge conflict (in https://github.com/alphagov/content-tagger/commit/598839c427075f376d00a3d8e45b2d9c5cd61924), and make sure to explain why changes are being made in the commit messages (eg, https://github.com/alphagov/content-tagger/commit/7477fb17f06fbb97020bb1f1fb8cb3fda34f450c - what's changed about the linter, and why?)

tijmenb:
Heart of the problem is which functionality to put where. We already know that there are some parts of the tagging world that could be in both content tagger and collections publisher. They are very closely related and might even turn out to be the same app.

I think the bulk creation of tags should be possible in the same flow as bulk tagging, so for now I think it might be reasonable to allow content-tagger to create taxons too.

Once we get a better understanding of how everything fits together, we could decide to merge collections-publisher and content-tagger, or move the taxonomy manager to content-tagger.

It would be good to have a whiteboard about the the responsibilities of the different apps.

mobaig:
Perhaps it would be easier to simply put everything in one place for now? Rather than having duplication and/or muddied responsibilities between two apps, we could simply build out this new taxonomy functionality in collections-publisher. If a clear boundary for extraction emerges then we can do so later on. Keeping the taxonomy-related code decently encapsulated and well-named will make that easier.

Unless of course this doesn't matter. For example, are there any significant negative implications of the app names we've chosen in lines 16-17 here?: https://github.com/alphagov/content-tagger/pull/13/files#diff-ff4e30814d2d29ca33294961ca432cd8R16

I've incorporated the PR feedback and got the lint checks passing. Some Jenkins config update was required. This is ready for review again, @tijmenb and @rboulton .

tijmenb:
If we are putting it in one app, I would prefer moving the taxonomy manager here. This alpha taxonomy feature is part of the future bulk tagger, which should probably be in this application.

mobaig:
That seems like a sound approach. Taxonomy-related functionality should exist in one place, irrespective of where we decide that is. Anything is better than dealing with taxons/mappings in two apps.

lib/alpha_taxonomy/taxon_presenter.rb
rboulton: Given the wide range of date formats, I think being explicit is probably better (ie, would prefer to see `iso_8601` used)

mobaig: In hindsight, I'm inclined to agree. Updated.

mobaig:
@tijmenb reworked the jenkins script. That was a royal mess because I didn't realise that the failures in CI were due to the RSpec `after(:suite)` callback running the linter, not the stuff in `jenkins.sh`. WDYT.

Gemfile
```diff
@@ -24,6 +24,7 @@ group :development, :test do
   gem 'rspec-rails', '~> 3.3'
   gem 'simplecov', '0.10.0', require: false
   gem 'simplecov-rcov', '0.2.3', require: false
+  gem 'govuk-content-schema-test-helpers'
```
tijmenb: Pin this to `~> 1.4`.

mobaig: Done.

PR #15, opened 2015-12-18
Alpha taxonomy refactor

mobaig:
- Naming changes and movement of code to clarify responsibilities of each
  class.
- Consolidate parsing/writing of downloaded data and handling of the file
  on disk into a single ImportFile class (renamed from SheetParser, which
  was a slight misnomer).
- Make SheetDownloader solely responsible for fetching across the
  network.
- Rename CreateTaxons to TaxonCreator, for consistency with the other
  noun-based class names.


spec/spec_helper.rb
```diff
@@ -28,7 +28,9 @@
   end
 
   config.after :suite do
-    require "govuk/lint/cli"
-    Govuk::Lint::CLI.new.run([])
+    unless ENV["SKIP_LINT"]
```
tijmenb: Should we add `SKIP_LINT=1` to jenkins.sh?

mobaig: Don't reckon we need to. This is just to run tests with a bit less noise in dev.

PR #16, opened 2015-12-23
Add functionality to bulk assign content items to taxons

mobaig:
This is required as part of the functionality to bulk assign content
items to taxons, based on the alpha taxonomy data that is currently
being produced by Information Architects in the Finding Things team.

This change adds a method which inspects the locally-downloaded taxonomy
data and groups list of taxons to each base path. Part of this change
includes moving the actual file creation to the #populate method rather
than doing this on initialize. Any client of the ImportFile class will
have to ensure the file is created via #populate before calling anything
else.

Trello: https://trello.com/c/bsZAwZ35

### DIFF
--- a/lib/alpha_taxonomy/taxon_linker.rb
+++ b/lib/alpha_taxonomy/taxon_linker.rb
@@ -1,0 +1,23 @@
+require "csv"
+
+module AlphaTaxonomy
+  class TaxonLinker
+    def run!
+      ImportFile.new.grouped_mappings.each do |base_path, taxon_titles|
+        taxon_content_ids = taxon_titles.map do |taxon_title|
+          taxon_base_path = TaxonPresenter.new(title: taxon_title).base_path
+          taxon_content_item = Services.content_store.content_item!(taxon_base_path)
+          taxon_content_item["content_id"]
+        end
+        target_content_item = Services.content_store.content_item!(base_path)
+
+        Services.publishing_api.put_links(
+          target_content_item["content_id"],
+          links: {
+            alpha_taxons: taxon_content_ids
+          }
+        )
+      end
+    end
+  end
+end
--- a/lib/alpha_taxonomy/taxon_linker.rb
+++ b/lib/alpha_taxonomy/taxon_linker.rb
@@ -2,22 +2,60 @@ require "csv"
 
 module AlphaTaxonomy
   class TaxonLinker
-    def run!
-      ImportFile.new.grouped_mappings.each do |base_path, taxon_titles|
-        taxon_content_ids = taxon_titles.map do |taxon_title|
-          taxon_base_path = TaxonPresenter.new(title: taxon_title).base_path
-          taxon_content_item = Services.content_store.content_item!(taxon_base_path)
-          taxon_content_item["content_id"]
-        end
-        target_content_item = Services.content_store.content_item!(base_path)
+    class TaxonNotInContentStoreError < StandardError; end
 
+    def initialize(logger: Logger.new(STDOUT))
+      @log = logger
+      @errors = []
+    end
+
+    def run!
+      grouped_mappings = ImportFile.new.grouped_mappings
+
+      grouped_mappings.each do |base_path, taxon_titles|
+        taxon_content_ids = find_content_ids_for(taxon_titles)
+        target_content_item_id = fetch_content_item_id_with(base_path)
+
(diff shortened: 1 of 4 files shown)


lib/alpha_taxonomy/taxon_linker.rb
```diff
@@ -0,0 +1,23 @@
...
+module AlphaTaxonomy
+  class TaxonLinker
+    def run!
+      ImportFile.new.grouped_mappings.each do |base_path, taxon_titles|
+        taxon_content_ids = taxon_titles.map do |taxon_title|
+          taxon_base_path = TaxonPresenter.new(title: taxon_title).base_path
+          taxon_content_item = Services.content_store.content_item!(taxon_base_path)
+          taxon_content_item["content_id"]
+        end
+        target_content_item = Services.content_store.content_item!(base_path)
```
tijmenb: What happens if this raises 404?

tijmenb:
It would be good to add a integration test that describes the full process. Something like given there is a google doc, when the importer runs, new taxons are created if they don't exist yet and links are sent to the publishing_api.

lib/alpha_taxonomy/taxon_linker.rb
tijmenb: Perhaps you can use `ContentLookupForm` for this lookup (maybe rename that), so that when we get a lookup endpoint on publishing-api we can change it in one place. It also validates the base path.

lib/alpha_taxonomy/taxon_linker.rb
```diff
@@ -0,0 +1,23 @@
+require "csv"
+
+module AlphaTaxonomy
+  class TaxonLinker
+    def run!
+      ImportFile.new.grouped_mappings.each do |base_path, taxon_titles|
+        taxon_content_ids = taxon_titles.map do |taxon_title|
```
tijmenb: This lookup is perhaps better done by fetching all taxons outside of this loop (`#get_content_items(..)`) and then a `find` by title. Could save a lot of requests.

spec/lib/alpha_taxonomy/taxon_linker_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      ).and_return("content_id" => content_id)
+    end
+
+    it "creates taxon links based on the grouped mappings" do
+      import_file = AlphaTaxonomy::ImportFile.new
+      allow(import_file).to receive(:grouped_mappings).and_return(
+        "/a-foo-content-item" => ["Foo Taxon"],
+        "/a-foo-bar-content-item" => ["Foo Taxon", "Bar Taxon"]
+      )
+      allow(AlphaTaxonomy::ImportFile).to receive(:new).and_return(import_file)
```
tijmenb: The stubbing makes this a little hard to follow. Perhaps it is clearer if you make `TaxonLinker.new.run!` take a `AlphaTaxonomy::ImportFile` as an argument. Then you could do this:

```ruby
import_file = AlphaTaxonomy::ImportFile.new
import_file.stub(grouped_mappings: [...])
AlphaTaxonomy::TaxonLinker.new.run!(import_file)
```

lib/alpha_taxonomy/taxon_linker.rb
mobaig: re: 404 - it fails noisily. This should probably be re-worked to report a list of failed paths at the end of the run.

Good suggestion re: `ContentLookForm`, will revise.

lib/alpha_taxonomy/taxon_linker.rb
mobaig: Good suggestion. Intended to optimise this later if it ever presented an issue, but that might be an easy drop-in.

spec/lib/alpha_taxonomy/taxon_linker_spec.rb
mobaig: A decent option, but it makes running this command more burdensome. You can't just chuck the `run!` invocation into the rails `runner`, you'd either have to load up a console in order to instantiate the `import_file`, or wrap the whole thing in a rake task. Added to that is that the `TaxonLinker` isn't used in multiple contexts, making the use of dependency injection here a slightly odd choice.

Not sure it's worth the trade off, particularly as the readability improvement is slight at the most.

mobaig:
closing temporarily - adding an integration test

PR #18, opened 2016-01-05
Add functionality to bulk assign content items to taxons

mobaig:
This is required as part of the functionality to bulk assign content
items to taxons, based on the alpha taxonomy data that is currently
being produced by Information Architects in the Finding Things team.

This change adds a method which inspects the locally-downloaded taxonomy
data and groups list of taxons to each base path. Part of this change
includes moving the actual file creation to the #populate method rather
than doing this on initialize. Any client of the ImportFile class will
have to ensure the file is created via #populate before calling anything
else.

Trello: https://trello.com/c/bsZAwZ35

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,6 +15,8 @@ LineLength:
 
 Metrics/AbcSize:
   Enabled: true
+  Exclude:
+    - spec/features/**/*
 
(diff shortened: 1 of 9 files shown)


bishboria:
:+1: I'll Merge tomorrow once someone else has had a look

PR #19, opened 2016-01-07
Add additional logging to alpha taxonomy bulk import code

mobaig:
To assist debugging and provide visual indicator of progress when running
imports from the command line.


jackscotti:
I know this could sound annoying and I don't know what do we do with loggers. But would it be worth testing that the logger is actually doing what we're expecting  it to do?

mobaig:
It's certainly a valid suggestion, but IMHO I don't think testing for the presence of those info messages is worth the overhead of adding (and maintaining) additional specs. We're testing the Logger API (via the injected StringIO logger), and there's some light testing of useful error messages (see `taxon_linker_spec.rb`). That's probably sufficient given this is an internal tool rather than something user-facing.

PR #21, opened 2016-01-12
Add bearer_token for publishing API authentication

danielroseman:
https://trello.com/c/tx50hSOx/492-add-bearer-token-value-to-all-apps-using-publishing-api


mobaig:
lg

PR #25, opened 2016-02-09
Allow tagging of collections-publisher documents

tijmenb:
`collections-publisher` publishes topics and mainstream browse pages. This commit allows those pages to be tagged.

This allows users, for example, to add link topics to other topics, and add organisations to topics. There are no plans to do this, but we want as few special cases as possible. It would be therefore be good to
treat the pages managed by collections-publisher as just a normal piece of content on GOV.UK.

After this is merged, collections-publisher should also be added to the list in https://github.com/alphagov/rummager/pull/555.

Trello: https://trello.com/c/GRt7NSbL

### DIFF
--- a/config/tagging-apps.yml
+++ b/config/tagging-apps.yml
@@ -7,7 +7,7 @@ a-migrated-app: content-tagger # for testing puposes
 businesssupportfinder: content-tagger
 calculators: content-tagger
 calendars: content-tagger
-collections-publisher: collections-publisher
+collections-publisher: content-tagger
 contacts: panopticon
 contacts-admin:
 design-principles:


rboulton:
This is probably fine.  However, have you thought about how this might interact with email alerts.  For example, if a topic A were tagged to a topic B, when would an email alert for topic B be sent?  I think it would be sent if topic A had its title or other metadata changed, because collections-publisher always sends updates as major updates.  It wouldn't be sent if content was tagged to topic A, but would be sent if topic A was curated.  Is this what we'd want to happen?

Would it be appropriate to change collections publisher to send updates to existing topics or mainstream browse pages as "minor" updates, to avoid risk of unwanted email notifications?

tijmenb:
That's a good point. Let's set the update type to minor there.

PR #29, opened 2016-02-18
Add support for `lock_version` rename

tijmenb:
`version` is going to be renamed to `lock_version` in https://github.com/alphagov/publishing-api/pull/206.


tijmenb:
Talked to @tuzz, and they will change `lock_version` back to `version`, so closing this PR.

PR #31, opened 2016-02-19
Taxon bulk importer - change sheet downloader to use env-var

mobaig:
Make the downloader read from an environment variable instead of hard-
coding sheet credentials in the class. This makes running different
imports easier by removing the need for a new deploy.

### DIFF
--- a/lib/alpha_taxonomy/sheet_downloader.rb
+++ b/lib/alpha_taxonomy/sheet_downloader.rb
@@ -5,24 +5,32 @@ module AlphaTaxonomy
     # a) it is stored on Google drive.
     # b) it is 'published' as a single sheet (not the entire document/workbook),
     #    with tab-seperated values.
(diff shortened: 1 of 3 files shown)


lib/alpha_taxonomy/sheet_downloader.rb
```diff
@@ -5,27 +5,35 @@ class SheetDownloader
...
+      sheet_credential_tuples.each do |sheet_credentials|
         yield remote_taxonomy_data(sheet_credentials)
       end
     end
 
+    def sheet_credential_tuples
+      environment_values = ENV.fetch("TAXON_SHEETS")
+      environment_values = environment_values.split(',')
+      if environment_values.count % 3 != 0
+        raise ArgumentError, "TAXON_SHEETS should be a sequence of comma-separated tuples, like so: name1,key1,gid1,name2,key2,gid2"
```
Davidslv: you mean triplets?

PR #33, opened 2016-02-19
Add a README for the alpha taxonomy bulk import scripts

mobaig:


tijmenb:
:+1: great!

tijmenb:
I do suspect we can combine a few steps of the readme in a rake task, is that right?

lib/alpha_taxonomy/README.md
```diff
@@ -0,0 +1,55 @@
...
+  +-------------------+-----------------+
+  ```
+
+0. Make the sheet public on Google Drive in TSV format. Its public URL should
+   look something like this:
+   `https://docs.google.com/spreadsheets/d/19GhkAQ9VEmsiPeoHbrz9Q-nTnbtLxC2kkD6szoGGam0/edit#gid=1102496302`.
+   The `19Ghka...` part is the `key` and the param at the end is the `gid`.
+0. In the environment in which you're running the import, set an environment
+   variable `TAXON_SHEETS` with values in the form
+   `name-of-sheet,the-key,the-gid,name-of-some-other-sheet,its-key,its-gid`.
```
jackscotti: gid == Google ID?

mobaig: Yes, I'd assume so.

mobaig:
@tijmenb re: Rake task - yes, definitely. I've been running them separately while developing and improving each stage but it would be useful to codify the run-order in a task. Will add in a followup.

tijmenb:
:+1:

PR #35, opened 2016-02-22
Allow contacts pages to be tagged

tijmenb:
contacts-admin has been migrated since alphagov/contacts-admin#231.

Trello: https://trello.com/c/afa0Mfa4


Davidslv:
:+1:

PR #36, opened 2016-02-23
Taxon bulk importer - update taxon linker to use live content store

mobaig:
Using the draft content store to fetch content items results in the
occasional 403 due to access limitation. This uses the live content
store to retrieve content IDs instead, where access limits aren't
present.


tijmenb:
Note that we will add an endpoint to publishing-api to fetch a content_id by base_path. After that is done this app won't use (draft) content-store anymore.

PR #37, opened 2016-02-23
Taxon bulk importer - bugfix in taxon linker

mobaig:
Return nil when a content item is not found, so no attempt at updating
links is made.


MatMoore:
:+1:

tijmenb:
Since this was an actual production bug, it might be wise to add a test for this..

PR #41, opened 2016-03-10
Use `get_linkables` endpoint for tag fetching

tijmenb:
This new endpoint is better suited to populate the dropdowns. This also fixes a current bug caused by the publishing-api returning more data than expected.


mobaig:
LGTM

PR #42, opened 2016-03-10
Add support for blacklisting fields, and blacklist parent for TAP content

mobaig:
This adds functionality to switch off certain kinds of links in the
content tagger form, and sets this up specifically for 'parent' fields
when the document being tagged is owned by Travel Advice Publisher.

We need to be able to do this in order to migrate Travel Advice Publisher
to the new tagging architecture. Currently, the parent link for TAP
content is set by default to be an array of hashes modelling a
breadcrumb. This isn't something we're able to handle in content-tagger
right now. The workaround in this commit allows us to migrate TAP away
from Panopticon while we determine how best to handle this on an ongoing
basis.

Part of: https://trello.com/c/Unia7Z0B

### DIFF
--- a/app/controllers/content_controller.rb
+++ b/app/controllers/content_controller.rb
@@ -2,6 +2,7 @@ class ContentController < ApplicationController
   def show
     @content_item = ContentItem.find!(params[:content_id])
     @tagging_update = TaggingUpdateForm.init_with_content_item(@content_item)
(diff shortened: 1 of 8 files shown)


tijmenb:
Very nice! :+1: 

I've added a little test for `TaggingUpdateForm` to make sure we don't start sending `parent` in the future.

I'm good with this, but because I've added some commits now I'll let @bishboria or @Davidslv merge this tomorrow.

PR #44, opened 2016-03-14
Use get_linkables endpoint to get taxon data.

danielroseman:
TaxonCreator was requesting the details hash, but was not using it (and the
hash is empty for all taxons anyway), so linkables provides all the necessary
data.


tijmenb:
Thanks for doing this! :+1:

PR #45, opened 2016-03-16
Add functionality to bulk delete mappings for a taxon

mobaig:
This is useful in cases we want to upload a revised dataset for a given
taxon level as part of the alpha taxonomy definition work.

Trello: https://trello.com/c/UiRIcF91/551-taxon-bulk-import-add-functionality-to-bulk-delete-mappings-and-rename-taxons-with-dodgy-base-paths

### DIFF
--- a/lib/alpha_taxonomy/taxon_link_deleter.rb
+++ b/lib/alpha_taxonomy/taxon_link_deleter.rb
@@ -1,0 +1,31 @@
+module AlphaTaxonomy
+  class TaxonLinkDeleter
+    def initialize(logger: Logger.new(STDOUT), base_paths:)
+      @log = logger
+      @base_paths = base_paths
+    end
+
+    def run!
+      @base_paths.each do |base_path|
+        @log.info "Clearing links for #{base_path}"
+
+        taxon_content_id = Services.publishing_api.lookup_content_id(base_path: base_path)
+        if taxon_content_id.blank?
+          raise ArgumentError, "No content ID found for base path #{base_path}"
+        end
+
+        linked_items = Services.publishing_api.get_linked_items(
+          taxon_content_id, link_type: 'alpha_taxons', fields: %w(content_id base_path)
+        )
+
+        linked_items.each do |linked_item|
+          @log.info "-- updated #{linked_item.fetch('base_path')}"
+          Services.publishing_api.patch_links(
+            linked_item.fetch("content_id"),
+            links: { alpha_taxons: [] }
+          )
+        end
+      end
+    end
+  end
+end
--- a/lib/tasks/taxonomy.rake
+++ b/lib/tasks/taxonomy.rake
@@ -1,0 +1,32 @@
+namespace :taxonomy do
+  desc "Delete all taxon links for the list of taxon base paths set in the environment"
+  task delete_links: :environment do
+    base_paths = ENV.fetch("TAXON_CLEANUP").split(",")
+    AlphaTaxonomy::TaxonLinkDeleter.new(base_paths: base_paths).run!
+  end
+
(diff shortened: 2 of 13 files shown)


tijmenb:
How do we execute this code?

mobaig:
@tijmenb `rails runner` on the target machine.

mobaig:
Added several rake tasks wrapping the various bulk import commands so we can run this via Jenkins on prod.

lib/tasks/taxonomy.rake
```diff
@@ -0,0 +1,31 @@
+namespace :taxonomy do
+  task delete_links: :environment do
+    desc "Delete all taxon links for the list of taxon base paths set in the environment"
```
tijmenb: It may be easy to mention the specific ENV variable you need to set to run the task.

lib/tasks/taxonomy.rake
```diff
@@ -0,0 +1,31 @@
+namespace :taxonomy do
```
tijmenb: The `desc` call must be above the `task` call, otherwise `rake -T` will not list the task.

lib/alpha_taxonomy/taxon_link_deleter.rb
```diff
@@ -0,0 +1,27 @@
...
+      base_paths = ENV.fetch("TAXON_CLEANUP").split(",")
+
+      base_paths.each do |base_path|
+        @log.info "Clearing links for #{base_path}"
+
+        linked_items = Services.live_content_store.incoming_links!(
+          base_path, types: ['alpha_taxons']
+        )['alpha_taxons']
+
+        linked_items.each do |linked_item|
```
tijmenb: `alpha_taxons` can be nil if the item is not tagged (anymore) to a taxon. I think this `each` would error in that case.

lib/alpha_taxonomy/taxon_link_deleter.rb
```diff
@@ -0,0 +1,27 @@
+module AlphaTaxonomy
+  class TaxonLinkDeleter
+    def initialize(logger: Logger.new(STDOUT))
+      @log = logger
+    end
+
+    def run!
+      base_paths = ENV.fetch("TAXON_CLEANUP").split(",")
```
tijmenb: If you move this over to the rake task this class can be re-used elsewhere and you don't have to stub `ENV` in the tests.

lib/alpha_taxonomy/taxon_link_deleter.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      base_paths.each do |base_path|
+        @log.info "Clearing links for #{base_path}"
+
+        linked_items = Services.live_content_store.incoming_links!(
```
tijmenb: Could we document why we're not using publishing-api here?

mobaig: Sure, will do.

lib/alpha_taxonomy/taxon_link_deleter.rb
mobaig: Yeah, this whole ENV thing is definitely becoming a bit leaky. I'd decided to defer this change but it makes sense given the rake tasks are now the main entry-point to all this logic. Will update.

lib/tasks/taxonomy.rake
mobaig: Not entirely sure what I was smoking yesterday. You're quite right.

lib/alpha_taxonomy/taxon_link_deleter.rb
mobaig: I'm pretty sure it's an empty array in that case. Will cast this anyway, it'll be a bit more robust at least.

PR #46, opened 2016-03-18
Use publishing-api lookup endpoint

tijmenb:
This PR makes the content-tagger use the publishing-api for everything. This is better because the publishing-api is synchronous and we'll have less dependencies.

https://trello.com/b/Kg3JL5Cd


mobaig:
Nice :+1:

PR #49, opened 2016-03-21
Hide parent tag for contacts pages

tijmenb:
Pages published by contacts should not have a "Primary mainstream browse page" tagging.
https://github.com/alphagov/content-tagger/pull/48 tried to add that, but disallowed it for the wrong publishing app.

I've removed `contacts-admin` from the list to avoid confusion. This value was used for certain pages, but isn't anymore. All pages by `contacts-admin` have the publishing_app value of `contacts`.

Trello: https://trello.com/c/afa0Mfa4

### DIFF
--- a/config/blacklisted_tag_types.yml
+++ b/config/blacklisted_tag_types.yml
@@ -1,4 +1,4 @@
-contacts-admin:
+contacts:
   - parent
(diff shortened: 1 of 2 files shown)


MatMoore:
:+1:

PR #52, opened 2016-03-30
Replace publishing-app whitelist with blacklist

MatMoore:
Only a few apps are left to be migrated now, and this will allow
us to build new publishers without having to modify content tagger
all the time.

https://trello.com/c/tZd23gW5


tijmenb:
Looks good to me, but I was there when this was written. Can @mobaig review this?

PR #53, opened 2016-04-01
Add AlphaTaxonomy::TaxonRenamer

Davidslv:
Script to rename taxons with spurious numeric prefix.

They were introduced in a spreadsheet in order to know the hierarchy of
certain taxonomies, but they should not belong in the base_path.

This problem wont recur in future because the spreadsheet has been
fixed.

paired with @benhyland

### DIFF
--- a/lib/alpha_taxonomy/taxon_renamer.rb
+++ b/lib/alpha_taxonomy/taxon_renamer.rb
@@ -1,0 +1,43 @@
+module AlphaTaxonomy
+  class TaxonRenamer
+    def initialize(logger: Logger.new(STDOUT), base_paths:)
+      @log = logger
+      @base_paths = base_paths
+    end
+
+    def run!
+      @base_paths.each do |base_path|
+        content_id = lookup_id_by_base_path(base_path)
+
+        next unless content_id
+
+        @log.info "-- Requesting base_path change to #{base_path[:to]}"
+
+        send_to_publishing_api(
+          content_id,
+          AlphaTaxonomy::TaxonPresenter.new(title: normalised_title(base_path[:to])).present
+        )
+      end
+    end
+
+  private
+
+    def normalised_title(base_path)
+      base_path.split('/').last.tr('-', ' ').capitalize
+    end
+
+    def send_to_publishing_api(content_id, payload)
+      Services.publishing_api.put_content(content_id, payload)
+      Services.publishing_api.publish(content_id, 'major')
+    end
+
+    def lookup_id_by_base_path(base_path)
+      @log.info "Requesting content_id to Publishing API for #{base_path[:from]}"
+      content_id = Services.publishing_api.lookup_content_id(base_path: base_path[:from])
+
+      @log.info "#{base_path[:from]} content_id was not found" unless content_id
+
+      content_id
+    end
+  end
+end
--- a/lib/alpha_taxonomy/taxon_renamer.rb
+++ b/lib/alpha_taxonomy/taxon_renamer.rb
@@ -6,16 +6,16 @@ module AlphaTaxonomy
     end
 
     def run!
-      @base_paths.each do |base_path|
-        content_id = lookup_id_by_base_path(base_path)
+      paths_and_content_ids = lookup_content_ids_by_base_paths
 
-        next unless content_id
+      @base_paths.each do |base_path_pair|
+        content_id = paths_and_content_ids[base_path_pair[:from]]
+        title = normalised_title(base_path_pair[:to])
 
-        @log.info "-- Requesting base_path change to #{base_path[:to]}"
-
+        @log.info "-- Requesting base_path change for #{base_path_pair[:from]}"
         send_to_publishing_api(
           content_id,
-          AlphaTaxonomy::TaxonPresenter.new(title: normalised_title(base_path[:to])).present
+          AlphaTaxonomy::TaxonPresenter.new(title: title).present
         )
       end
     end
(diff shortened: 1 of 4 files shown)


tijmenb:
Good work!

Perhaps it would be good to split the input/parse thing from the setting up of redirects. The current method of passing in the base paths is very coupled to the (limitations of) the rake task. If we would ever create a UI for this feature, having the method take the string with base paths is not very easy to use.

Would there be an better way to pass in the redirects? I suspect the concatenating the base paths is error prone - it could be quite hard to see what is the from/to from a long list.

lib/alpha_taxonomy/taxon_renamer.rb
```diff
@@ -0,0 +1,72 @@
...
+    def lookup_id_by_base_path(base_path)
+      @log.info "Requesting content_id to Publishing API for #{base_path[:from]}"
+      content_id = Services.publishing_api.lookup_content_id(base_path: base_path[:from])
+
+      @log.info "#{base_path[:from]} content_id was not found" unless content_id
+
+      content_id
+    end
+
+    def create_payload(title, base_path)
```
tijmenb: Would it be possible to use the [presenter](https://github.com/alphagov/content-tagger/blob/4c52589e4ef8b24aefdda26eaa22df0db20dd23b/lib/alpha_taxonomy/taxon_presenter.rb)?

spec/lib/alpha_taxonomy/taxon_renamer_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+          publishing_app: 'collections-publisher',
+          rendering_app: 'collections',
+          format: 'taxon',
+          routes: [
+            { path: base_path, type: "exact" },
+          ]
+        }
+      end
+
+      it "requests Publishing API to change title and base_path" do
```
tijmenb: It would be good to test the payload against the schema like:

https://github.com/alphagov/content-tagger/blob/4c52589e4ef8b24aefdda26eaa22df0db20dd23b/spec/lib/alpha_taxonomy/taxon_creator_spec.rb#L89-L91

mobaig:
Re: base_paths argument. Agreed that we can probably do more to make the Renamer easier to use. Just spitballing here, but what if it accepted something like:

```
[
    { from: "taxon/old-path", to: "taxon/new-path" },
    { from: "taxon/base-path", to: "taxon/good-path" },
]
```

With the parsing of the command-line args to produce the above happening outside the Renamer. Maybe in the rake task, or some helper method invoked in the rake task.

Note that it'd probably be worthwhile writing a test for this arg-parsing logic, wherever it ends up.

lib/alpha_taxonomy/taxon_renamer.rb
Davidslv: Yep, you guys are right, I misread the presenter usage, will change this.

Davidslv:
I've updated the PR, hopefully it addresses both your comments @mobaig and @tijmenb.

Thank you.

tijmenb:
@Davidslv looks good! Some minor things but we'll do those in person this morning.

lib/alpha_taxonomy/taxon_renamer.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    def run!
+      paths_and_content_ids = lookup_content_ids_by_base_paths
+
+      @base_paths.each do |base_path_pair|
+        content_id = paths_and_content_ids[base_path_pair[:from]]
+        title = normalised_title(base_path_pair[:to])
+
+        @log.info "-- Requesting base_path change for #{title}"
```
mobaig: Sorry, this is really picky but - I think I'd prefer `base_path_pair[:from]` here rather than `title`. I've just found the URL to be more readable (and easier to spot on the screen) when running these tasks in the console. Also, we've been 'handling' urls in the run up to this task executing - it's a bit less of a cognitive switch if we keep referring to these taxons by their URL.

lib/tasks/taxonomy.rake
```diff
@@ -30,4 +36,21 @@ namespace :taxonomy do
...
+
+  def parse_base_paths_string(base_paths)
+    pair_of_paths = base_paths.split('|')
+
+    unless pair_of_paths.all? { |pair| (pair.split(',').size % 2).zero? }
+      raise ArgumentError, base_paths_error_message
+    end
+
+    pair_of_paths.map do |pair|
+      splited_pair = pair.split(',')
```
mobaig: `split_pair`

lib/tasks/taxonomy.rake
```diff
@@ -30,4 +36,21 @@ namespace :taxonomy do
...
+      { from: splited_pair.first, to: splited_pair.last }
+    end
+  end
+
+  def base_paths_error_message
+    "base_paths is expected to be composed of two comma-separated values, like so: '/path1,/path1-rename|/path2,/path2-rename,...'"
```
mobaig: Maybe - `base_paths should be a set of pairs, delimited by a pipe character, like so: ...`

spec/lib/tasks/taxonomy_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'rails_helper'
+require 'rake'
+
+RSpec.describe 'taxonomy rake task' do
+  describe 'taxonomy:rename_base_paths' do
```
mobaig: Nice.

mobaig:
Some minor bits from me, but otherwise LG. Give us a shout when resolved and I'll merge.

lib/tasks/taxonomy.rake
```diff
@@ -30,4 +36,21 @@ namespace :taxonomy do
...
+
+  def parse_base_paths_string(base_paths)
+    pair_of_paths = base_paths.split('|')
+
+    unless pair_of_paths.all? { |pair| (pair.split(',').size % 2).zero? }
+      raise ArgumentError, base_paths_error_message
+    end
+
+    pair_of_paths.map do |pair|
+      split_pair = pair.split(',')
```
dougdroper: from, to = pair.split(',')

Davidslv: sorry @dougdroper, we missed your comment, but just wanted to let you know that we do agree with your suggestion, thank you.

dougdroper: Yes sure no problem @Davidslv, I didn't see that it was merged already.

PR #58, opened 2016-04-15
Use publisher style dropdown

MatMoore:
Use publisher style dropdown for tags

- Add Linkables class to handle fetching tags
- Add specs for Linkables
- Rewrite features to use more realistic example tags
- Ensure stubs are reset between specs

This is very similar to the implementation in publisher except that we use content ids to refer to tags instead of URLs. This will be the standard going forward and we would like to extract this functionality to a common library.

We also took out the in memory caching on the Linkables class itself, because this will prevent tags from being updated during the lifecycle of the unicorn process, and it is difficult to test. We don't think we need caching for these requests, but if we do, we can always enable it on the gds-api-helper instance.

Paired with @Davidslv 
Trello ticket: https://trello.com/c/HbDi3Yqx/586-content-tagger-use-publisher-style-dropdown-in-content-tagger-for-topics-browse-pages


tijmenb:
> We also took out the in memory caching on the Linkables class itself, because this will prevent tags from being updated during the lifecycle of the unicorn process, and it is difficult to test. We don't think we need caching for these requests.

That only applies because it's only class methods now. Both `parent` and `mainstream_browse_pages` use the same data but will make the request twice. Can we prevent that? The publishing-api is slow enough for this to make a difference - we regularly get timeouts on content-tagger pages.

spec/features/tagging_content_spec.rb
```diff
@@ -1,6 +1,15 @@
 require "rails_helper"
 
-RSpec.describe "Tagging content" do
+RSpec.describe "Tagging content", type: :feature do
+  include PublishingApiHelper
+
+  before do
```
tijmenb: There's a bit of a trade off here, but I generally prefer to avoid "mystery guests":

https://robots.thoughtbot.com/mystery-guest

In this test, we start seeing a "magic" thing named "Business tax / Pension scheme administration". The reader would have to have the knowledge that this is a side effect from `publishing_api_has_topics` to understand that.

There are two ways of preventing it:

1. Move the entire `PublishingApiHelper` into this class
2. Explicitly create the data we need for this test via the test helper (`publishing_api_has_topics(title: "Business tax / Pension scheme administration")` and have the test set the whole thing up properly

MatMoore: I agree with the problem but not entirely sure how to fix it. Option 2 sounds the best - I'd ideally like some kind of factory that will default the things I don't care about. Is there some existing example of this I can look at?

What do you think about something like this:

```ruby
  def publishing_api_has_organisations(base_paths)
    publishing_api_has_linkables(
      stubbed_organisations.select { |org| base_paths.include?(org["base_path"]) },
      document_type: 'organisation'
    )
  end
```

```ruby
    publishing_api_has_organisations(
      [
        "/government/organisations/student-loans-company",
        "/my-content-item",
      ]
    )
```

MatMoore:
> That only applies because it's only class methods now. Both parent and mainstream_browse_pages use the same data but will make the request twice. Can we prevent that? The publishing-api is slow enough for this to make a difference - we regularly get timeouts on content-tagger pages.

We've prevented this by making Linkables instances rather than using caching on the class. So now we share the same data within the context of a request, and only make one request per document type. Do you think it's worth enabling the publishing api adapter cache as well? I thought these requests had been optimised already.

tijmenb:
@MatMoore excellent choice! The reason for disabling the gds-api-adapters caching is that it's annoying for users to have to wait for up to 15 minutes before a new tag appears.

spec/features/tagging_content_spec.rb
tijmenb: Yes, something like that would be great 👍

MatMoore: We've made this change, and also removed some redundant stubs we added by accident.

PR #60, opened 2016-04-28
Allow content-tagger to deal with specialist-publisher content

benhyland:
As part of migrating specialist-publisher tagging to content-tagger,
allow content-tagger to handle browse pages and topics for specialist-publisher.

We already prevented panopticon from handling specialist-publisher tags
(https://github.com/alphagov/panopticon/pull/357 and https://github.com/alphagov/panopticon/pull/358).

We make an exception for the link type 'topics' for specialist-publisher finders.
Specialist publisher still hardcodes these, and they are not yet in publishing-api,
so it is not safe to handle them in content-tagger.

https://trello.com/c/YoeWtswr/609-totally-migrate-specialist-publisher-v1-tagging

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -1,13 +1,14 @@
 class ContentItem
   TAG_TYPES = %w(mainstream_browse_pages parent topics organisations alpha_taxons)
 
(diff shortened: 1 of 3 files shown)


mobaig:
LG

PR #61, opened 2016-04-28
Send links for specialist finders to publishing-api

tijmenb:
This adds a migration with a number of patch-links calls to the publishing-api.

The links it will send are the same as defined in the finder JSON blobs on specialist-publisher:

https://github.com/alphagov/specialist-publisher/blob/28ec15008b07affa8e6f0af5bc3d2f039690cbf7/finders/metadata/medical-safetyalerts.json#L17-L20

https://github.com/alphagov/specialist-publisher/blob/28ec15008b07affa8e6f0af5bc3d2f039690cbf7/finders/metadata/cma-cases.json#L13-L19

https://github.com/alphagov/specialist-publisher/blob/28ec15008b07affa8e6f0af5bc3d2f039690cbf7/finders/metadata/drug-safety-updates.json#L19-L21

This is necessary because we will soon start looking up links in the publishing-api from rummager (https://github.com/alphagov/rummager/pull/632). The publishing-api therefore needs to be the source of truth for any tagging.

### DIFF
--- a/lib/tasks/send_finder_links.rake
+++ b/lib/tasks/send_finder_links.rake
@@ -1,0 +1,45 @@
+task send_finder_links: [:environment] do
+  # CMA cases.
+  # [
+  #   "/topic/competition/competition-act-cartels",
+  #   "/topic/competition/consumer-protection",
+  #   "/topic/competition/markets",
+  #   "/topic/competition/mergers",
+  #   "/topic/competition/regulatory-appeals-references"
+  # ]
+  Services.publishing_api.patch_links(
+    "fef4ac7c-024a-4943-9f19-e85a8369a1f3",
+    links: {
+      topics: [
+        "4a6f14ad-baa1-4b15-8026-8282913ef693",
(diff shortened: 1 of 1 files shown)


benhyland:
The first link should be https://github.com/alphagov/specialist-publisher/blob/059d79e9c2bac5d027edca3ffa2bfc2069a32afc/finders/metadata/medical-safety-alerts.json#L17-L20

This looks fine.
My assumptions: we run this task manually, not on deploy; and, it is easier to have looked up the ids for this change than to have written code using human readable topic names which performs the id lookup automatically. In other words, I haven't examine the deployment lifecycle or that the ids are correct!

👍

tijmenb:
Yes on the assumptions!

- Choice was between a rake task and a data migration, both of which are not ideal. At least the rake task doesn't interfere with normal development and can be removed without leaving an artefact.
- I used http://www.rubydoc.info/gems/gds-api-adapters/GdsApi/PublishingApiV2#lookup_content_ids-instance_method to fetch the content_ids.

PR #63, opened 2016-05-17
Add empty `details` to content item payload

tijmenb:
`details` will be required from https://github.com/alphagov/govuk-content-schemas/pull/310


Davidslv:
👍

PR #64, opened 2016-05-17
Set correct permissions for jenkins-schema.sh

tijmenb:
Forgot that we had to `chmod 755 jenkins-schema.sh`, otherwise this won't work:

https://ci.dev.publishing.service.gov.uk/job/govuk_content_tagger_schema_tests/4/console


benhyland:
at least we're using git so it actually has first-class understanding of the permissions flags

PR #68, opened 2016-06-06
Rake task for sending links

MatMoore:
This will be used to correct 4 content items that have organisations
in rummager but not publishing api.

https://trello.com/c/csue5VYU/646-specialist-publisher-manuals-have-missing-organisations-in-publishing-api


tijmenb:
👍

PR #75, opened 2016-07-20
Add taxon related functionality

carvil:
This PR moves taxon-related functionality from `collections-publisher` into `content-tagger`, where it belongs.

**NOTE:**
In doing so, we ported the entire commit history of those files so we can keep track of past edits. That is why you see so many commits in this PR. The only thing that matters for review purposes is the files being added to the repo.

In order to make sure these files include the history, you can open one of them and check git-blame or git-history.

It now looks like this:

screen shot 2016-07-20 at 10 39 49

Trello ticket: https://trello.com/c/merLW6kn/23-move-taxonomy-management-and-tagging-to-content-tagger

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -50,3 +50,37 @@ form label {
     display: none;
   }
 }
+
+table {
+  @extend .table;
+  @extend .table-bordered;
+  @extend .table-striped;
+}
+
+.heading-with-actions {
+  margin: 10px 0 20px;
+  overflow: hidden;
+
+  h1, h2 {
+    float: left;
+    line-height: 1.4;
+    margin: 0;
+  }
+
+  .actions {
+    float: right;
+  }
+}
+
+.actions {
+  @extend .btn-group;
+
+  a {
+    @extend .btn;
+    @extend .btn-default;
+  }
+}
+
+input[type='text'] {
+  @extend .form-control;
+}
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,4 +7,17 @@ class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
+
+private
+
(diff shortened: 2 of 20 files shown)


tijmenb:
As discussed offline: we're going to ignore some of the files so there's less history and possible confusion. Also removing the `Edit Taxonomy` permission as it's not needed for content-tagger currently.

MatMoore:
Nice work. Is the long term plan to get rid of collections publisher for taxonomy stuff and curation?

Couple of questions:

Should we do anything about the navigation bar? When the user has the permission to edit taxons, there is only one item in the menu. Should we add an item for the content tagging part.

Do we need to do cleanup to remove the "Edit taxonomy" permissions from collections publisher?

When I tested this locally I got an `undefined method `any?' for nil:NilClass` error, coming from this line of the taxon form:
`form.parent_taxons = links.parent_taxons if links.present? && links.parent_taxons.any?`

This could be a data issue with the data in my dev environment, but it shouldn't cause a 500 error.

carvil:
Thanks @MatMoore 

As far as I know, yes. The reason is that we tag content in content tagger, so taxonomy should also be in there as it is related and makes the workflow simpler (hopefully).

Tijmen mentioned we should probably delete the 'Edit Taxonomy' permissions in content tagger, so taxon will be the only navigation in the nav bar. Does that make sense?

You are completely right, we should delete the 'Edit Taxonomy' permissions in collections-publisher. There is a pending task in my trello ticket about that once everything is merged and deployed. At that point, I will open a PR to remove all taxon-related stuff (including permissions) from collections-publisher.

Regarding your error - there is a rake task to be run locally before you can use this. Apologies, I didn't copy that info from the collections-publisher's PR. The rake task is:

```
bundle exec rake publishing_api:populate_parent_taxons
```

It will copy existing `parent` links into the newly created `parent_taxons` link.

tijmenb:
@carvil should we guard against that nil-error, since it's still going to be possible to have `parent_taxons` be nil (even after rake task)|?

carvil:
I will close this one, please refer to https://github.com/alphagov/content-tagger/pull/76

carvil:
@tijmenb Yup, makes sense. I will add a commit to #76

PR #76, opened 2016-07-20
Copy taxon related functionality

carvil:
**NOTE 1:** This PR is a slimmer version of https://github.com/alphagov/content-tagger/pull/75 That PR should be disregarded.

------

This PR moves taxon-related functionality from `collections-publisher` into `content-tagger`, where it belongs.

**NOTE 2:**
In doing so, we ported the entire commit history of those files so we can keep track of past edits. That is why you see so many commits in this PR. The only thing that matters for review purposes is the files being added to the repo.

In order to make sure these files include the history, you can open one of them and check git-blame or git-history.

It now looks like this:

screen shot 2016-07-20 at 10 39 49

Trello ticket: https://trello.com/c/merLW6kn/23-move-taxonomy-management-and-tagging-to-content-tagger


carvil:
@tijmenb I managed to bring down the commits to the only ones we are interested in, please have a look and let me know.

@MatMoore I've also added a link to the nav bar for the Tags, good catch 👍 

After merging the code from collections-publisher I added more changes. Those commits start from c37295a

Davidslv:
Great job keeping the history from `collections-publisher` and breaking it down in small commits, very neat!

PR #77, opened 2016-07-20
Rename link type: alpha_taxons to taxons

Davidslv:
When we started experimenting with new taxonomy we used `alpha_taxonomy` as the link type.
Now that we're no longer in alpha we'd like to be consistent in using the link type.

Trello: https://trello.com/c/tt0UdAV3/35-use-taxons-as-link-type-everywhere
2349d27


Davidslv:
This PR needs to be revisited after #76 is merged because of: https://github.com/alphagov/content-tagger/pull/76/files#diff-63a8e971bcba960dd75982227cab40e0R67

Basically this depends on #76 so we just do this work once.

jackscotti:
`ActionView::MissingTemplate:Missing partial content/_form_for_taxons, application/_form_for_taxons with {:locale=>[:en], :formats=>[:html], :variants=>[], :handlers=>[:erb, :builder, :raw, :ruby]}
`

app/views/content/_form_for_taxons.html.erb
```diff
@@ -0,0 +1,7 @@
+<h3>Taxonomy</h3>
+
+<%= f.select :taxons,
+    linkables.taxons,
+    {},
+    { multiple: true, class: "select2", data: { placeholder: "Choose taxons..." } } %>
```
tijmenb: Could we make the explain text:

> This taxonomy is currently being developed by the Finding Things team.

Also, we should probably change "Choose taxons" to "Choose one..." or something like that. We shouldn't use "taxon" in the UI (all tough we don't have a better name yet - https://trello.com/c/KoydaDC8).

Davidslv:
@tijmenb I think I've addressed your comment

tijmenb:
👍

PR #79, opened 2016-07-21
Fix status message sent to GitHub

tijmenb:
This message will be sent to GitHub when testing a branch. This copy-paste error causes GitHub to override the status collections-publisher.


carvil:
👍

tijmenb:
screen shot 2016-07-21 at 16 36 04

PR #80, opened 2016-07-21
Use jenkins.sh from ci-scripts

tijmenb:
This is the first usage of the govuk-ci-scripts repo. The idea is that instead of copy-pasting jenkins scripts, we use the file from GitHub directly. This will prevent duplication and allow us to update the build scripts more easily.

We do this now specifically because we're having trouble hooking up content-tagger to the govuk-content-schemas schema tests.


carvil:
👍

PR #82, opened 2016-07-22
Clone Jenkins CI script and run locally

carvil:
This commit replaces `curl ... | bash` with cloning the repo and running the script locally.

What do you think, @tijmenb ?


carvil:
Closing this for now until we know from infrastructure about what's the preferred way to do this (cc @tijmenb )

PR #83, opened 2016-07-22
Fix publishing_app name on the Taxon Presenter

carvil:
This code was originally copied from `collections-publisher`. Back then, the `publishing_app` was correct, but not anymore. This commit fixes it.

### DIFF
--- a/spec/presenters/taxon_presenter_spec.rb
+++ b/spec/presenters/taxon_presenter_spec.rb
@@ -2,15 +2,20 @@ require 'rails_helper'
 
 RSpec.describe TaxonPresenter do
   describe "#payload" do
-    it "generates a valid payload" do
-      presenter = TaxonPresenter.new(
+    let(:presenter) do
+      TaxonPresenter.new(
         title: "My Title",
         base_path: "/taxons/my-taxon"
       )
+    end
+    let(:payload) { presenter.payload }
 
-      payload = presenter.payload
-
+    it "generates a valid payload" do
       expect(payload).to be_valid_against_schema('taxon')
+    end
+
+    it 'assigns the expected rendering app' do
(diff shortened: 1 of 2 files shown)


spec/presenters/taxon_presenter_spec.rb
```diff
@@ -2,15 +2,20 @@
...
   describe "#payload" do
-    it "generates a valid payload" do
-      presenter = TaxonPresenter.new(
+    let(:presenter) do
+      TaxonPresenter.new(
         title: "My Title",
         base_path: "/taxons/my-taxon"
       )
+    end
+    subject { presenter.payload }
```
tijmenb: I think `subject` only introduces mental overhead when reading the tests.  

```
expect(subject).to be_valid_against_schema('taxon')
```

vs

```
expect(payload).to be_valid_against_schema('taxon')
```

WDYT?

carvil: I can see why the second version is more readable in isolation. My thinking was that in this case, we already have a `describe "#payload"` block, so `subject` is in that context (i.e. we are testing the payload).

I think it comes down to how we read the spec, though - either from top/bottom or looking at the expectations first. 

That said, I'm happy to drop `subject` in favour of `payload` if it's not helping.

Davidslv: When I read subject I thought of the class `TaxonPresenter` not the payload, I think it would be better to explicitly say payload, if this had been a larger file I would be mislead.

carvil: Ok, I will make the change now!

carvil:
Thanks for your comments @tijmenb and @Davidslv , I've replaced `subject` with `payload`.

Davidslv:
Awesome 👍

PR #84, opened 2016-07-22
[REBASE FIXUPS] Self-serve Bulk Tagging

mobaig:
A general purpose interface for bulk tagging of GOV.UK content via
mapping spreadsheets stored on Google Drive. "General purpose" in the
sense that any piece of content can be tagged with any link type, as
long as the required fields are found in the source spreadsheet.

Includes:

- A BulkTagging model storing the public URLs of the spreadsheets.
- A TagMapping model capturing all of the fields required to tag a piece
  of content.
- Rudimentary error handling when fetching tags from the spreadsheet,
  storing them locally, and sending them to the publishing API.
- A series of screens for creating a new BulkTagging, previewing its
  TagMappings, refreshing these TagMappings following an update to the
  spreadsheet, and sending these tags to the publishing API.

trello https://trello.com/c/yndptBXH/22-allow-self-service-bulk-tagging-with-google-docs

### DIFF
--- a/app/models/tag_mapping.rb
+++ b/app/models/tag_mapping.rb
@@ -1,0 +1,5 @@
+class TagMapping < ActiveRecord::Base
+  belongs_to :tagging_spreadsheet
+  scope :by_content_base_path, -> { order(content_base_path: :asc) }
+  scope :by_link_title, -> { order(link_title: :asc) }
+end
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,6 +1,7 @@
 <% content_for :head do %>
   <%= stylesheet_link_tag "application", media: "all" %>
   <%= javascript_include_tag 'application' %>
+  <%= csrf_meta_tags %>
 <% end %>
 
 <% content_for :navbar_items do %>
@@ -9,6 +10,9 @@
   </li>
   <li class='<%= active_navigation_item == 'taxons' ? 'active' : nil %>'>
     <%= link_to 'Taxons', taxons_path %>
+  </li>
+  <li class='<%= active_navigation_item == 'tagging_spreadsheets' ? 'active' : nil %>'>
+    <%= link_to 'Bulk Tagging', tagging_spreadsheets_path %>
   </li>
 <% end %>
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -14,6 +14,11 @@ Rails.application.routes.draw do
     post '/content/taggings', action: :update_links, as: :content_update_links
   end
 
+  resources :tagging_spreadsheets, except: [:update, :edit]  do
+    post 'refetch'
+    post 'publish_tags'
+  end
+
   get '/healthcheck', to: proc { [200, {}, ['OK']] }
 
   mount GovukAdminTemplate::Engine, at: "/style-guide" if Rails.env.development?
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160727095240) do
+ActiveRecord::Schema.define(version: 20160727101957) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -29,9 +29,12 @@ ActiveRecord::Schema.define(version: 20160727095240) do
   add_index "tag_mappings", ["tagging_spreadsheet_id"], name: "index_tag_mappings_on_tagging_spreadsheet_id", using: :btree
 
   create_table "tagging_spreadsheets", force: :cascade do |t|
-    t.string   "url",        null: false
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
+    t.string   "url",               null: false
+    t.datetime "created_at",        null: false
+    t.datetime "updated_at",        null: false
+    t.string   "added_by"
+    t.string   "last_published_by"
+    t.datetime "last_published_at"
   end
 
   create_table "users", force: :cascade do |t|
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,9 +11,28 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20151115132026) do
+ActiveRecord::Schema.define(version: 20160727095240) do
+
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
+
+  create_table "tag_mappings", force: :cascade do |t|
+    t.integer  "tagging_spreadsheet_id"
+    t.string   "content_base_path"
+    t.string   "link_title"
+    t.string   "link_content_id"
+    t.string   "link_type"
+    t.datetime "created_at",             null: false
+    t.datetime "updated_at",             null: false
+  end
+
+  add_index "tag_mappings", ["tagging_spreadsheet_id"], name: "index_tag_mappings_on_tagging_spreadsheet_id", using: :btree
+
+  create_table "tagging_spreadsheets", force: :cascade do |t|
+    t.string   "url",        null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
 
   create_table "users", force: :cascade do |t|
     t.string   "name"
@@ -27,4 +46,6 @@ ActiveRecord::Schema.define(version: 20151115132026) do
     t.datetime "created_at",              null: false
     t.datetime "updated_at",              null: false
   end
+
+  add_foreign_key "tag_mappings", "tagging_spreadsheets", on_delete: :cascade
 end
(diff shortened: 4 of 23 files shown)


spec/features/bulk_tagging_spec.rb
```diff
@@ -0,0 +1,127 @@
...
+    click_link "Preview taggings"
+    expect(page).to have_content "This import broke."
+  end
+
+  def given_no_tagging_data_was_found
+    stub_request(:get, google_sheet_url(key: sheet_key, gid: sheet_gid))
+      .to_return(status: 404)
+  end
+
+  def sheet_key
```
tijmenb: Perhaps make this and `sheet_gid` constants so there are only "step" methods.

app/views/bulk_taggings/index.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    <th>Google Sheet URL</th>
+    <th>Job Created</th>
+    <th>Actions</th>
+  </tr>
+  <% @bulk_taggings.each do |tagging| %>
+    <tr>
+      <td><%= tagging.mapping_url %></td>
+      <td><%= tagging.created_at %></td>
+      <td>
+        <%# <%= link_to "Preview taxons", taxon_import_path(tagging), class: "btn btn-primary" %1> %>
```
tijmenb: Does this need to be removed?

config/routes.rb
```diff
@@ -14,6 +14,11 @@
     post '/content/taggings', action: :update_links, as: :content_update_links
   end
 
+  resources :bulk_taggings do
```
tijmenb: Can we make the url `/bulk-tagging`?

app/views/layouts/application.html.erb
```diff
@@ -1,6 +1,19 @@
 <% content_for :head do %>
   <%= stylesheet_link_tag "application", media: "all" %>
   <%= javascript_include_tag 'application' %>
+  <%= csrf_meta_tags %>
+<% end %>
+
+<% content_for :navbar_items do %>
```
tijmenb: I think this needs to be merged with the `navbar_items` below.

app/views/bulk_taggings/show.html.erb
```diff
@@ -0,0 +1,43 @@
...
+  </div>
+
+<% else %>
+
+  <p>
+    <%= link_to "Create these tags", bulk_tagging_tag_imports_path(@bulk_tagging), method: :post, class: "btn btn-lg btn-success" %>
+    <%= link_to "Fetch from Google again", bulk_tagging_refetch_path(@bulk_tagging), class: "btn btn-lg btn-warning" %>
+  </p>
+
+  <table border="0" cellspacing="5" cellpadding="5">
```
tijmenb: Is `border="0" cellspacing="5" cellpadding="5"` necessary?

app/models/bulk_tagging.rb
```diff
@@ -0,0 +1,25 @@
...
+  validate :public_google_sheet_url?
+
+  has_many :tag_mappings, dependent: :destroy
+
+  scope :newest_first, -> { order(created_at: :desc) }
+
+private
+
+  def https_url?
+    unless URI.parse(mapping_url).scheme == "https"
```
tijmenb: Any particular reason for this validation?

app/models/bulk_tagging.rb
```diff
@@ -0,0 +1,25 @@
+require 'csv'
```
tijmenb: Not needed anymore I think?

app/views/bulk_taggings/new.html.erb
```diff
@@ -0,0 +1,8 @@
+<h1>Bulk Tagging</h1>
+
+<%= form_for @bulk_tagging, url: bulk_taggings_path do |form| %>
+  <div class="form-group">
+    <%= form.text_field :mapping_url, class: "form-control" %>
```
tijmenb: Can we link to a publicly available example sheet here?

app/services/publish_taggings.rb
```diff
@@ -0,0 +1,41 @@
...
+        hash[content_base_path] = {} unless hash[content_base_path].present?
+        hash[content_base_path][link_type] = hash[content_base_path].fetch(link_type, []) << link_content_id
+      end
+
+      payloads.each do |base_path, links|
+        logger.tagged("BULK-TAG") { logger.info "Patching links on #{base_path}" }
+        target_content_id = Services.publishing_api.lookup_content_id(base_path: base_path)
+        if target_content_id.blank?
+          logger.tagged("BULK-TAG") { logger.info "no content ID found for #{base_path}" }
+          return
```
tijmenb: Will this `return` from the `run` method? You mean have meant `next`.

db/migrate/20160718104316_create_bulk_taggings.rb
```diff
@@ -0,0 +1,9 @@
+class CreateBulkTaggings < ActiveRecord::Migration
+  def change
+    create_table :bulk_taggings do |t|
+      t.string :mapping_url, null: false
```
tijmenb: We could also save the authenticated user here, so we have some audibility.

tijmenb:
Looking good! As disussed offline, this needs Sidekiq to work because the publishing-api calls will take too long.

db/schema.rb
```diff
@@ -11,10 +11,40 @@
...
+    t.string   "link_title"
+    t.string   "link_content_id"
+    t.string   "link_type"
+    t.datetime "created_at",        null: false
+    t.datetime "updated_at",        null: false
+  end
+
+  add_index "tag_mappings", ["bulk_tagging_id"], name: "index_tag_mappings_on_bulk_tagging_id", using: :btree
+
+  create_table "taxon_tags", force: :cascade do |t|
```
tijmenb: This table seems not to be used?

app/models/tag_mapping.rb
```diff
@@ -0,0 +1,6 @@
+class TagMapping < ActiveRecord::Base
```
tijmenb: Wondering about the naming. Maybe `BulkTagMapping` or something? Or maybe we could put it all in a namespace?

app/controllers/bulk_taggings_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+    if @bulk_tagging.tag_mappings.count.zero?
+      @fetch_errors = FetchTaggingSheet.new(@bulk_tagging).run
+    end
+
+    @tag_mappings = @bulk_tagging.tag_mappings.by_content_base_path.by_link_title
+  end
+
+  def refetch
+    @bulk_tagging = BulkTagging.find(params[:bulk_tagging_id])
+    @bulk_tagging.tag_mappings.destroy_all
```
tijmenb: If we set the foreign key constraint to cascade delete, we can make this `delete_all`. May prevent us from having to do deletion inside a sidekiq worker.

spec/support/google_sheet_helper.rb
```diff
@@ -0,0 +1,30 @@
...
+  def google_sheet_url(key:, gid:)
+    "https://docs.google.com/spreadsheets/d/#{key}/pub?gid=#{gid}&single=true&output=tsv"
+  end
+
+  def google_sheet_fixture(extra_rows=[])
+    [
+      google_sheet_row(content_base_path: "content_base_path" , link_title: "link_title"     , link_content_id: "link_content_id"           , link_type: "link_type")     ,
+      google_sheet_row(content_base_path: "/content-1/"       , link_title: "Education"      , link_content_id: "education-content-id"      , link_type: "taxons")        ,
+      google_sheet_row(content_base_path: "/content-1/"       , link_title: "Cabinet Office" , link_content_id: "cabinet-office-content-id" , link_type: "organisations") ,
+      google_sheet_row(content_base_path: "/content-2/"       , link_title: "Early Years"    , link_content_id: "early-years-content-id"    , link_type: "taxons")        ,
```
tijmenb: Can you add a test case for adding two links of the same type to a single item? Pretty sure it works, but just to be safe.

app/views/bulk_taggings/show.html.erb
mobaig: No - cruft from a vim snippet. Removing...

app/models/bulk_tagging.rb
mobaig: Went in early on as a cursory validation that the mapping_url was actually a proper url. The validation below this one (checking that it matches the structure of a published google sheet url) is probably sufficient. Removing this one, it's just extra code with minimal value.

app/views/bulk_taggings/new.html.erb
mobaig: Sure, good shout. Adding...

app/models/tag_mapping.rb
mobaig: Reworked names a bit.

A TaggingSpreadsheet has many TagMappings. Keeping the latter because its reasonably clear. 

Only stuff in a namespace is the service objects.

app/controllers/bulk_taggings_controller.rb
mobaig: yeh, good idea. updating..

carvil:
👍

PR #88, opened 2016-07-28
Remove one-off rake tasks

tijmenb:
This PR removes a number of one-off rake tasks that will no longer be used.


tijmenb:
cc @carvil

carvil:
👍

PR #89, opened 2016-07-28
Various UI tweaks

tijmenb:
First pass at some UI tweaks for the import functionality. Hopefully we'll start some user research on this functionality soon so we can nail down some the language ("bulk tagging" vs "import").

## Before

screen shot 2016-07-28 at 18 01 08
screen shot 2016-07-28 at 18 01 12
screen shot 2016-07-28 at 18 01 17

## After

screen shot 2016-07-28 at 18 01 26
screen shot 2016-07-28 at 18 01 29
screen shot 2016-07-28 at 18 01 34

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -7,10 +7,6 @@
 
 @import 'govuk_admin_template';
 
-.btn.bulkTaggingButton {
-  margin-bottom: 1rem;
-}
-
 .select2 {
   width: 100%;
 }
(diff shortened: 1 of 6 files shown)


tijmenb:
cc @mobaig

jackscotti:
What do you mean with 'fixup' in the commit titles?

tijmenb:
That the commit "upfixes", or "fixes up", something. But now I'm not sure if that's English.

jackscotti:
The only place where I've seen the word 'fixup' is the git prompt when rebasing. And there it means to squash the commit onto the previous one, this is why its usage confuses me.

tijmenb:
Yeah, you're right - https://en.wikipedia.org/wiki/Fix-up_(disambiguation). My usage doesn't make much sense. I'll fix that right up.

mobaig:
http://www.merriam-webster.com/dictionary/fix%20up

mobaig:
👍

PR #90, opened 2016-07-29
Use publishing-api for tagged content

tijmenb:
Use publishing-api endpoint for linked items. Publisher apps should never use the content-store, because it's not synchronous and could cause confusion for users. In this case, taggings would not immediately show up because publishing-api sends the tags to the content-store via a Sidekiq worker.

This commit replaces the call to content-store with a call to the publishing-api, so we only have one dependency.

Also does some other cleanup. See commits for details.


mobaig:
👍

PR #91, opened 2016-07-29
Bugfix - handle taxon's unprocessable entity errors

carvil:
When the publishing API rejects a taxon with some validation issues, we should display the error to the user rather than a 500 error page.

Furthermore, we should also allow the creation of taxons with the same title (but different base paths).

This PR solves both problems.

More information here:
https://trello.com/c/gwG1mEop/49-handle-existing-base-paths-when-creating-taxons


tijmenb:
👍 excellent

PR #92, opened 2016-07-29
Add link from tagged item to tagger

tijmenb:
Since we now have all the functionality in this app, we can create cross-links. This commit adds a link from the "Tagged content" page (content tagged to a certain taxon) to the Tagging page (where the user can update the tags).

## Before

screen shot 2016-07-29 at 14 59 12

## After

screen shot 2016-07-29 at 14 59 04


carvil:
👍

PR #93, opened 2016-07-29
Refactor taxon form

carvil:
This is an attempt at removing some responsibilities from the `TaxonForm` class.

This class was being used in views and it was also responsible for building itself (via API calls) and create itself (via instantiating a presenter with some details and calling several API endpoints).

This PR adds the following:
- A builder class, which knows how to build a `TaxonForm` object from a `content_id`;
- A publisher class, which knows how to present and publish a `TaxonForm` object;
- A new interface to the `TaxonPresenter` class, passing on the `TaxonForm` object rather than just the `base_path` and the `title`.

Is this an improvement?


tijmenb:
👍 🎉

PR #94, opened 2016-08-02
Tag importer - add workers and more info on preview screen

mobaig:
The tag importer times out in the absence of out-of-band workers to offload patch_links calls to. This adds a sidekiq worker to handle this processing, along with frontend changes indicating the progress of a given import. See commits for more info.

https://trello.com/c/yndptBXH/22-allow-self-service-tag-importer-with-google-docs

screen shot 2016-08-02 at 16 16 28


app/views/tagging_spreadsheets/show.html.erb
```diff
@@ -19,20 +15,36 @@
...
-  </table>
+      <% @tag_mappings.each do |tag| %>
+        <tr>
+          <td><%= link_to tag.content_base_path, website_url(tag.content_base_path) %></td>
+          <td><%= link_to tag.link_title, content_path(tag.link_content_id) %> (<%= tag.link_type %>)</td>
+          <td class="tag-mapping-status">
+            <% if tag.publish_confirmed? %>
+              <span class="label label-success">Yes</span>
+            <% else %>
+              <span class="label label-danger">No</span>
```
tijmenb: Red means it's waiting for the publishing right? Perhaps `label-default` would be better. `label-danger` can then be used if the job can't be completed/errors.

tijmenb:
👍 cool

app/workers/publish_links_worker.rb
```diff
@@ -0,0 +1,15 @@
+class PublishLinksWorker
+  include Sidekiq::Worker
+
+  sidekiq_options retry: 5
```
carvil: What do you think of implementing the hook `sidekiq_retries_exhausted` and notity Airbrake with the job details? The hook gets called when we exhausted all retries, which could be helpful for debugging purposes.

I'm mentioning it because it saved me a few times in the past!

carvil: I just realised now that we already have error handling going on here: https://github.com/alphagov/govuk_sidekiq/blob/master/lib/govuk_sidekiq/sidekiq_initializer.rb#L18 so 👍

app/views/tagging_spreadsheets/show.html.erb
mobaig: Sure, updated.

PR #95, opened 2016-08-03
Add tag importer progress bar and some additional worker behaviour

mobaig:
See commits for details.

### DIFF
--- a/app/assets/javascripts/import-progress.js
+++ b/app/assets/javascripts/import-progress.js
@@ -1,0 +1,23 @@
+$(function updateImportProgress() {
+  var $importProgress = $('.js-import-progress');
+  if ($importProgress.length == 0) { return; }
+
+  var currentProgress = $importProgress.find('.progress-bar').attr("aria-valuenow");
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/import-progress.js
```diff
@@ -0,0 +1,23 @@
+$(function updateImportProgress() {
```
tijmenb: We can use the framework in govuk_admin_template for this, for consistency with other publishing apps:

https://github.com/alphagov/govuk_admin_template/blob/master/JAVASCRIPT.md

mobaig: Hmm...coulda sworn I tried this and found it to be a bit broken. Decided to drop it on the basis that the function is pretty simple and I couldn't find a working example in collections-publisher or service manual. Should have checked the publisher app!

Anyway, second time's a charm.

tijmenb:
🎉 very cool!

tijmenb:
BTW we should revisit this when we've upgraded to Rails 5. Super curious if ActionCable is easy enough to be used for this.

PR #96, opened 2016-08-03
Rename Bulk Tagger to Tag Importer

mobaig:
We've renamed this feature to avoid confusion with the concept of 'bulk
tagging' a content item with multiple tags manually via the UI. The
spreadsheet-driven tagging is now referred to as the Tag Importer.


rubenarakelyan:
👍

PR #98, opened 2016-08-03
Remove AlphaTaxonomy bulk tagging scripts

mobaig:
The Tag Importer UI now makes these rake tasks and scripts largely
redundant - they're fiddly to run and offer no particular advantage
over the browser-based option.


rubenarakelyan:
👍

PR #104, opened 2016-08-09
Highlight the correct navigation link on content pages

carvil:
When we go to pages in `/content`, we should highlight the navigation link
`Tags` so we know where we are in the application. This commit makes sure that
happens.

The naming is still not coherent between the URLs and the navigation links. There are tasks in trello to fix that, so I will do that in a separate PR.

Trello: https://trello.com/c/RrWiwHb4/81-content-tagger-when-we-navigate-to-a-content-page-the-nav-link-tab-isn-t-highlighted

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -5,7 +5,7 @@
 <% end %>
 
 <% content_for :navbar_items do %>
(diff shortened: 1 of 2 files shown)


rubenarakelyan:
👍

PR #105, opened 2016-08-09
Make initial tagging import action asynchronous

tijmenb:
At the moment, when the user creates a new spreadsheet import by submitting a URL, the application will start the initial import step (parsing the CSV, creating records for each row) in the request cycle.

This is slow and does not allow us to add more expensive validation to this step.

This commit is the minimum we have to do to move the initial import action to a Sidekiq worker.

- It adds a very crude state thing, which could perhaps be replaced by a real state machine.
- The workflow is a bit clunky now. For example, it's not clear to the user what state the spreadsheet is in. We'll address this in follow up PRs (https://trello.com/c/7tcrUgcz).

Trello: https://trello.com/c/xi83pGTZ

### DIFF
--- a/app/models/tagging_spreadsheet.rb
+++ b/app/models/tagging_spreadsheet.rb
@@ -1,5 +1,8 @@
 class TaggingSpreadsheet < ActiveRecord::Base
   validates :url, presence: true
+  validates_presence_of :state
+  validates_inclusion_of :state, in: %w(uploaded errored ready_to_import imported)
+
   has_many :tag_mappings, dependent: :delete_all
   scope :newest_first, -> { order(created_at: :desc) }
 end
--- a/app/views/tagging_spreadsheets/_import_preview.html.erb
+++ b/app/views/tagging_spreadsheets/_import_preview.html.erb
@@ -2,7 +2,9 @@
   <br/>
 
   <div data-module="import-progress">
-    <%= render partial: "import_progress_bar", locals: { tag_mappings: tag_mappings } %>
+    <% if tag_mappings.any? %>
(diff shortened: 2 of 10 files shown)


app/views/tagging_spreadsheets/_import_preview.html.erb
```diff
@@ -2,7 +2,9 @@
   <br/>
 
   <div data-module="import-progress">
-    <%= render partial: "import_progress_bar", locals: { tag_mappings: tag_mappings } %>
+    <% if tag_mappings.any? %>
```
carvil: Do you think we need an `else` with a message to the user?

app/models/tagging_spreadsheet.rb
```diff
@@ -1,5 +1,8 @@
 class TaggingSpreadsheet < ActiveRecord::Base
   validates :url, presence: true
+  validates_presence_of :state
+  validates_inclusion_of :state, in: %w(uploaded errored ready_to_import imported)
```
carvil: Nice! If you wanted to use a state machine, I've tried this in the past and really liked it https://github.com/aasm/aasm

app/views/tagging_spreadsheets/_import_preview.html.erb
tijmenb: Yeah, good point, I'll do that in a follow up PR to address the workflow UI (making the state of the import clearer).

carvil: Sounds good 👍

app/models/tagging_spreadsheet.rb
carvil: 👍

PR #108, opened 2016-08-10
Mark tagging spreadsheets as deleted

carvil:
Instead of deleting TaggingSpreadsheets, we should mark them as deleted and keep
a record of them in the backend for audit purposes. As they perform actions on
content items, we should probably keep a record of them in case something goes
wrong.

This change makes sure we don't perform a hard-delete on those spreadsheets. We
instead mark them as deleted and remove those ones from the index page.

Trello: https://trello.com/c/dAR3TgV2/84-instead-of-allowing-a-user-to-delete-a-tagging-spreadsheet-we-should-probably-just-mark-it-as-deleted-in-order-to-keep-a-record-

Part of a bigger task: https://trello.com/c/7tcrUgcz/66-initial-round-of-improvements-to-error-handling-in-tag-importer

### DIFF
--- a/app/controllers/tagging_spreadsheets_controller.rb
+++ b/app/controllers/tagging_spreadsheets_controller.rb
@@ -1,6 +1,6 @@
 class TaggingSpreadsheetsController < ApplicationController
   def index
-    @tagging_spreadsheets = TaggingSpreadsheet.all.newest_first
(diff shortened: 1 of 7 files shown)


rubenarakelyan:
👍

PR #112, opened 2016-08-10
Validate google spreadsheet url

carvil:
Validate the URL provided to `TaggingSpreadsheet` models in order to make sure only valid Google Spreadsheet URLs are given by the user.

There are 3 validations:
- the hostname (it should be `docs.google.com`)
- the path (it needs to match `/spreadsheets/d/<a key>/pub`
- the params (it needs to include a Google ID param called `gid`)

This PR also adds a factory for building `TaggingSpreadsheet` objects in the tests with the correct URL structure.

Trello: https://trello.com/c/AGgxBxcJ/87-validate-the-data-given-to-the-tag-importer-it-should-be-a-google-spreadsheet-url-currently-it-accepts-anything

Part of: https://trello.com/c/7tcrUgcz/66-initial-round-of-improvements-to-error-handling-in-tag-importer

### DIFF
--- a/app/validators/google_url_validator.rb
+++ b/app/validators/google_url_validator.rb
@@ -1,0 +1,34 @@
+class GoogleUrlValidator < ActiveModel::Validator
+  def validate(record)
+    uri = URI.parse(record.url)
+
+    validate_host(uri, record)
+    validate_path(uri, record)
+    validate_parameters(uri, record)
+  end
+
+private
+
+  def validate_host(uri, record)
+    unless uri.host == "docs.google.com"
+      record.errors[:url] << I18n.t('errors.invalid_hostname')
+    end
+  end
+
+  def validate_path(uri, record)
+    unless uri.path =~ %r{spreadsheets\/d\/.+\/pub}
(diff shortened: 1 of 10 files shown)


app/validators/google_url_validator.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+  end
+
+  def validate_path(uri, record)
+    unless uri.path =~ %r{spreadsheets\/d\/.+\/pub}
+      record.errors[:url] << I18n.t('errors.invalid_path')
+    end
+  end
+
+  def validate_parameters(uri, record)
```
mobaig: Do you think it's worth validating somewhere that `output=tsv` in the params? The CSV parsing logic in `app/services/tag_importer/fetch_remote_data.rb` expects tab-separated values.

carvil: Ah, well spotted @mobaig I will add that! Thanks

carvil: @mobaig : I added that validation below, is that correct?

mobaig:
👍

PR #113, opened 2016-08-10
A TaxonPayloadBuilder builds a payload

carvil:
This commit fixes:

- naming issues, where we were calling the TaxonPayloadBuilder a presenter;
- the interface with TaxonPayloadBuilder - instead of having a `payload` method,
  we let the `build` method return a payload, as the class name suggests.


mobaig:
👍

PR #116, opened 2016-08-10
Display state information on tag importer

carvil:
For each of the spreadsheet taggings in the system, we should display the state
information so we are aware of the progress of each of the imports.

In case there is an error, we should also display an error message as a tooltip
in the label.

This PR also adds an experimental Facade. It works pretty much like a decorator, but it has no view context, so it can't access view helpers. Its responsibility is to hold logic specific to the object in the context of its presentation in the view. Are people ok with this approach? 

This change looks like this:

1) With an error:

screen shot 2016-08-10 at 17 12 25

2) With success:

screen shot 2016-08-10 at 17 13 51

3) With other states (`ready_to_import` and `uploaded`):

screen shot 2016-08-10 at 17 11 41

### DIFF
--- a/app/controllers/tagging_spreadsheets_controller.rb
+++ b/app/controllers/tagging_spreadsheets_controller.rb
@@ -1,6 +1,6 @@
 class TaggingSpreadsheetsController < ApplicationController
   def index
-    @tagging_spreadsheets = TaggingSpreadsheet.active.newest_first.includes(:added_by)
+    render :index, locals: { tagging_spreadsheets: presented_tagging_spreadsheets }
   end
 
   def new
@@ -54,6 +54,16 @@ class TaggingSpreadsheetsController < ApplicationController
   end
 
 private
+
+  def tagging_spreadsheets
+    TaggingSpreadsheet.active.newest_first.includes(:added_by)
+  end
+
+  def presented_tagging_spreadsheets
+    tagging_spreadsheets.map do |tagging_spreadsheet|
+      TaggingSpreadsheetPresenter.new(tagging_spreadsheet)
+    end
+  end
 
   def tagging_spreadsheet_params
     params.require(:tagging_spreadsheet).permit(:url, :description)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -6,4 +6,15 @@ module ApplicationHelper
       title
     end
   end
+
+  def time_tag_for(date)
+    data_attributes = {
+      'toggle': 'tooltip',
+      'original-title': date,
+    }
+
+    content_tag :time, data: data_attributes do
+      distance_of_time_in_words_to_now(date)
+    end
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,2 +1,9 @@
 module ApplicationHelper
+  def state_label_for(label_type:, title:, data_attributes:)
+    css_class = ['label', label_type].join(' ')
+
+    content_tag :span, class: css_class, data: data_attributes do
+      title
+    end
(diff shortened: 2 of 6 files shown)


carvil:
With an error, this happens:

screen shot 2016-08-10 at 17 21 44

tijmenb:
Cool, very useful!

I am having second thoughts about the Facade. Perhaps it seems too much for the problem we're solving? In collections-publisher the same effect is achieved by having this helper:

https://github.com/alphagov/collections-publisher/blob/master/app/helpers/status_helper.rb#L2-L10

This isn't _super_ nice, but it's fairly easy to follow. Thoughts?

Davidslv:
I quite like this approach, looks really neat.

carvil:
@tijmenb I see what you mean and I wasn't too sure as well.

However, I would argue that rails helpers are more of an anti-pattern. In that case you mentioned, we would have a `status` method available in every view, but that method is specific to an object with a state.

I've done that myself with the `TaggingSpreadsheetHelper` and I regret it I think. That method I created called `state_label_for` isn't generic as-is, but if we can make it generic enough (by receiving the `label_type` and `data_attributes` as params), then it probably belongs to a helper. But even that I wouldn't have it in a helper.

The reason I think rails helpers are an anti-pattern is that every single one of those methods ends up polluting every view in the system. They are harder to test and not scoped to any particular model/PORO. I think that applies to many ruby modules.

I think there are 2 good alternatives to these helpers, though:

- having view model objects (e.g. decorators like this facade or something more off the shelf like [draper](https://github.com/drapergem/draper)/ These objects may or may not have view helpers in them (for example, draper uses view helpers via the method `h`, isolating those methods neatly).
- introducing the concept of cells by using a gem called [cell](https://github.com/apotonick/cells). These are basically small components we can render anywhere, but they take the form of an object we can call things on (and test in isolation).

I guess what I'm trying to avoid is:
- many helper methods not really connected with a particular object
- complex views with logic in them, making them harder to test (and therefore relying heavily on feature tests rather than unit tests)

What do you think?

app/facades/tagging_spreadsheet_facade.rb
```diff
@@ -0,0 +1,21 @@
+class TaggingSpreadsheetFacade < SimpleDelegator
```
mobaig: Out of interest - why have we opted to call this a 'facade' when all of these methods are related to view-layer presentation? Seems like `presenter` would be a more fitting, less obscure naming choice.

carvil: Sorry, this came out of an offline conversation with Tijmen. This is a presenter, or decorator (or a facade). I avoided the word presenter because we use that in other places in the context of serialisation (building payloads). But I'm happy to change this name to presenter if that's clearer!

mobaig: There's certainly similarity between facades and presenters. While both are a mechanism for providing a simpler API for underlying objects, I think it's more typical for the latter name to be used when the context is presentational. Specifically - when creating some kind of response at the boundaries of a system for external clients, whether that's forming a JSON payload or rendering ERB. I see the clients of a facade as being internal.

We're deep in bike-shedding territory at this point, but I think the name 'facade' raises more questions than it answers in a system of this size.

carvil: That makes sense to me, I will rename it to a presenter to make it clearer.

carvil:
@tijmenb do you have more thoughts on this PR? I would like to re-use some of this in my current branch and I was wondering if you're OK with it.

tijmenb:
Yeah, let's go. We'll do some philosophising/bikeshedding about this later.

PR #107, opened 2016-08-09
Switch content tagger to use simple form medium

Davidslv:
Trello: https://trello.com/c/UZtlasW8/69-switch-content-tagger-to-use-simple-form-medium

PR #119, opened 2016-08-11
Fix select box for taxons

tijmenb:
The source of the select box was inadvertently changed to topics in https://github.com/alphagov/content-tagger/pull/107. Adds a test that would've prevented it.


carvil:
👍

PR #120, opened 2016-08-11
Move tagging into single controller

tijmenb:
This merges the `LookupController` and the `ContentController` into a single `TaggingsController`.


carvil:
🚢

PR #121, opened 2016-08-11
Make sure we don't put invalid data into tagmappings table

tijmenb:
This adds NOT NULL constraints to a bunch of columns that should never have null-values. Makes it easier to spot mistakes.

https://trello.com/c/7tcrUgcz


carvil:
:shipit:

PR #123, opened 2016-08-11
Add proper flash messages for updating actions

tijmenb:
See individual commits for the details.

https://trello.com/c/7tcrUgcz

### DIFF
--- a/app/controllers/tagging_spreadsheets_controller.rb
+++ b/app/controllers/tagging_spreadsheets_controller.rb
@@ -15,7 +15,7 @@ class TaggingSpreadsheetsController < ApplicationController
     if tagging_spreadsheet.valid?
       tagging_spreadsheet.save!
       InitialTaggingImport.perform_async(tagging_spreadsheet.id)
-      redirect_to tagging_spreadsheet, success: "Spreadsheet uploaded. It will be imported within a few minutes."
+      redirect_to tagging_spreadsheet, success: I18n.t('tag_import.import_created')
     else
       @tagging_spreadsheet = tagging_spreadsheet
       render :new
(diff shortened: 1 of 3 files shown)


app/controllers/tagging_spreadsheets_controller.rb
```diff
@@ -15,7 +15,7 @@ def create
     if tagging_spreadsheet.valid?
       tagging_spreadsheet.save!
       InitialTaggingImport.perform_async(tagging_spreadsheet.id)
-      redirect_to tagging_spreadsheets_path
+      redirect_to tagging_spreadsheet, success: "Spreadsheet uploaded. It will be imported within a few minutes."
```
carvil: What do you think of moving strings like these to I18n (not for translation purposes, just to avoid having long strings)?

Davidslv: I think that's one more file to maintain and do search/replace. I don't think we are getting a big benefit from it since we are not translating content.

carvil: Since we already use that pattern for error messages (check config/locales/en.yml), I thought it would be nice to continue doing that. It also improves the readability of the code and makes it easier to add longer and more self a explanatory messages. Just a thought. Though.

Davidslv: Ah! I wasn't aware of it. Don't mind what I said :)

carvil: Sorry, David, I wrote that message from my phone and the autocorrect didn't really help much :)

carvil:
👍  👍  👍

tijmenb:
@carvil good to merge?

carvil:
Yup was just waiting on travis

PR #126, opened 2016-08-12
Add table search for all tables in content-tagger

Davidslv:
Trello: https://trello.com/c/AP70lKei/94-add-table-search-for-all-tables-in-content-tagger

## Look & Feel

screen shot 2016-08-12 at 15 08 27

screen shot 2016-08-12 at 14 29 17
screen shot 2016-08-12 at 15 05 11
screen shot 2016-08-12 at 15 07 17

### DIFF
--- a/app/views/tagging_spreadsheets/index.html.erb
+++ b/app/views/tagging_spreadsheets/index.html.erb
@@ -2,7 +2,7 @@
   <%= link_to 'Upload spreadsheet', new_tagging_spreadsheet_path, class: 'btn btn-lg btn-default' %>
 <% end %>
 
-<table>
+<table class="table queries-list table-bordered" data-module="filterable-table">
   <thead>
     <tr>
       <th>Google Sheet URL</th>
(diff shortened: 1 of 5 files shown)


carvil:
Nice one @Davidslv !

I think there is only one table missing [here](https://github.com/alphagov/content-tagger/blob/master/app/views/tagging_spreadsheets/_import_preview.html.erb#L10). In order to see it you will have to import a spreadsheet first. Please have a look into `content-tagger` in integration for an example spreadsheet to use locally. After importing it, click on `Preview` to see the table of `tag_mapping` associations.

On your last screenshot, do you think the search row could be stretched to use the entire width of the table?

app/views/tagging_spreadsheets/index.html.erb
```diff
@@ -12,6 +12,7 @@
       <th></th>
       <th></th>
     </tr>
+    <%= render partial: 'shared/table_filter' %>
```
jackscotti: Shouldn't `'shared/table_filter'` be 'shared/_table_filter'?

tijmenb: This is the correct way to call partials: http://guides.rubyonrails.org/layouts_and_rendering.html#using-partials

Davidslv:
I've updated the PR, thanks @carvil

PR #127, opened 2016-08-15
Remove `state` column from schema.db

carvil:
`TagMapping` records don't have a state (yet); It looks like this field ended up in the schema.rb even though there isn't a migration associated with it.

I need to create this field in another PR and found the issue there.


tijmenb:
Sorry about that, that was probably due to me running a migration in a branch and not cleaning up.

carvil:
No worries, thanks for merging. I got confused with the migration errors :S then I realised what was wrong

PR #128, opened 2016-08-15
Allow taxons to be deleted

Davidslv:
Trello: https://trello.com/c/vBJSZakA/93-allow-taxons-to-be-deleted

Flow:

When the user presses the `Delete` button, and the taxon is a Parent taxon, the children of that taxon no longer have that taxon as parent. Also documents with such taxon no longer will have the taxon tagged to them.

This is all possible by requesting Publishing API to unpublish a given taxon, with the type "gone"

### Look & Feel

screen shot 2016-08-16 at 15 46 09

### DIFF
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -48,7 +48,21 @@ class TaxonsController < ApplicationController
     redirect_to taxons_path
   end
 
+  def destroy
+    response_code = Services.publishing_api.unpublish(params[:id], type: "gone").code
+
+    redirect_to taxons_path, flash: destroy_flash_message(response_code)
+  end
+
 private
+
+  def destroy_flash_message(response_code)
+    if response_code == 200
+      { success: I18n.t('messages.controller.taxons.success') }
+    else
+      { alert: I18n.t('messages.controller.taxons.alert') }
+    end
+  end
 
   def taxons_for_select
     taxon_fetcher.taxons_for_select
--- a/app/views/taxons/index.html.erb
+++ b/app/views/taxons/index.html.erb
@@ -9,6 +9,7 @@
       <th>Content ID</th>
       <th></th>
       <th></th>
+      <th></th>
     </tr>
 
     <%= render partial: 'shared/table_filter' %>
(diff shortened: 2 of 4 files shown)


app/controllers/taxons_controller.rb
```diff
@@ -48,6 +48,18 @@ def update
     redirect_to taxons_path
   end
 
+  def destroy
```
mobaig: Is this already tested somewhere? If not, can we add one?

app/controllers/taxons_controller.rb
```diff
@@ -48,6 +48,18 @@ def update
...
+    resp = Services.publishing_api.unpublish(params[:id], type: "gone")
```
mobaig: Rather we didn't have the abbreviated variable name here. Reckon you can you just be rid of it by renaming it to `response_code` and moving the `.code` call below to this line.

app/controllers/taxons_controller.rb
```diff
@@ -48,6 +48,18 @@ def update
...
+
+    message = if resp.code == 200
```
mobaig: Pretty sure you'll get the linter complaining here about the alignment and indentation of the if block.

You can resolve this by either wrapping the whole thing in a begin/end, or by lining up the `else` and `end` with the `if`.  I think the latter is ugly as hell but either would do the job.

jackscotti:
Did you get input from a designer on the button colours? I have never seen the light blue being used on GOV.UK. @alextea

app/views/taxons/index.html.erb
```diff
@@ -19,8 +18,18 @@
       <tr>
         <td><%= taxon['title'] %></td>
         <td><%= taxon['content_id'] %></td>
-        <td><%= link_to 'View tagged content', taxon_path(taxon['content_id']), class: 'view-tagged-content' %></td>
-        <td><%= link_to 'Edit taxon', edit_taxon_path(taxon['content_id']) %></td>
+        <td>
+          <%= link_to 'View', taxon_path(taxon['content_id']),
+            class: 'view-tagged-content btn btn-primary' %>
+
+          <%= link_to 'Edit', edit_taxon_path(taxon['content_id']),
```
tijmenb: "View" and "Edit" are links to pages, so they need to look like links ("Delete" is an action, so it's good that it's a button).

app/controllers/taxons_controller.rb
```diff
@@ -48,6 +48,18 @@ def update
...
+      { success: 'You have successfully deleted the taxon' }
```
carvil: It would be nice to move those to I18n as well (both the success and the alert)

Davidslv:
@tijmenb, @jackscotti, @alextea  I've reverted the buttons on View & Edit, they remain as links
@carvil I've added `I18n` translations
@mobaig I've created a private method inside the controller instead

Thanks!

app/controllers/taxons_controller.rb
Davidslv: I've added a controller spec for it now. Thanks

Davidslv:
Updated description.

PR #129, opened 2016-08-15
Fix tests after re-arrangement of redirects for tagging spreadsheets

carvil:
PR #123  changed the redirects of some controller actions in order to improve the user experience. 

Unfortunately, Jenkins didn't run the tests for some reason and I ended up merging the PR too soon.

Some of the tests were broken because they relied on a flow that doesn't exist anymore, so this commit fixes those tests.

### DIFF
--- a/spec/features/tag_importer_spec.rb
+++ b/spec/features/tag_importer_spec.rb
@@ -164,14 +164,17 @@ RSpec.feature "Tag importer", type: :feature do
   end
 
   def then_i_can_see_it_is_ready_for_importing
+    visit tagging_spreadsheets_path
     tagging_spreadsheet = TaggingSpreadsheet.first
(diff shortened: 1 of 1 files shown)


spec/features/tag_importer_spec.rb
```diff
@@ -164,14 +164,19 @@ def and_it_has_been_marked_as_deleted
   end
 
   def then_i_can_see_it_is_ready_for_importing
+    old_path = current_path
```
mobaig: Can we just use the path helper for the page we mean? Makes it easier to follow.

carvil: Yup, that makes sense, pushing it now

carvil:
@mobaig I've pushed the change including the explicit paths.

mobaig:
Noice.

PR #130, opened 2016-08-16
State and validation for tag mappings

carvil:
This PR adds `state` and `message` fields into TagMapping objects.

It then validates TagMappings before submitting the links to the Publishing API. The validations are:
- making sure the `content_id` is valid;
- making sure we only submit `taxon` link types for now;
- making sure the taxons already exist in our system.

Finally, it shows the TagMapping states in the import preview screen like this:

screen shot 2016-08-15 at 23 15 54

Trello: https://trello.com/c/wXQYsPTC/98-adding-state-validations-for-individual-tag-mappings

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -6,13 +6,15 @@ en:
     missing_gid: Is missing a Google Spreadsheet ID parameter (gid)
     missing_output: Is missing the parameter output as TSV
     spreadsheet_download_error: There is a problem downloading the spreadsheet, please make sure the URL works.
-
   tag_import:
     import_created: Spreadsheet uploaded. It will be imported within a few minutes.
     import_refetched: The import will be imported again.
     import_started: The tagging import has started. It can take a few minutes for the tags to be published.
     import_removed: Import has been removed.
-
+    errors:
+      invalid_content_id: We could not find this URL on GOV.UK.
+      invalid_link_types: Invalid link types found - only taxons are allowed.
+      invalid_taxons_found: Invalid taxons found - please make sure you only use existing taxons when tagging.
   messages:
     controller:
       taxons:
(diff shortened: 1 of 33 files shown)


config/locales/en.yml
```diff
@@ -6,9 +6,12 @@ en:
...
     missing_output: Is missing the parameter output as TSV
     spreadsheet_download_error: There is a problem downloading the spreadsheet, please make sure the URL works.
-
   tag_import:
     import_created: Spreadsheet uploaded. It will be imported within a few minutes.
     import_refetched: The import will be imported again.
     import_started: The tagging import has started. It can take a few minutes for the tags to be published.
     import_removed: Import has been removed.
+    errors:
+      invalid_content_id: We could not find the associated content id.
```
tijmenb: Perhaps something like "We could not find this URL on GOV.UK" (people won't know content_id).

spec/services/tag_importer/links_publisher_spec.rb
```diff
@@ -0,0 +1,153 @@
+require 'rails_helper'
+
+RSpec.describe TagImporter::LinksPublisher do
```
tijmenb: I find this test quite hard to follow. Could we make it smaller and do less, or move parts to an integration test? Feels like it's trying to do too much.

config/locales/en.yml
carvil: Sounds good, I will change that.

spec/services/tag_importer/links_publisher_spec.rb
carvil: Yup, especially the setup for each scenario...

I will try to make it better without writing an integration test first if that's ok. If that doesn't work I might split the work into more chunks.

carvil:
@tijmenb The last commit 13e75029eadb69104151226fd7bc48122882e8f1 breaks that big class (and test file) into a few pieces so it's easier to test and to see what is going on.

tijmenb:
👍

PR #136, opened 2016-08-19
Add link to content store in taxons list

mobaig:
It's convenient to see the content store representation for these
taxons. The content ID values aren't currently linked to anything so
let's have them point to the content store until a better alternative
emerges.


carvil:
:shipit:

PR #137, opened 2016-08-19
Upgrade sass-rails to prevent deprecation warnings

carvil:
After upgrading rails, we started seeing deprecation errors like this:

```
DEPRECATION WARNING: Sprockets method `register_engine` is deprecated.
Please register a mime type using `register_mime_type` then
use `register_compressor` or `register_transformer`.
```

Upgrading sass-rails fixes those.


mobaig:
👍

PR #141, opened 2016-08-25
Add error message under the taxon title

Davidslv:
The aim here is to make the error message clearer instead of hiding it
under the `Errored` tooltip.

Trello: https://trello.com/c/Vcv88VZS/109-error-message-styles-for-tag-importer

## Updated Error message

screen shot 2016-08-26 at 14 03 10


app/presenters/tag_mapping_presenter.rb
```diff
@@ -11,11 +11,19 @@ def state_title
   end
 
   def data_attributes
-    return {} unless state == 'errored'
+    return {} unless errored?
```
carvil: If we are showing the error in the tabe, we don't need the `data_attributes` anymore. Could you please delete those and also delete them form the view?

app/assets/stylesheets/application.scss
```diff
@@ -8,6 +8,12 @@
 @import 'import-preview';
 @import 'forms';
 
+// FIX: govuk_admin_template seems to be out of sync with the guidance
+// `.error-message` should be removed once govuk_admin_template is updated.
```
carvil: ok, cool.

Could you please open a PR on `govuk_admin_template` in order to add that? We can then update this once that's merged (in another PR, no need to wait for that to me merged)

carvil:
@Davidslv there is also another table that uses the same pattern [here](https://github.com/alphagov/content-tagger/blob/master/app/views/tagging_spreadsheets/index.html.erb#L23) Could you please also update that one and remove the presenter methods that are no longer needed?

Davidslv:
@carvil have addressed your comments.

app/presenters/tagging_spreadsheet_presenter.rb
```diff
@@ -18,4 +18,8 @@ def data_attributes
       'original-title': error_message
     }
   end
+
+  def errored?
+    state == 'errored'
+  end
 end
```
carvil: @Davidslv this class also has a `data_attributes` to remove: https://github.com/alphagov/content-tagger/blob/12195a1276575678c8e7de44df17612886c74926/app/presenters/tagging_spreadsheet_presenter.rb#L13-L20

app/views/tagging_spreadsheets/_import_preview.html.erb
```diff
@@ -23,11 +23,24 @@
...
+                  <% end %>
+                </ul>
+              </span>
+            <% end %>
+          </td>
           <td class="tag-mapping-status">
             <%= state_label_for(label_type: tag.label_type,
                                 title: tag.state_title,
-                                data_attributes: tag.data_attributes) %>
+                                data_attributes: []) %>
```
carvil: I think we could remove this parameter from the `state_label_for` since nothing is (or will) be using it going forward. https://github.com/alphagov/content-tagger/blob/12195a1276575678c8e7de44df17612886c74926/app/helpers/application_helper.rb#L5

app/views/tagging_spreadsheets/index.html.erb
```diff
@@ -20,9 +20,20 @@
       <tr>
         <td><%= state_label_for(label_type: spreadsheet.label_type,
                                 title: spreadsheet.state_title,
-                                data_attributes: spreadsheet.data_attributes) %></td>
-        <td><%= spreadsheet.url %></td>
-        <td><%= spreadsheet.description %></td>
+                                data_attributes: []) %></td>
```
carvil: Same as above, we can remove this parameter here and delete code from the helper.

app/views/tagging_spreadsheets/_import_preview.html.erb
Davidslv: It's gone 👍

app/views/tagging_spreadsheets/index.html.erb
Davidslv: I've removed it

spec/models/links_udpate_spec.rb
```diff
@@ -69,12 +67,24 @@
     end
 
     it 'adds an error message' do
+      stub_request(:post, "https://publishing-api.test.gov.uk/lookup-by-base-path")
```
carvil: There is a helper method to allow us to do this more succinctly:

```
publishing_api_has_lookups('/my-base-path' => 'my-content-id')
```

Without worrying with what it's actually doing behind the scenes.

spec/models/links_udpate_spec.rb
```diff
@@ -69,12 +67,24 @@
...
+        .with(body: "{\"base_paths\":[\"/a/base/path\"]}")
+        .to_return(status: 200, body: "{}")
+
+      tag_mapping = create(:tag_mapping, link_type: 'non-taxons')
+      links_update = build(:links_update, links: { 'non-taxons' => ['a-taxon-content-id'] })
+      links_update.tag_mappings = TagMapping.all
+
       expectation = lambda do
```
carvil: So this expectation was here to make sure we could call this:

```
expect { expectation.call }.to change { tag_mapping.message }
```

Since you are asserting the error message on the tag mapping, both the expectation block and assertion can go away.

This means you would only need:

```
expect(tag_mapping.reload.message).to eql("We could not find this URL on GOV.UK. Invalid link types found - only taxons are allowed.")
```

We might not need to assert for equality on the exact error message, otherwise the test will fail if we change it ever so slightly. What do you think of doing something like:

```
expect(tag_mapping.reload.message).to match(/invalid link types found/i)
```

spec/features/tag_importer_spec.rb
```diff
@@ -187,10 +187,8 @@ def then_i_see_the_import_failed
     row = first('table tbody tr')
 
     expect(row).to have_selector('.label-danger', text: state)
-    expect(row).to have_selector('.label-danger[data-toggle="tooltip"]')
-    expect(row).to have_selector(
-      ".label-danger[data-original-title='#{tagging_spreadsheet.error_message}']"
-    )
+    expect(row).to have_selector('span.error-message', text: tagging_spreadsheet.error_message)
```
carvil: We could look for only the class name here (e.g. `have_selector('.error-message'`). If we change the `span` to something else, the tests would still be green.

spec/models/links_udpate_spec.rb
Davidslv: @carvil I've changed it to the following instead:

```
      expect(tag_mapping.message).to_not match(/Content We could not find/)
      expect(tag_mapping.message).to_not match(/Link types/)
      expect(tag_mapping.message).to match(/invalid link types found/i)
```

Since we were aiming to not have the type in the beginning of each error message (in this case is `Content` and `Link types`)

spec/models/links_udpate_spec.rb
```diff
@@ -53,28 +53,43 @@
...
+      links_update = build(:links_update, links: { 'invalid' => ['my-content-id'] })
+      links_update.tag_mappings = TagMapping.all
 
-      expect { expectation.call }.to change { tag_mapping.state }.to('errored')
+      links_update.valid?
+      links_update.mark_as_errored
+      tag_mapping.reload
+
+      expect(tag_mapping.state).to eql('errored')
+      expect(tag_mapping.message).to_not match(/^Link types/)
```
mobaig: Let's remove this assertion that the message is no longer in the previous style. We've changed the behaviour. Asserting what it _should_ be is sufficient.

spec/models/links_udpate_spec.rb
```diff
@@ -53,28 +53,43 @@
...
+      links_update = build(:links_update, links: { 'taxons' => ['a-taxon-content-id'] })
+      links_update.tag_mappings = TagMapping.all
+
+      links_update.valid?
+      links_update.mark_as_errored
+      tag_mapping.reload
 
-      expect { expectation.call }.to change { tag_mapping.message }
+      expect(tag_mapping.state).to eql('errored')
+      expect(tag_mapping.message).to_not match(/^Content We could not find/)
```
mobaig: As above.

mobaig:
Looks good to me, some minor feedback on the specs. Happy to merge once updated.

spec/models/links_udpate_spec.rb
Davidslv: I would agree with you in other scenarios, but my intention was to explicitly say that we don't want the "error type" to be displayed in the string by writing this extra expectation, which seems to be a standard rails behaviour to match with the form field that has invalid data. 

I could assert the whole string but that will be extra work for the next person that changes the string, it's easier to say what you don't want to be shown.

mobaig: Right, I follow your reasoning. I think the underlying issue here is that the test for `#mark_as_errored` is difficult to follow and a little confusing in its current form. For example - both of the test cases trigger both of the validation errors, but we only match on one message at a time and there's test setup present that seems to have no implication for the actual end result.

I've had a quick chat with @carvil and have taken a slightly different approach. Given we already have tests for the validators on LinksUpdate elsewhere, the spec for `#mark_as_errored` can concentrate on the error message concatenation and state change behaviour. Pushing a commit now - WDYT?

carvil:
:shipit:

PR #142, opened 2016-08-25
Allow taxon export

carvil:
We would like to allow users to export taxons in order to make it easier to
create tagging spreadsheets.

Those spreadsheets need a mapping between taxon names and their content IDs. The
CSV file we are generating gives us exactly that, together with the link type.
This way, we can download a list of user-selected taxons into a format expected
by the tagging spreadsheet.

More information in trello: https://trello.com/c/gRoOrJoo/121-spike-build-functionality-into-content-tagger-which-makes-creating-a-tag-importer-spreadsheet-easier

It looks like this:

export-taxons

@mobaig thoughts?

### DIFF
--- a/app/controllers/export_taxons_controller.rb
+++ b/app/controllers/export_taxons_controller.rb
@@ -1,0 +1,17 @@
+class ExportTaxonsController < ApplicationController
+  def create
+    respond_to do |format|
+      format.csv { send_data exporter.data, filename: exporter.filename }
+    end
+  end
+
+private
+
+  def exporter
+    @exporter ||= Taxonomy::Exporter.new(content_ids)
+  end
+
+  def content_ids
+    params[:content_ids]
+  end
+end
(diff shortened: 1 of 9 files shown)


mobaig:
This looks solid, but I'm not sure we actually need a CSV export. If we just render an HTML table with the three required fields the user can just copy paste that straight from the browser window into Google Sheets. 

Taking that approach allows us to achieve basically the same result with less code. What do you reckon?

carvil:
@mobaig I thought about a few options - the HTML table, the Google API and then this CSV exporter.

The google API added a bit more complexity, so I discarded it for now.  I thought that adding an HTML table for the purpose of copy/pasting taxons had the potential to be less clear than exporting a CSV that Excel users are already used to work with.

However, if you think it's easier to have an HTML table I can go ahead and change that part of the PR.

By the way, is the pre-selection of taxons ok (the checkboxes)?

carvil:
Opening a new PR soon with a different approach

PR #143, opened 2016-08-26
Add taxonomy visualisation

mobaig:
View the descendants of any taxon, displayed as a tree.

- Adds an ExpandedTaxonomy class to fetch required data from the
  publishing API and use it to construct a tree.
- Uses the ruby-tree gem to provide a convenient interface for tree
  traversal and determining node depth.
- Some basic styling to visually distinguish different levels of the
  taxonomy.

screen shot 2016-08-26 at 13 54 56

### DIFF
--- a/app/assets/stylesheets/taxonomy-viz.scss
+++ b/app/assets/stylesheets/taxonomy-viz.scss
@@ -1,3 +1,7 @@
+// A set of classes governing the degree of indent depending on the depth of a
+// given node in the taxonomy. 20 has no special significance. It's large
+// enough that we're unlikely to ever need to update this rule, but small
+// enough that we don't end up creating large amounts of unnecessary CSS.
 @for $i from 1 through 20 {
   .taxon-depth-#{$i} {
     margin-left: 4rem * $i;
--- a/app/assets/stylesheets/taxonomy-viz.scss
+++ b/app/assets/stylesheets/taxonomy-viz.scss
@@ -1,0 +1,43 @@
+@for $i from 1 through 20 {
+  .taxon-depth-#{$i} {
+    margin-left: 4rem * $i;
+  }
+}
+
+$taxon-level-font-size-lg: 16px;
+$taxon-level-font-size-sm: 13px;
+
+.taxon-level {
+  padding: 0.5rem;
+  border-left: 1px solid #bfc1c3;
+  border-bottom: 1px solid #bfc1c3;
+  margin-bottom: 1rem;
+  max-width: 40%;
+
+  a.btn {
+    padding: 1px 4px;
+    margin-bottom: 0.5rem;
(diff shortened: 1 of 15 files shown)


app/models/expanded_taxonomy.rb
```diff
@@ -0,0 +1,47 @@
+class ExpandedTaxonomy
```
carvil: What do you think of creating a `Taxon` object with the response from the Publishing API in here and therefore replace calls like `taxon['content_id']` with `taxon.content_id`?

Something like this:

```
raw_root_content_item = Services.publishing_api.get_content(@content_id)
root_content_item = Taxon.new(raw_root_content_item)
```

I think it would improve readability in some places, especially in `app/views/taxonomies/show.html.erb` where we have quite a few value calls to a hash.

mobaig: Yeah, good shout. Will rework and push up some changes in a bit.

mobaig: @carvil Rewritten in this form it doesn't actually help much in terms of readability - certainly not relative to the additional logic and naming complexity it introduces. The main issue being that a lot of the hash access you see is the API of a TreeNode rather than a hash from the GdsApiAdapters.

A potential refactor might be to replace `TreeNode` with our own version that provides exactly the same interface but puts `content_id` and `parent_taxons` behind proper methods - rather than the `content` hash provided by `TreeNode`. That feels like a little more than this PR should aim to cover right now, though.

carvil: Ok, that rename makes things better (the naming got me a bit confused).

I tried to use Taxon objects instead of the hashes and ended up building a small Tree class to replace rubytree. https://github.com/alphagov/content-tagger/commit/8f6099e04ce9dc962063712545bbdbe32755670d If there is something useful there feel free to use, otherwise the renaming is fine.

mobaig:
@carvil Looks good, and nice that we can drop the ruby-tree dependency. Looks fairly straightforward to extend if we ever need other parts of the tree API (siblings, for example).

I'll work that in and push some changes.

mobaig:
Latest commit adds @carvil 's custom tree implementation with a few changes. This is ready for a final review and merge.

app/assets/stylesheets/taxonomy-viz.scss
```diff
@@ -0,0 +1,47 @@
+// A set of classes governing the degree of indent depending on the depth of a
+// given node in the taxonomy. 20 has no special significance. It's large
+// enough that we're unlikely to ever need to update this rule, but small
+// enough that we don't end up creating large amounts of unnecessary CSS.
+@for $i from 1 through 20 {
```
tijmenb: TIL

PR #144, opened 2016-08-30
Strip leading and trailing spaces from google spreadsheet url

Davidslv:
Trello: https://trello.com/c/giDY8odT/135-s-remove-white-spaces-on-spreadsheet-url

### DIFF
--- a/app/controllers/tagging_spreadsheets_controller.rb
+++ b/app/controllers/tagging_spreadsheets_controller.rb
@@ -82,6 +82,9 @@ private
   end
 
   def tagging_spreadsheet_params
-    params.require(:tagging_spreadsheet).permit(:url, :description)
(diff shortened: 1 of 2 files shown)


app/validators/google_url_validator.rb
```diff
@@ -1,6 +1,6 @@
 class GoogleUrlValidator < ActiveModel::Validator
   def validate(record)
-    uri = URI.parse(record.url)
+    uri = URI.parse(record.url.strip)
```
tijmenb: Perhaps we can pass in the stripped version into the model? Otherwise it'll still be saved in the db with the whitespace.

Davidslv: good point. will change that

PR #145, opened 2016-08-31
Allow copy of taxons

carvil:
We would like to allow users to export taxons in order to make it easier to
create tagging spreadsheets.

Those spreadsheets need a mapping between taxon names and their content IDs.
This commit adds a link from the taxons index page into a new view which
contains a table with the necessary columns: title, content ID and link type.

The user can then copy that HTML page into a spreadsheet and it would work as if
we copied CSV data.

From here:

screen shot 2016-08-31 at 12 18 33

We go here:

screen shot 2016-08-31 at 12 18 43

Trello ticket: https://trello.com/c/gRoOrJoo/121-spike-build-functionality-into-content-tagger-which-makes-creating-a-tag-importer-spreadsheet-easier

### DIFF
--- a/app/controllers/copy_taxons_controller.rb
+++ b/app/controllers/copy_taxons_controller.rb
@@ -1,0 +1,12 @@
+class CopyTaxonsController < ApplicationController
+  def index
+    taxons = taxon_fetcher.taxons.map { |taxon| Taxon.new(taxon) }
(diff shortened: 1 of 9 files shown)


carvil:
@mobaig is this more in line with your initial thoughts?

mobaig:
@carvil spot on 👍

PR #146, opened 2016-08-31
Add button to download any taxonomy visualisation as a CSV file

mobaig:
Outputs a CSV in the format expected by TreeJack.

Shout out to @carvil for the feature spec code which tests CSV download. Basically just stole this from one of his PRs.

content_tagger

screen_shot_2016-08-31_at_14_14_37


app/views/taxonomies/show.html.erb
```diff
@@ -1,6 +1,12 @@
 <%= display_header title: '', breadcrumbs: [:taxons, @taxonomy.tree.first.title] do %>
 <% end %>
 
+<%=
```
carvil: If you put this link inside the header block above, it will float right without the need of extra CSS I think. Have a look into https://github.com/alphagov/content-tagger/blob/master/app/views/taxons/index.html.erb#L1-L12 for an example.

It would be nice to have an icon too 😸

mobaig:
@carvil updated

carvil:
🚢

PR #147, opened 2016-08-31
SCSS linting

carvil:
Enable SCSS linting in `content-tagger`.

Trello: https://trello.com/c/hBkZ5MG2/140-add-sass-lint-to-content-tagger

**NOTE**: This PR needs https://github.com/alphagov/govuk-ci-scripts/pull/3 merged in first. The actual linter runs from that jenkins script. By setting `SCSS_LINT_PATH` with the path to the files we want to lint, the linter is enabled and reports any linting issues.


app/assets/stylesheets/application.scss
```diff
@@ -45,36 +44,22 @@
   }
 }
 
-table {
```
tijmenb: :(

This means a lot more classes in the HTML but I understand @robinwhittleton's concerns (https://github.com/alphagov/govuk-lint/pull/59#issuecomment-243806623).

We should really look into a bootstrap thing that allows us to include mixins, would make our lives a lot easier.

carvil: @tijmenb it would mean rewrite quite a few bootstrap classes as placeholder selectors, which would be quite a lot of work.

Without using `@extend` we generate less CSS, which is good, but I don't think it makes any difference to content tagger (small admin app for now).

I think we should try this approach and see how far we can go with it. If it becomes painful, we can always revert to what I've done in commit 95c5d40dac4c82b27b3ecccaafd90390cb92d4c5 (more specifically https://github.com/alphagov/content-tagger/blob/95c5d40dac4c82b27b3ecccaafd90390cb92d4c5/app/assets/stylesheets/application.scss#L48

carvil:
@tijmenb The latest build now runs the linter:

```
Bundle complete! 25 Gemfile dependencies, 128 gems now installed.
Gems in the group development were not installed.
Bundled gems are installed into /home/jenkins/bundles/govuk_content_tagger_branches.
Running ruby linter
0 files inspected, no offenses detected
Running SASS linter
-- enable_extension("plpgsql")
   -> 0.0160s
```

FYI, there is no output on success.

PR #148, opened 2016-08-31
Move taxon delete button

mobaig:
- Remove the button from the index page - was serving no real purpose
  being this prominent and made it a little too easy to delete a taxon.
- Add a delete button to taxon edit and show pages instead.

**Index page**

taxons

**Edit page**
14_to_19_years

**Show page**
early_years_funding

### DIFF
--- a/app/views/taxons/show.html.erb
+++ b/app/views/taxons/show.html.erb
@@ -1,5 +1,7 @@
 <%= display_header title: taxon.title, breadcrumbs: [:taxons, taxon] do %>
-  <%= link_to taxon_path(taxon.content_id), method: :delete, data: { confirm: I18n.t('messages.views.confirm') } do %>
+  <%= link_to taxon_path(taxon.content_id), method: :delete,
+    class: 'btn btn-md btn-default',
+    data: { confirm: I18n.t('messages.views.confirm') } do %>
     <i class="glyphicon glyphicon-trash"></i>
       <%= I18n.t('views.taxons.delete_taxon') %>
     </i>
--- a/app/views/taxons/show.html.erb
+++ b/app/views/taxons/show.html.erb
@@ -1,4 +1,13 @@
-<%= display_header title: taxon.title, breadcrumbs: [:taxons, taxon] %>
+<%= display_header title: taxon.title, breadcrumbs: [:taxons, taxon] do %>
+  <%= link_to taxon_path(taxon.content_id), method: :delete, data: { confirm: I18n.t('messages.views.confirm') } do %>
+    <i class="glyphicon glyphicon-trash"></i>
+      <%= I18n.t('views.taxons.delete_taxon') %>
+    </i>
+  <% end %>
+<% end %>
+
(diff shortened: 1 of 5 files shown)


app/views/taxons/edit.html.erb
```diff
@@ -10,4 +10,9 @@
       input_html: { class: :select2, multiple: true, include_blank: true } %>
 
   <%= f.submit I18n.t('views.taxons.edit_button'), class: "btn btn-lg btn-success" %>
+
+  <%= link_to 'Delete taxon', taxon_path(taxon.content_id),
+    method: :delete,
+    data: { confirm: I18n.t('messages.views.confirm') },
+    class: 'btn btn-lg btn-danger' %>
```
carvil: In this specific view, I think Mark mentioned it should be a text link rather than a button. But it's fine in the show page.

carvil:
👍 just a minor comment above

app/views/taxons/edit.html.erb
```diff
@@ -10,4 +10,9 @@
       input_html: { class: :select2, multiple: true, include_blank: true } %>
 
   <%= f.submit I18n.t('views.taxons.edit_button'), class: "btn btn-lg btn-success" %>
+
+  <%= link_to 'Delete taxon', taxon_path(taxon.content_id),
```
tijmenb: Can we move this to the header perhaps? Having it next to the submit button makes me nervous.

app/views/taxons/show.html.erb
```diff
@@ -1,5 +1,14 @@
 <%= display_header title: taxon.title, breadcrumbs: [:taxons, taxon] %>
 
+
+<%= link_to 'Delete taxon', taxon_path(taxon.content_id),
```
tijmenb: Having this inside the `display_header` block would be more in line with other apps I think.

app/views/taxons/edit.html.erb
tijmenb: I think button is appropriate here. I think we've tried to keep to link = leads you to a page, button = does a thing right?

mobaig: I've removed it altogether. Adding it to the header causes weird spacing issues, but more crucially I think Mark's point in adding a delete link to the show page was to allow the user to see the content they would be 'unlinking' from a taxon before hitting the button.

For an action the user is unlikely to do very often, I'm not sure it makes sense to include two different ways to delete a taxon. I think we should leave this for now and only add it back in if this raises confusion or seems missing to users.

app/views/taxons/edit.html.erb
mobaig: See comment above, re: removing it from the edit page.

app/views/taxons/show.html.erb
mobaig: 👍  done. This turns the button grey because of how we're styling button groups in the header, but I think it's probably better to make this functionality less prominent.

PR #149, opened 2016-08-31
Display erroneous tag mappings at the top

carvil:
When multiple tag mappings fail for some reason, they are scattered all over the
table, which might contain hundreds of entries. This makes it harder to find out
what spreadsheet rows we need to fix.

This commit groups tag mappings by state, making sure all erroneous tag mappings
appear together and at the top.

It now looks like this:

screen shot 2016-08-31 at 16 36 13


rubenarakelyan:
👍

PR #151, opened 2016-09-06
Add bulk tagging feature

carvil:
This PR allows us to:

- search for a collection;
- select which content items we want to tag;
- select which taxons we want to tag those content items with;
- preview a tag migration;
- bulk-tag all the content items.

**Note**: Until the publishing API removes the permissions from the `get_content_items` endpoint, we need to add `view_all` permissions to the `dummyapiuser` in development mode in order to see results from the search.

Trello: https://trello.com/c/9LajM53J/10-epic-bulk-tagging-from-existing-taxonomies-to-the-new-one

Work done by both myself and @mobaig (the rebase erased some of the commit details).

What's next:
- [ ] support for topics and browse pages
- [ ] tidy up navigation (perhaps merge some functionality into a single navigation item).

These will come in separate PRs.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,7 @@ gem 'plek', '~> 1.10'
 gem 'gds-sso', '~> 11.1'
 gem 'govuk_admin_template', '~> 4.1.1'
 gem 'simple_form', '~> 3.2', '>= 3.2.1'
-gem 'gds-api-adapters', '~> 32.3'
+gem 'gds-api-adapters', '~> 35.0'
 
 gem 'airbrake', '~> 4.3.1'
 
--- a/Gemfile
+++ b/Gemfile
@@ -32,6 +32,7 @@ end
 
 group :development do
   gem 'web-console', '~> 2.0'
+  gem "better_errors"
 end
 
 group :test do
--- a/app/controllers/bulk_tagging/collections_controller.rb
+++ b/app/controllers/bulk_tagging/collections_controller.rb
@@ -1,0 +1,15 @@
+module BulkTagging
+  class CollectionsController < ApplicationController
+    def show
+      expanded_links = ExpandedLinksFetcher.expanded_links(params[:content_id])
+      taxons = Taxonomy::TaxonFetcher.new.taxons
+
+      render :show, locals: {
+        bulk_tagging: TagMigration.new,
+        content_id: params[:content_id],
+        taxons: taxons,
+        expanded_links: expanded_links
+      }
+    end
+  end
+end
--- a/app/controllers/bulk_tagging/update_tags_controller.rb
+++ b/app/controllers/bulk_tagging/update_tags_controller.rb
@@ -1,0 +1,30 @@
+module BulkTagging
+  class UpdateTagsController < ApplicationController
+    def create
+      tag_migration.save!
+      redirect_to tag_migration_path(tag_migration)
+
+    rescue BulkTagging::BuildTagMigration::InvalidArgumentError => e
+      flash[:error] = e.message
+      redirect_to bulk_tagging_collection_path(params[:collection_content_id])
+    end
+
+  private
+
+    def tag_migration
+      @tag_migration ||= BulkTagging::BuildTagMigration.perform(
+        original_link_content_id: params[:collection_content_id],
+        taxon_content_ids: taxon_content_ids,
+        content_base_paths: content_base_paths
+      )
+    end
+
+    def taxon_content_ids
+      params[:taxons]
+    end
+
+    def content_base_paths
+      params[:content_base_paths]
+    end
+  end
+end
--- a/app/controllers/bulk_taggings_controller.rb
+++ b/app/controllers/bulk_taggings_controller.rb
@@ -1,0 +1,24 @@
+class BulkTaggingsController < ApplicationController
+  def new
+    render :new, locals: { results: [] }
+  end
+
+  def search
+    query = params[:collection_search][:query]
+    gds_response = Services.publishing_api.get_content_items(
+      document_type: "document_collection",
+      per_page: 20,
+      q: query
+    )
+
+    results = gds_response['results'].map { |result| ContentItem.new(result) }
+
+    render :new, locals: { results: results }
+  end
+
+  private
+
+  def search_params
+    params.require(:collection_search).permit(:query)
+  end
+end
--- a/app/controllers/bulk_taggings_controller.rb
+++ b/app/controllers/bulk_taggings_controller.rb
@@ -1,22 +1,31 @@
 class BulkTaggingsController < ApplicationController
   def new
-    render :new, locals: { results: [] }
+    render :new, locals: { results: [], query: '' }
   end
 
   def search
-    query = params[:collection_search][:query]
-    gds_response = Services.publishing_api.get_content_items(
-      document_type: "document_collection",
-      per_page: 20,
-      q: query
-    )
+    warning_about_multiple_pages
 
-    results = gds_response['results'].map { |result| ContentItem.new(result) }
-
-    render :new, locals: { results: results }
+    render :new, locals: { results: search_response.results, query: query }
   end
 
-  private
+private
+
+  def warning_about_multiple_pages
+    if search_response.multiple_pages?
+      flash[:warning] = I18n.t('controllers.bulk_taggings.too_many_results')
+    else
+      flash.delete(:warning)
+    end
+  end
+
+  def search_response
+    @search_response ||= BulkTagging::Search.perform(query: query)
+  end
+
+  def query
+    params[:collection_search][:query]
+  end
 
   def search_params
     params.require(:collection_search).permit(:query)
--- a/app/models/search_response.rb
+++ b/app/models/search_response.rb
@@ -11,10 +11,6 @@ class SearchResponse
     gds_api_response_hash['results'].map { |result| ContentItem.new(result) }
   end
 
-  def successful?
-    gds_api_response.code == 200
-  end
-
   def multiple_pages?
     gds_api_response_hash['pages'] > 1
   end
--- a/app/models/search_response.rb
+++ b/app/models/search_response.rb
@@ -1,0 +1,21 @@
+class SearchResponse
+  attr_reader :gds_api_response, :gds_api_response_hash, :document_type
+
+  def initialize(gds_api_response, document_type)
+    @gds_api_response = gds_api_response
+    @gds_api_response_hash = gds_api_response.to_hash
+    @document_type = document_type
+  end
+
+  def results
+    gds_api_response_hash['results'].map { |result| ContentItem.new(result) }
+  end
+
+  def successful?
+    gds_api_response.code == 200
+  end
+
+  def multiple_pages?
+    gds_api_response_hash['pages'] > 1
+  end
+end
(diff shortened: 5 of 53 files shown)


carvil:
Screenshots to follow shortly

carvil:
The flow is:

- Go to the Bulk Tagging page:

screen shot 2016-09-06 at 17 02 32

- Search for a collection:

screen shot 2016-09-06 at 17 03 41

- Select which collection to use and then pick the taxons and content items to bulk tag:

screen shot 2016-09-06 at 17 04 44

- Preview your new bulk tagging import:

screen shot 2016-09-06 at 17 06 18

- Create the tags:

screen shot 2016-09-06 at 17 06 44

app/controllers/bulk_tagging/update_tags_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    def taxon_content_ids
+      params[:taxons]
+    end
+
+    def content_base_paths
+      params[:content_base_paths]
+    end
+
+    # TODO: make this more readable, weird pattern
```
tijmenb: Could this validation be moved into `BulkTagging::BuildTagMigration`? It looks like a form object anyway.

app/controllers/bulk_tagging/collections_controller.rb
```diff
@@ -0,0 +1,15 @@
+module BulkTagging
```
tijmenb: Would it be clearer for now to merge this and `UpdateTagsController`? Feels like there would be lots of small controllers if we specify one per feature (topics/browse/search etc).

app/controllers/bulk_tagging/update_tags_controller.rb
carvil: Yes, that would make more sense. i will move that into the service class instead.

config/locales/en.yml
```diff
@@ -26,11 +26,16 @@ en:
...
       new_button: Create taxon
       edit_button: Update taxon
-  messages:
-    controller:
-      taxons:
-        success: You have sucessfully deleted the taxon
-        alert: It was not possible to delete the taxon
+  controllers:
+    bulk_taggings:
+      too_many_results: Your search returned too many results. We are only displaying a subset, please try norrowing down your search.
```
tijmenb: * narrowing

app/controllers/bulk_taggings_controller.rb
```diff
@@ -0,0 +1,37 @@
+class BulkTaggingsController < ApplicationController
+  def new
+    render :new, locals: { results: [] }
+  end
+
+  def search
+    if search_response.successful?
```
tijmenb: What kind of errors do we expect here? Perhaps we should let the error bubble up.

app/controllers/bulk_tagging/collections_controller.rb
carvil: So originally it was like that, but it makes the routes more complicated and less readable (custom get/post entries with custom paths etc).

By using the conventional way in rails of having a resource and a separate controller we make the routes more readable (at the expense of having more controller classes). But that could be good, otherwise the original controller will be quite big after we are done with the bulk tagging.

What do you think?

app/models/search_response.rb
```diff
@@ -0,0 +1,21 @@
...
+    @gds_api_response_hash = gds_api_response.to_hash
+    @document_type = document_type
+  end
+
+  def results
+    gds_api_response_hash['results'].map { |result| ContentItem.new(result) }
+  end
+
+  def successful?
+    gds_api_response.code == 200
```
tijmenb: I don't think this is needed, api-adapters will raise an error for all 4XX and 5XX errors anyway.

config/locales/en.yml
carvil: Whoops!

app/controllers/bulk_taggings_controller.rb
carvil: If the API does not respond properly (i.e. with code 200). I'm imagining a timeout (we saw some locally). If we don't do this we get a 500 error...

app/models/search_response.rb
carvil: Does that mean we would get an exception?

app/views/bulk_taggings/new.html.erb
```diff
@@ -0,0 +1,38 @@
+<%= display_header title: 'Bulk tagging', breadcrumbs: ["Search"] %>
+
+<div class="row">
+  <div class="col-md-4">
+    <%= simple_form_for :collection_search, url: search_bulk_tagging_path , method: :get do |f| %>
+      <div class="form-group">
+        <%= f.input :query, input_html: { type: :text, class: 'form-control' }, label: 'Query' %>
```
tijmenb: Should have the current query pre-filled.

app/controllers/bulk_taggings_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      flash[:error] = I18n.t('controllers.bulk_taggins.failure')
+      render :new, locals: { results: [] }
+    end
+  end
+
+private
+
+  def warning_about_multiple_pages
+    if search_response.multiple_pages?
+      flash[:warning] = I18n.t('controllers.bulk_taggings.too_many_results')
```
tijmenb: Don't forget some comment in readme to explain why too much is too much.

app/views/bulk_tagging/collections/show.html.erb
```diff
@@ -0,0 +1,42 @@
...
+            <th>Select</th>
+            <th>Title</th>
+            <th>Content ID</th>
+          </tr>
+        </thead>
+        <tbody>
+          <% expanded_links.each do |expanded_link| %>
+            <tr>
+              <td><%= check_box_tag 'content_base_paths[]', expanded_link.base_path, true %></td>
+              <td><%= expanded_link.title %></td>
```
tijmenb: Perhaps add link to web url as well.

tijmenb:
From IRL conversation:

- Perhaps replace `original_link_content_id` with a text explanation of the original search

For follow up:

- Let's talk about the naming of all these things
- We need to look at the UI of the collections-tag page (make it different from search)
- Merge tags in index bulk views
- "Select all" checkbox for collections

app/controllers/bulk_tagging/collections_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+  class CollectionsController < ApplicationController
```
tijmenb: I thought this referred to the `document_collection` format - perhaps rename to something else (don't know what).

app/models/search_response.rb
carvil: Talked IRL, we can delete this branch, no need to check for errors as exceptions are raised.

Gemfile
```diff
@@ -9,7 +9,7 @@ gem 'plek', '~> 1.10'
 gem 'gds-sso', '~> 11.1'
 gem 'govuk_admin_template', '~> 4.1.1'
 gem 'simple_form', '~> 3.2', '>= 3.2.1'
-gem 'gds-api-adapters', '~> 32.3'
+gem 'gds-api-adapters'
```
tijmenb: Pinning should stay here so we can do `bundle update` and not break.

app/services/bulk_tagging/build_tag_mapping.rb
```diff
@@ -0,0 +1,24 @@
...
+      @taxon = taxon
+      @content_base_path = content_base_path
+    end
+
+    def self.perform(taxon:, content_base_path:)
+      new(taxon: taxon, content_base_path: content_base_path).perform
+    end
+
+    def perform
+      TagMapping.new(
```
tijmenb: I'll investigate why this doesn't raise rubocop complaints.

config/routes.rb
```diff
@@ -19,6 +19,21 @@
     get  'import_progress'
   end
 
+  resources :tag_migrations, except: %i(update edit new) do
```
tijmenb: Can we flip this to `only`? Easier to read the whole routes.

app/views/tag_migrations/show.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= display_header title: 'Preview import', breadcrumbs: [:tag_migrations, "Preview"] do %>
```
mobaig: Just a note/reminder, possibly this'll get picked up when we revisit naming of things - we should make this header more generic (it's a migration not an import) or it should be passed in from the controller.

config/routes.rb
carvil: 👍

app/services/bulk_tagging/build_tag_mapping.rb
carvil: I will look into this in a separate PR if that's ok.

Gemfile
carvil: Ok, will do.

app/controllers/bulk_tagging/collections_controller.rb
carvil: I will move this to a new ticket. I think it makes sense to reconsider the name when we add topics/browse pages too. We would be in a better place to rename this by then.

app/views/bulk_tagging/collections/show.html.erb
carvil: Yup, makes sense. I will add that link.

app/controllers/bulk_taggings_controller.rb
carvil: Ok, I will add an entry regarding the bulk tagger to the README.

app/views/bulk_taggings/new.html.erb
carvil: 👍

tijmenb:
I finally thought of what makes "migration" seem like a wrong word for this. Migration is the moving of things from one thing to another thing. Since we are keeping the original (document collection in this) case, it's not a "migration".

app/views/tag_migrations/show.html.erb
carvil: Yes, I will add that and Tijmen's suggestion to that trello card. I think we should revisit naming throughout.

carvil:
@tijmenb I've addressed most of the issues you mentioned in the PR. 

I've also added a bunch of follow up tasks to Trello:
- https://trello.com/c/xUoBfOoT/150-bulk-tagging-search-for-topics-and-browse-pages-besides-collections
- https://trello.com/c/Erx6NIwe/151-content-tagger-tidy-up-navigation-in-order-to-consolidate-all-tagging-actions-into-one-navigation-entry
- https://trello.com/c/H403aTpk/152-bulk-tagging-add-description-field-to-tag-migration
- https://trello.com/c/bw6RqXta/153-bulk-tagging-make-sure-the-bulk-taggings-new-and-collections-show-views-are-different-from-each-other-to-avoid-confusion
- https://trello.com/c/lnE4VdSj/154-bulk-tagging-in-the-preview-of-tag-migrations-we-should-consolidate-tag-mapping-records-in-order-to-see-one-row-per-content-item
- https://trello.com/c/IJoqzkbq/155-bulk-tagging-add-select-all-checkbox-that-selects-deselects-all-checkboxes-in-collections-show
- https://trello.com/c/35hTnnQk/156-content-tagger-naming-all-the-things


@mobaig @tijmenb I think we are now in a state where we can merge this and carry on with the tasks above. This PR is already too big and merging and breaking tasks apart would allow all the devs to work on it. WDYT?

tijmenb:
Yeah, let's merge this now and fix stuff in follow ups.

PR #152, opened 2016-09-07
Display context around a tag migration

carvil:
This commits adds the following to a TagMigration:
- The original query string
- The original base path from the collection used

These 2 new fields will provide context when looking at the overview page of
TagMigrations.

Trello: https://trello.com/c/H403aTpk/152-bulk-tagging-add-a-description-to-tag-migration-in-order-to-provide-context

Before this change, it looked like this:

screen shot 2016-09-07 at 14 49 19

Now, it looks like this:

screen shot 2016-09-07 at 14 46 21

### DIFF
--- a/app/controllers/bulk_tagging/collections_controller.rb
+++ b/app/controllers/bulk_tagging/collections_controller.rb
@@ -7,8 +7,10 @@ module BulkTagging
       render :show, locals: {
         bulk_tagging: TagMigration.new,
         content_id: params[:content_id],
+        base_path: params[:base_path],
         taxons: taxons,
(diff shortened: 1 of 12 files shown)


tijmenb:
Do we know how this will hold up when we expand the search capabilities to free-text-search/topics/collections?

carvil:
@tijmenb At the moment, the text field is free-text-search.

I'm imagining that the selection of the `document_type` (collection, topics, etc) will happen in something like checkboxes (enable/disable certain document types), which means this would generalise well. We also save the the base_path of the collection/topics/etc that we choose, which is also general enough.

PR #155, opened 2016-09-09
Order taxons by updated at

carvil:
This PR does the following:

- changes the way we fetch Taxons from `get_linkables` to `get_content_items`;
- sorts taxons by their `public_updated_at` attribute rather than by `title`. This is now possible because we are using `get_content_items`, which returns a lot more attributes for each of the taxons.

Trello: https://trello.com/c/y9VdJshp/111-add-created-at-updated-at-ordering-to-taxons-list

### DIFF
--- a/app/services/taxonomy/taxon_fetcher.rb
+++ b/app/services/taxonomy/taxon_fetcher.rb
@@ -13,10 +13,6 @@ module Taxonomy
       taxons.map(&:content_id)
     end
 
-    def taxons_for_select
-      taxons.map { |taxon| [taxon.title, taxon.content_id] }
-    end
-
     def parents_for_taxon(taxon_child)
       taxons.select do |taxon|
         taxon_child.parent_taxons.include?(taxon.content_id)
--- a/app/services/taxonomy/taxon_fetcher.rb
+++ b/app/services/taxonomy/taxon_fetcher.rb
@@ -1,11 +1,7 @@
 module Taxonomy
   class TaxonFetcher
-    # Return a list of taxons from the publishing API with links included.
     def taxons
-      @taxons ||=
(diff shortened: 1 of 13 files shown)


app/services/taxonomy/taxon_fetcher.rb
```diff
@@ -1,11 +1,9 @@
 module Taxonomy
   class TaxonFetcher
-    # Return a list of taxons from the publishing API with links included.
     def taxons
-      @taxons ||=
-        Services.publishing_api.get_linkables(document_type: 'taxon')
-          .map { |taxon_hash| Taxon.new(taxon_hash) }
-          .sort_by(&:title)
+      @taxons ||= taxon_list.sort do |taxon_1, taxon_2|
```
tijmenb: I think it might be better to use the publishing-api's ordering method instead of doing it here. That would enable us to use pagination later.

tijmenb:
Cool, good to see that we no longer misuse linkables for the index page with this.

app/services/taxonomy/taxon_fetcher.rb
carvil: @tijmenb cool, I will switch to that, it makes sense to me.

Unfortunately, that means we can't "test" ordering, so we have to trust that our mocks work (which is fine).

carvil:
I will fix this so we use linkables where needed and get_content_items elsewhere.

carvil:
@tijmenb I've changed things a bit:

- we use the publishing api to sort by `public_updated_at`;
- we use `get_linkables` for dropdowns and `get_content_items` for the taxons' index page.

WDYT?

tijmenb:
Most excellent!

PR #156, opened 2016-09-09
Refactor bulk tagging code

mobaig:
See commits for details.


config/routes.rb
```diff
@@ -19,19 +19,13 @@
     get  'import_progress'
   end
 
-  resources :tag_migrations, only: %i(index show create) do
+  resources :tag_migrations, only: [:index, :new, :create, :show, :destroy] do
```
carvil: What do you think of writing `except: %i(edit update)` instead (as we are allowing most actions)?

carvil:
Very nice! 👍

config/routes.rb
mobaig: Tempted because it's shorter, but I always find it easier to see whats available when the `only` form is used.

app/views/tag_migrations/index.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= display_header title: 'Tag Migrations', breadcrumbs: ['Bulk Tagging'] do %>
```
carvil: Sorry, only spotted this now. Should we rename the breakcrumbs to `Tag Searches` instead of `Bulk Tagging`?

config/routes.rb
carvil: Yeh, that's a fair point.

app/views/tag_migrations/index.html.erb
mobaig: Good spot - but `Tag Migration` might make more sense here. See latest commit.

carvil:
🚢

PR #157, opened 2016-09-12
Use simple_form configuration files from govuk_admin_template

carvil:
We have moved a default configuration and locale files into
`govuk_admin_template`. This commit makes sure we use them.

Note: when https://github.com/alphagov/govuk_admin_template/pull/131 gets merged, we can remove the `label` in the taxon form below.


mobaig:
👍

PR #159, opened 2016-09-12
Append to existing links

carvil:
This PR:

- makes sure we append to the links of content items rather than overriding them;
- changes the namespace of 2 service classes, making it more obviously where they are being used.

More details in the commit messages below.


app/services/bulk_tagging/links_publisher.rb
```diff
@@ -14,12 +14,24 @@ def publish
       if links_update.valid?
         Services.publishing_api.patch_links(
```
tijmenb: Perhaps we should use the publishing-api versioning system here.

The `publishing_api#get_links` call returns a `version`, which you can pass into `patch_links`. 

If somebody else modified the links in the meantime, publishing-api will return a HTTP Conflict status code (raises an exception). This prevents accidental overriding of data.

app/controllers/tag_migrations_controller.rb
```diff
@@ -50,7 +50,7 @@ def destroy
   end
 
   def publish_tags
-    TagImporter::PublishTags.new(tag_migration, user: current_user).run
+    BulkTagging::PublishTags.new(tag_migration, user: current_user).run
```
mobaig: Not a huge thing, but what do you reckon about _not_ namespacing this at all until we sort out the naming fiasco?

app/services/bulk_tagging/links_publisher.rb
carvil: I didn't know about this version param in `get_links`! I will make the change.

app/controllers/tag_migrations_controller.rb
carvil: Sure, I will get rid of that namespacing for now.

Davidslv:
Looks good to me, let me know when it gets merged so I can pull master to resolve a possible conflict.

Thanks 👍

PR #160, opened 2016-09-13
Update tag migrator to support copying tags from topics and browse pages

mobaig:
Also: Rename and namespace ExpandedLinksFetcher


app/models/bulk_tagging_source.rb
```diff
@@ -0,0 +1,19 @@
+class BulkTaggingSource
+  def source_name_to_content_key_map
+    {
+      document_collection: 'documents',
+      topic: 'children',
+      mainstream_browse_page: 'children',
```
carvil: These values got me a bit confused.  I thought we would access links by `topic` and `mainstream_browse_page` rather than `children`.

carvil:
@mobaig do you think it would be worthwhile to expand the bulk tagging feature with the new new document types in order to make sure we can see them and select them?

I was thinking maybe expanding this: https://github.com/alphagov/content-tagger/blob/master/spec/features/bulk_tagging_spec.rb#L36-L46

app/models/bulk_tagging_source.rb
mobaig: It's because of the shape of the response from get_expanded_links.

Take a look at this in a console on your local or on integration: `Services.publishing_api.get_expanded_links("3e275a11-0fae-425b-a7a1-fe434594693f").to_h`

carvil: 👍

carvil:
🚢

PR #161, opened 2016-09-13
Error handling improvements when importing spreadsheet data

mobaig:
Fix a bug where blank values for certain fields in the import TSV file would result in the worker choking on a DB not null constraint.

Includes some small behavioural changes in how we handle the import. See commits for details.

https://trello.com/c/0JxMuAxU/166-content-tagger-fix-error-handling-behaviour-when-tag-mapping-base-path-is-blank


Davidslv:
Looks good to me 👍

PR #162, opened 2016-09-13
Fix progress bars on TSV import and tag migration screens

mobaig:
These broke somewhere along the line of introducing the tag migrator
functionality. This commit fixes them.

- Ensure both controllers implement the 'progress' ajax action
  correctly, passing in the correct locals to the progress bar partial.
- Use a more generic name applicable to both workflows - the progress
  bar and preview screens relate to a 'tag update' on several content
  items.


rubenarakelyan:
👍

carvil:
@mobaig what do you think of adding some assertions on the markup? I'm just afraid we will change something else in the future and it breaks again..

mobaig:
@carvil what kind of assertions did you have in mind?

Another approach is a really lightweight poltergeist test to ensure that the bar progresses without refresh (i.e - the js actually works).

carvil:
@mobaig I think we should probably go for the latter. We could probably get away with `capybara-webkit` and add `js: true` to this specific test. Maybe that's worth a try and it would be less flaky.

Otherwise, we could try to make sure the module is loaded in the pages we need it to load (looking for the data attribute) and make sure both controllers respond to that action. This obviously isn't great as it doesn't test end-to-end...

mobaig:
@carvil cool, will write a followup ticket.

PR #163, opened 2016-09-13
Workaround for linkables with internal names of nil

kevindew:
Accessing linkables on the Publishing API can return nil for fields such as internal_name or title when the linkable item is a redirect. In this instance we have a problem when the internal_name is nil as we require it to be a string.

To address this we filter the linkables from the Publishing API to only include instances where the internal_name is not nil.

Whether redirects should be in the linkables is something we need to work out over at Publishing API land, but hopefully this can resolve the issues you're currently having.

### DIFF
--- a/lib/linkables.rb
+++ b/lib/linkables.rb
@@ -52,6 +52,9 @@ private
   end
 
   def get_tags_of_type(document_type)
-    Services.publishing_api.get_linkables(format: document_type)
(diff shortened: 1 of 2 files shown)


tijmenb:
Thanks @kevindew.

We'll revert this PR when the bug is fixed in the publishing-api.

PR #165, opened 2016-09-14
Add feature spec to test the progress bar

carvil:
This test makes sure we can see the progress bar and that it completes.

The side effects of adding this feature test (and other JS tests in the future)
is that:
- we cannot use `:rack_test`, therefore we need to rely on a slower `webkit`
  driver to run those particular javascript tests;
- we cannot enable RSpec fixtures, and have to use `database_cleaner` instead (see [this](https://github.com/jnicklas/capybara/blob/master/README.md#transactions-and-database-setup));
- we have to use a `truncation` strategy, which means we then need to create our
  User record before each example (or re-create it after the js test)

On the plus side, we can select the Capybara javascript driver depending on the
type of tests, which means most of the tests still use `:rack_test`.

Trello: https://trello.com/c/qoimvKOL/167-add-a-basic-poltergeist-test-for-the-progress-bar-in-content-tagger

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -38,4 +38,7 @@ end
 group :test do
   gem 'webmock', '~> 1.22'
   gem 'capybara'
+  gem 'capybara-webkit'
+  gem 'headless'
+  gem 'database_cleaner'
 end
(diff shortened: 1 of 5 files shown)


mobaig:
Nice one 👍 .

PR #166, opened 2016-09-14
Make service object naming clearer and more consistent

mobaig:
Right now we have a mix of verb-based and noun-based names for these
classes. This commit updates them so that they're all verb-based. It
also makes some class names more explicit about the behaviour or
business logic they implement.


app/services/queue_links_for_publishing.rb
```diff
@@ -1,4 +1,4 @@
-class PublishTags
+class QueueLinksForPublishing
```
carvil: How about `LinksForPublishing.enqueue`?

app/services/taxonomy/build_taxon_payload.rb
```diff
@@ -1,5 +1,5 @@
 module Taxonomy
-  class TaxonPayloadBuilder
+  class BuildTaxonPayload
```
carvil: Since we have a `build` method, how about just `TaxonPayload`?

app/services/taxonomy/fetch_all_taxons.rb
```diff
@@ -1,5 +1,5 @@
 module Taxonomy
-  class TaxonFetcher
+  class FetchAllTaxons
```
carvil: I'm not too sure about this one, as this class does more than just fetch all taxons. Not super convinced of the following suggestion, but here it goes:

```
module Taxonomy
  class RemoteTaxons
    def all
       # existing taxons method
    end

    def content_ids
      # existing taxon_content_ids
    end

    def parents_for(taxon)
      # existing parents_for_taxon
    end
  end
end
```

WDYT?

mobaig: What if we went with that name and that API, but put this in `app/models`? All this class is doing is wrapping responses from our persistence layer, i.e. - the pub api.

carvil: That makes sense to me. We could even move them into the existing `Taxon` model, and have make `all` a class method. If you don't think that's a good idea, then let's just move this class into `app/models` for now. 👍

app/services/queue_links_for_publishing.rb
mobaig: As discussed IRL, aiming for consistency in the verb vs noun naming of these service objects.

app/services/taxonomy/build_taxon_payload.rb
mobaig: as above, re: verb vs noun

app/controllers/taxons_controller.rb
```diff
@@ -73,7 +73,7 @@ def parent_taxons
   end
 
   def taxon_fetcher
```
carvil: We should probably rename this method name too. WDYT?

mobaig: 👍

PR #167, opened 2016-09-14
Consolidate TagMappings by content_base_path

Davidslv:
This approach is to consolidate records in order to see them in one row per content item change.

I would like to receive some feedback on this approach and also make sure that it's not missing anything. I'm not sure if the local data I'm seeing is the whole picture as I seem to only have one `link_type` per content_base_path. 

Will leave the failing specs to be fixed after some discussion.

Trello: https://trello.com/c/lnE4VdSj

## Before

screen shot 2016-09-14 at 16 12 59

## After

screen shot 2016-09-15 at 16 07 30
screen shot 2016-09-15 at 16 06 47

screen shot 2016-09-15 at 16 56 15

### DIFF
--- a/app/models/aggregated_tag_mapping.rb
+++ b/app/models/aggregated_tag_mapping.rb
@@ -1,0 +1,21 @@
+class AggregatedTagMapping
+  include ActiveModel::Model
+
+  attr_accessor :content_base_path, :tag_mappings
+
+  def links
+    tag_mappings.map do |tag_mapping|
+      Link.new(
+        link_title: tag_mapping.link_title,
+        link_content_id: tag_mapping.link_content_id,
+        link_type: tag_mapping.link_type,
+      )
+    end
+  end
+
+  class Link 
+    include ActiveModel::Model
+
+    attr_accessor :link_title, :link_content_id, :link_type
+  end
+end
--- a/app/models/aggregated_tag_mapping.rb
+++ b/app/models/aggregated_tag_mapping.rb
@@ -13,7 +13,7 @@ class AggregatedTagMapping
     end
   end
 
-  class Link 
+  class Link
     include ActiveModel::Model
 
     attr_accessor :link_title, :link_content_id, :link_type
--- a/app/models/tag_migration.rb
+++ b/app/models/tag_migration.rb
@@ -18,4 +18,25 @@ class TagMigration < ActiveRecord::Base
   def error_count
     tag_mappings.errored.count
   end
+
+  def aggregated_tag_mappings
+    tag_mappings_grouped_by_content_base_path.reduce([]) do |accumulator, aggregation|
+      accumulator << AggregatedTagMapping.new(content_base_path: aggregation.first, tag_mappings: aggregation.last)
+    end
+  end
+
+private
+
+  def tag_mappings_grouped_by_content_base_path
+    tag_mappings.by_state.by_content_base_path.by_link_title
+      .select(
+        :id,
+        :link_type,
+        :link_title,
+        :content_base_path,
+        :messages,
+        :link_content_id,
+        :state)
+      .group_by(&:content_base_path)
+  end
 end
--- a/app/models/tag_migration.rb
+++ b/app/models/tag_migration.rb
@@ -1,4 +1,6 @@
 class TagMigration < ActiveRecord::Base
+  include AggregatableTagMappings
+
   has_many :tag_mappings, dependent: :destroy, as: :tagging_source
   validates :source_content_id, presence: true
 
@@ -17,26 +19,5 @@ class TagMigration < ActiveRecord::Base
 
   def error_count
     tag_mappings.errored.count
-  end
-
-  def aggregated_tag_mappings
-    tag_mappings_grouped_by_content_base_path.reduce([]) do |accumulator, aggregation|
-      accumulator << AggregatedTagMapping.new(content_base_path: aggregation.first, tag_mappings: aggregation.last)
-    end
-  end
-
-private
-
-  def tag_mappings_grouped_by_content_base_path
-    tag_mappings.by_state.by_content_base_path.by_link_title
-      .select(
-        :id,
-        :link_type,
-        :link_title,
-        :content_base_path,
-        :messages,
-        :link_content_id,
-        :state)
-      .group_by(&:content_base_path)
   end
 end
--- a/app/presenters/aggregated_tag_mapping_presenter.rb
+++ b/app/presenters/aggregated_tag_mapping_presenter.rb
@@ -1,17 +1,11 @@
 class AggregatedTagMappingPresenter < SimpleDelegator
   def errored?
-    presented_tag_mappings.any? do |tag_mapping|
-      tag_mapping.errored?
-    end
+    presented_tag_mappings.any?(&:errored?)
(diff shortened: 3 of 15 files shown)


carvil:
Hey @Davidslv , a couple of thoughts:

- we also need to make the same change in the `tag_migrations_controller`
- we need to see the list of all the taxons we are tagging the content items with in the consolidated view.

So in my example, before we had this:

screen shot 2016-09-14 at 16 56 30

As you can see, there are 3 unique content items, each of them being tagged with 2 different taxons. 

I would expect the consolidated view to include the 2 taxons per row (Both Fruit and Apples in this dummy example). However, it only shows one:

screen shot 2016-09-14 at 16 57 34

Another important aspect of this feature is how we represent the state of the consolidated rows. Since one row might represent multiple TagMapping objects, we need to display a state label that is consistent with the state of all those TagMapping objects.

Does this make sense?

app/models/aggregated_tag_mapping.rb
```diff
@@ -0,0 +1,28 @@
+class AggregatedTagMapping
+  class Link
+    attr_accessor :link_title, :link_content_id, :link_type
+
+    def initialize(link_title:, link_content_id:, link_type:)
```
carvil: If you use `include ActiveModel::Model` you don't need to create an accessor, the `attr_accessor` is enough. More details [here](http://api.rubyonrails.org/classes/ActiveModel/Model.html)

app/presenters/aggregated_tag_mapping_presenter.rb
```diff
@@ -0,0 +1,20 @@
...
+  def errored?
+    presented_tag_mappings.any? { |tag_mapping| tag_mapping.state == 'errored' }
+  end
+
+  def error_messages
+    presented_tag_mappings.flat_map do |tag_mapping|
+      next unless tag_mapping.message
+
+      # splits message on ".", but keeps it in the message.
+      tag_mapping.message.split(/(?<=[.])/)
```
carvil: We no longer split messages because they are now an array in the database. Please rebase so you can see that change (more details [here](https://github.com/alphagov/content-tagger/blob/master/app/models/tag_mapping.rb#L12))

app/presenters/aggregated_tag_mapping_presenter.rb
```diff
@@ -0,0 +1,20 @@
...
+    end.reject(&:blank?)
+  end
+
+  def presented_tag_mappings
+    tag_mappings.map do |tag_mapping|
+      TagMappingPresenter.new(tag_mapping)
```
carvil: If we are moving towards the aggregated tag mapping, then we can delete this presenter and port the methods across.

app/models/tag_migration.rb
```diff
@@ -14,4 +14,25 @@ class TagMigration < ActiveRecord::Base
...
+    tag_mappings_grouped_by_content_base_path.reduce([]) do |accumulator, aggregation|
+      accumulator << AggregatedTagMapping.new(content_base_path: aggregation.first, tag_mappings: aggregation.last)
+    end
+  end
+
+private
+
+  def tag_mappings_grouped_by_content_base_path
+    tag_mappings.by_state.by_content_base_path.by_link_title
+      .select(
```
carvil: Do we need all these fields here?

app/models/tag_migration.rb
```diff
@@ -14,4 +14,25 @@ class TagMigration < ActiveRecord::Base
...
+
+  def aggregated_tag_mappings
+    tag_mappings_grouped_by_content_base_path.reduce([]) do |accumulator, aggregation|
+      accumulator << AggregatedTagMapping.new(content_base_path: aggregation.first, tag_mappings: aggregation.last)
+    end
+  end
+
+private
+
+  def tag_mappings_grouped_by_content_base_path
```
carvil: This method seems to be repeated in 2 models, what do you think we could do here to avoid that?

tijmenb:
@Davidslv can you rebase this?

app/models/aggregated_tag_mapping.rb
Davidslv: 👍

app/presenters/aggregated_tag_mapping_presenter.rb
Davidslv: cool 👍

app/presenters/aggregated_tag_mapping_presenter.rb
Davidslv: @carvil I'm not sure how I can accomplish that

app/models/tag_migration.rb
Davidslv: I've used a shared module for this

app/models/tag_migration.rb
Davidslv: I've removed the `:id`, the rest is being used on the view

app/presenters/aggregated_tag_mapping_presenter.rb
carvil: Ok so the aggregated tag mapping object will have a number of TagMappings, so you still need this in order to show the label. That's fine then, nothing to do here 👍

app/models/tag_migration.rb
carvil: Nice, thanks!

app/models/tag_migration.rb
carvil: 👍

app/views/application/_tag_update_preview.html.erb
```diff
@@ -1,10 +1,10 @@
 <div class="tag-update-preview">
   <br/>
 
-  <div data-module="tag-update-progress">
-    <% if tag_mappings.any? %>
+  <div data-module="import-progress">
```
carvil: @Davidslv The standard way of loading Javascript modules in GOV.UK is to refer to them on a `data-module` attribute. The module is called `TagUpdateProgress`, so we need to keep this as `tag-update-progress`. More details [here](https://github.com/alphagov/govuk_admin_template/blob/master/JAVASCRIPT.md)

spec/presenters/aggregated_tag_mapping_presenter_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
+
+RSpec.describe AggregatedTagMappingPresenter do
+  let!(:tag_mappings) { [create(:tag_mapping), create(:tag_mapping)] }
+  let(:aggregated_tag_mapping) { TaggingSpreadsheet.first.aggregated_tag_mappings.first } 
+  let(:presenter) { described_class.new(aggregated_tag_mapping) } 
+
+  describe '#errored?' do
+    it 'should return true when any tag mapping state is "errored"' do
+      binding.pry
```
carvil: Could you please remove this `pry`?

app/views/application/_tag_update_preview.html.erb
Davidslv: Fixed

spec/presenters/aggregated_tag_mapping_presenter_spec.rb
Davidslv: (this was removed)

PR #168, opened 2016-09-15
Add error count to tag update table

mobaig:


app/views/application/_tag_update_preview.html.erb
```diff
@@ -4,10 +4,14 @@
   <div data-module="tag-update-progress">
     <% if tag_mappings.any? %>
       <%= render partial: "tag_update_progress_bar",
-        locals: { tag_mappings: tag_mappings, confirmed: confirmed, progress_path: progress_path } %>
+        locals: { tag_mappings: tag_mappings, error_count: error_count, confirmed: confirmed, progress_path: progress_path } %>
```
carvil: Do we need to pass in the `error_count` to this partial? Are there plans to show this in the progress bar somehow?

carvil: This change has been addressed.

PR #170, opened 2016-09-15
Improve error handling

carvil:
This PR fixes:

1) how we display errors in the Tagging Spreadsheet's page

Before:

screen shot 2016-09-15 at 13 54 58

After:

screen shot 2016-09-15 at 13 55 18

2) Makes sure we can store multiple messages as a serialized Array into TagMapping records. With this change, we can store multiple messages in an Array and display it as-is in the view without needing to manipulate a string (by splitting in `.`, which isn't ideal).

3) Updates the `state` and `error_message` of a Tagging Source if at least one TagMapping fails.

4) Adds `error_message` and the `errored` state to TagMigrations, as those can be a Tagging Source. Without this, the point 3) would fail in this case.

Trello: https://trello.com/c/ETZ95krI/176-bug-content-tagger-spreadsheet-errors

### DIFF
--- a/app/models/links_update.rb
+++ b/app/models/links_update.rb
@@ -35,6 +35,6 @@ class LinksUpdate
   def mark_as_errored
     return if errors.messages.blank?
     message = errors.messages.values.join(' ')
-    tag_mappings.update_all(state: :errored, message: message)
+    tag_mappings.update_all(state: :errored, messages: message)
   end
 end
--- a/app/models/links_update.rb
+++ b/app/models/links_update.rb
@@ -35,9 +35,17 @@ class LinksUpdate
   def mark_as_errored
     return if errors.messages.blank?
 
-    tag_mappings.update_all(
-      state: :errored,
-      messages: errors.messages.values.flatten
-    )
+    ActiveRecord::Base.transaction do
+      tag_mappings.update_all(
+        state: :errored,
+        messages: errors.messages.values.flatten
+      )
+
+      tag_mapping = tag_mappings.first
+      tag_mapping.tagging_source.update(
+        state: :errored,
+        error_message: I18n.t('tag_import.errors.tag_mappings_failed')
+      )
+    end
   end
 end
--- a/app/models/links_update.rb
+++ b/app/models/links_update.rb
@@ -34,7 +34,10 @@ class LinksUpdate
 
   def mark_as_errored
     return if errors.messages.blank?
-    message = errors.messages.values.join(' ')
-    tag_mappings.update_all(state: :errored, messages: message)
+
+    tag_mappings.update_all(
+      state: :errored,
+      messages: errors.messages.values.flatten
+    )
   end
 end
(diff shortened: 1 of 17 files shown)


spec/models/links_update_spec.rb
```diff
@@ -77,8 +77,17 @@
         expect(tag_mapping.state).to eql("errored")
       end
 
-      it "concatenates the error messages and sets them on the tag mapping" do
-        expect(tag_mapping.message).to eql("Broken. Rubbish.")
+      it "assigns the error messages to the record" do
+        expect(tag_mapping.messages).to eql(["Broken.", "Rubbish."])
+      end
+
+      it 'does changes the state of the tagging source to errored' do
```
mobaig: can drop the 'does' here and across the following two similar scenarios.

spec/models/links_update_spec.rb
```diff
@@ -94,8 +103,12 @@
         expect { expectation.call }.to_not change { tag_mapping.state }
       end
 
-      it "doesn't change the tag mapping message" do
-        expect { expectation.call }.to_not change { tag_mapping.message }
+      it "doesn't change the tag mapping messages" do
+        expect { expectation.call }.to_not change { tag_mapping.messages }
+      end
+
+      it 'does not change the state of the tagging source' do
```
mobaig: can we make `doesn't` vs `does not`consistent in this file

spec/models/tag_migration_spec.rb
```diff
@@ -1,6 +1,12 @@
 require 'rails_helper'
 
 RSpec.describe TagMigration do
+  describe '#state' do
```
mobaig: it might be worthwhile wrapping this in a `describe "valid states"` block and having similar assertions for the other valid states. Same with `tagging_spreadsheet_spec.rb`

spec/models/links_update_spec.rb
carvil: Yes, that sentence doesn't make sense. I spotted another one like that, fixing now.

spec/models/links_update_spec.rb
carvil: Yup, sticking to `doesn't`

spec/models/tag_migration_spec.rb
carvil: 👍

carvil:
@mobaig Thanks for the review. I've addressed your comments.

app/models/links_update.rb
```diff
@@ -34,7 +34,20 @@ def mark_as_tagged
...
-    tag_mappings.update_all(state: :errored, message: message)
+
+    ActiveRecord::Base.transaction do
+      tag_mappings.update_all(
+        state: :errored,
+        messages: errors.messages.values.flatten
+      )
+
+      tag_mapping = tag_mappings.first
+      return if tag_mapping.blank?
```
mobaig: Could hoist this guard  clause up to the top of the method.

config/locales/en.yml
```diff
@@ -15,6 +15,7 @@ en:
       invalid_content_id: We could not find this URL on GOV.UK.
```
mobaig: Presumably we still get that weird wrapping / bullet point issue with this error? (see screenshot on trello card)

app/models/links_update.rb
carvil: I think we can even get rid of this one, 👍

config/locales/en.yml
carvil: As discussed IRL, this has been addressed.

carvil:
@mobaig comments addressed

PR #171, opened 2016-09-19
Serialize TagMigration#messages as an Array

carvil:
Now that we ran the migration 20160915141004 in production and converted all
pre-existing messages into an Array, we can now enforce the serialization type
to be an Array in order to prevent us from assigning it a different type in the
future.


Davidslv:
Looks good to me 👍

PR #172, opened 2016-09-21
Add taxons to tag search

Davidslv:
Trello: https://trello.com/c/HdJ7VzvF/177-add-taxons-to-content-tagger-s-tag-search

### DIFF
--- a/app/services/bulk_tagging/fetch_tagged_content.rb
+++ b/app/services/bulk_tagging/fetch_tagged_content.rb
@@ -12,6 +12,30 @@ module BulkTagging
     end
 
     def call
+      if taxon_document_type?
+        request_tagged_content_to_taxon
+      else
+        request_tagged_content_items
+      end
+    end
+
+  private
+
+    def taxon_document_type?
+      document_type == "taxon"
+    end
+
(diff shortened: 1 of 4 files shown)


app/services/bulk_tagging/fetch_tagged_content.rb
```diff
@@ -12,6 +12,30 @@ def initialize(content_id, document_type)
...
+      end
+    end
+
+  private
+
+    def taxon_document_type?
+      document_type == "taxon"
+    end
+
+    def request_tagged_content_to_taxon
```
tijmenb: Cool, using `get_linked_items` seems better than using `get_expanded_links`. Could you change the current implementation to use it too?

tijmenb:
Looks good!

app/services/bulk_tagging/fetch_tagged_content.rb
Davidslv: Thanks @tijmenb, I've changed it, you can see it in this commit: https://github.com/alphagov/content-tagger/pull/172/commits/8c87e8057926ed5c0485d5519729a47b4c9b0709

PR #174, opened 2016-09-26
Simplify data flow for tagging migration creation

tijmenb:
This is an attempt to simplify the controller data flow when creating a `TagMigration`.

Instead of passing on the base path and document type of the thing we're retagging, we just pass on the content_id (`source_content_id`) and query the publishing-api whenever we need something from the content item.

This means there's less data passed down through the URL and tag migration form, making it easier to follow. We can also link more easily to the retagging page (`http://content-tagger.dev.gov.uk/tag_migrations/new?source_content_id=CONTENT_ID`). It didn't seem useful to keep the `query`.

The `source_description` is not strictly necessary, because on the index & show pages we could query the publishing-api and generate it with that. But this would cause a publishing-api request per `TagMigration` so pre-calculating and persisting it locally is a lot faster.

## Before

1. In the view for `tag_searches#new`, we pass in `document_type`, `query` and `source_base_path` and `source_content_id` to the path helper, which goes to `tag_migrations#new`
2. `tag_migrations#new` passes them into the form and sends them on submit to `tag_migrations#create`
3. `tag_migrations#create` takes the params and pass them into `BuildTagMigration`, which saves them to the database.

screen shot 2016-09-26 at 18 55 47
screen shot 2016-09-26 at 18 55 50
screen shot 2016-09-26 at 18 55 54

## After

1. In `tag_searches#new`, we only pass through `source_content_id`
2. `tag_migrations#new` uses the `source_content_id` to infer `document_type`, so that it can load tagged content
3. `tag_migrations#create` uses `source_content_id` to create a human-readable `source_description`.

screen shot 2016-09-26 at 18 53 02
(note the url)
screen shot 2016-09-26 at 18 53 05
screen shot 2016-09-26 at 18 53 13

### DIFF
--- a/app/controllers/tag_migrations_controller.rb
+++ b/app/controllers/tag_migrations_controller.rb
@@ -4,35 +4,41 @@ class TagMigrationsController < ApplicationController
   end
 
   def new
-    unless tag_migration_params_present?
+    unless params[:source_content_id]
       redirect_to new_tag_search_path
       return
     end
 
+    source_content_item = ContentItem.find!(params[:source_content_id])
+
     expanded_links = BulkTagging::FetchTaggedContent.call(
-      tag_content_id: tag_migration_params[:source_content_id],
-      tag_document_type: tag_migration_params[:document_type],
+      tag_content_id: source_content_item.content_id,
+      tag_document_type: source_content_item.document_type,
     )
+
     taxons = RemoteTaxons.new.all
 
     render :new, locals: {
-      tag_migration: TagMigration.new(tag_migration_params),
+      tag_migration: TagMigration.new(source_content_id: source_content_item.content_id),
       taxons: taxons,
       expanded_links: expanded_links,
     }
   end
 
   def create
+    source_content_item = ContentItem.find!(params[:tag_migration][:source_content_id])
+
     tag_migration = BulkTagging::BuildTagMigration.call(
-      tag_migration_params: tag_migration_params,
+      source_content_item: source_content_item,
       taxon_content_ids: params[:taxons],
       content_base_paths: params[:content_base_paths],
     )
+
     tag_migration.save!
     redirect_to tag_migration_path(tag_migration)
   rescue BulkTagging::BuildTagMigration::InvalidArgumentError => e
     flash[:error] = e.message
-    redirect_to new_tag_migration_path(tag_migration: tag_migration_params)
+    redirect_to new_tag_migration_path(source_content_id: source_content_item.content_id)
   end
 
   def show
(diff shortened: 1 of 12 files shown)


app/controllers/tag_migrations_controller.rb
```diff
@@ -4,35 +4,41 @@ def index
...
-      tag_document_type: tag_migration_params[:document_type],
+      tag_content_id: source_content_item.content_id,
+      tag_document_type: source_content_item.document_type,
     )
+
     taxons = RemoteTaxons.new.all
 
     render :new, locals: {
-      tag_migration: TagMigration.new(tag_migration_params),
+      tag_migration: TagMigration.new(source_content_id: params[:source_content_id]),
```
Davidslv: You could continue using `source_content_item.content_id` for consistency

app/controllers/tag_migrations_controller.rb
```diff
@@ -4,35 +4,41 @@ def index
...
       taxon_content_ids: params[:taxons],
       content_base_paths: params[:content_base_paths],
     )
+
     tag_migration.save!
     redirect_to tag_migration_path(tag_migration)
   rescue BulkTagging::BuildTagMigration::InvalidArgumentError => e
     flash[:error] = e.message
-    redirect_to new_tag_migration_path(tag_migration: tag_migration_params)
+    redirect_to new_tag_migration_path(source_content_id: params[:tag_migration][:source_content_id])
```
Davidslv: can you replace it with `source_content_item.content_id` here too?

app/services/bulk_tagging/build_tag_migration.rb
```diff
@@ -2,20 +2,16 @@ module BulkTagging
   class BuildTagMigration
     class InvalidArgumentError < ArgumentError; end
 
-    attr_reader :tag_migration_params, :taxon_content_ids, :content_base_paths
+    attr_reader :source_content_item, :taxon_content_ids, :content_base_paths
 
-    def initialize(tag_migration_params:, taxon_content_ids:, content_base_paths:)
-      @tag_migration_params = tag_migration_params
+    def initialize(source_content_item, taxon_content_ids, content_base_paths)
```
Davidslv: I think you should keep the signature of the method, its easier to read when you specify which parameter is which

app/services/bulk_tagging/build_tag_migration.rb
```diff
@@ -2,20 +2,16 @@ module BulkTagging
...
       @content_base_paths = content_base_paths
     end
 
-    def self.call(tag_migration_params:, taxon_content_ids:, content_base_paths:)
-      new(
-        tag_migration_params: tag_migration_params,
-        taxon_content_ids: taxon_content_ids,
-        content_base_paths: content_base_paths
-      ).call
+    def self.call(source_content_item:, taxon_content_ids:, content_base_paths:)
```
Davidslv: I still don't understand why we doing all this _dance_ of playing with a class method that calls the constructor and then uses the instance of `call`

This could had all been invoked like: `BuildTagMigration.new(arguments_here).call` and we would have less mental stress reading this code.

tijmenb:
@Davidslv Updated commit.

app/services/bulk_tagging/build_tag_migration.rb
tijmenb: I agree - the pattern makes it hard to refactor. I've kept it to keep it consistent. Changing it can be done in another PR since I'd like to discuss it with @carvil / @mobaig first.

PR #175, opened 2016-09-29
Clear parent links

carvil:
This PR allows us to clear the `parent` link of all our Taxons via the `PublishLinks` service class.

By specifically setting a link type to be an empty array, we are able to clear existing links for that link type. Otherwise, `PublishLinks` works as usual by appending links.

Furthermore, this PR adds a rake task that talks to the publishing API and allows us to clear all existing parent links for all taxons. We only have 100 taxons at the moment, which means that performing this via a rake task should be fine.

Trello task: https://trello.com/c/BAyP2r6c/187-fix-outdated-link-expansion-of-parent-taxons-clean-up-parent

```
irb(main):001:0> content_id = "e184c8d6-7142-4637-ae1e-3270247e0f4f"
=> "e184c8d6-7142-4637-ae1e-3270247e0f4f"
irb(main):002:0> Services.publishing_api.get_links(content_id)
=> #<GdsApi::Response:0x007fd9a7fbd668 @http_response=<RestClient::Response 200 "{\"content_i...">>
irb(main):003:0> Services.publishing_api.get_links(content_id)['links']
=> {"parent"=>["502eac33-76ad-499e-b15f-a1544be548c5"], "parent_taxons"=>["502eac33-76ad-499e-b15f-a1544be548c5"]}
irb(main):004:0> Taxonomy::ClearParentLink.call(content_id)
=> #<GdsApi::Response:0x007fd9a809a630 @http_response=<RestClient::Response 200 "{\"content_i...">>
irb(main):005:0> Services.publishing_api.get_links(content_id)['links']
=> {"parent_taxons"=>["502eac33-76ad-499e-b15f-a1544be548c5"]}
irb(main):006:0>
```

### DIFF
--- a/app/models/taxon_parent_links_update.rb
+++ b/app/models/taxon_parent_links_update.rb
@@ -1,0 +1,11 @@
+class TaxonParentLinksUpdate
+  attr_reader :content_id
+
+  def initialize(content_id)
+    @content_id = content_id
+  end
+
+  def links_to_update
+    { 'parent' => [] }
+  end
+end
--- a/app/services/publish_links.rb
+++ b/app/services/publish_links.rb
@@ -21,7 +21,11 @@ private
 
   def updated_links
     links_update.links_to_update.merge(existing_links) do |_, new_links, old_links|
-      (old_links || []).concat(new_links).uniq
+      if new_links.empty?
+        []
+      else
+        (old_links || []).concat(new_links).uniq
+      end
     end
   end
 
--- a/app/services/publish_links.rb
+++ b/app/services/publish_links.rb
@@ -10,16 +10,11 @@ class PublishLinks
   end
 
   def publish
-    if links_update.valid?
-      Services.publishing_api.patch_links(
-        links_update.content_id,
-        links: updated_links,
-        previous_version: previous_version
-      )
-      links_update.mark_as_tagged
-    else
-      links_update.mark_as_errored
-    end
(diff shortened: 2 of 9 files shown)


app/models/taxon_parent_links_update.rb
```diff
@@ -0,0 +1,11 @@
+class TaxonParentLinksUpdate
```
tijmenb: Is two classes perhaps overkill, since we're only going to clear the `parent` once?

carvil: This is so we can re-use `PublishLinks`. This class has the same interface as `LinksUpdate`, which means `PublishLinks` doesn't have to be changed. Does that make sense?

tijmenb: What I was trying to say is that I'm a bit worried about the growth of content-tagger as a whole. It looks like clearing of the parent links could be reduced to a couple of lines in the rake task. Extracting some parts it into `ClearParentLink` and `TaxonParentLinksUpdate` seems to introduce a lot of indirection. Especially because those classes are single-use and will never be called again after running it now. WDYT?

carvil:
@tijmenb Sorry, I should have been clearer. These classes should be deleted after we run the rake task. I added these classes so I could test how this would work with our existing service classes and make sure it would do the right thing. They don't need to stay in the codebase after we run this. But if prefer, I can rework this PR now...

PR #176, opened 2016-09-30
Don't validate taxons before sending

tijmenb:
This validation doesn't function properly at the moment. It will use the `RemoteTaxons` class to check if the taxon we're trying to tag actually exists. Since `RemoteTaxons` only returns a subset of taxons, this will result in a false negative and show an error to the user.

As far as I can see, this validation is run when sending a number of taggings to the publishing-api via the "import spreadsheet" feature.

I think this validation is not necessary at this point in the import since we've already validated the existence of the taxons during the initial import of the spreadsheet.

When it's used by the "tag migrations", we don't need the validation because the source of the taxon `content_id` is our own form. I don't think the user can do anything to do this wrong.

cc @mobaig or @carvil


app/validators/taxons_validator.rb
```diff
@@ -1,15 +0,0 @@
-class TaxonsValidator < ActiveModel::Validator
-  def validate(record)
-    return if record.taxons.empty?
-
-    unless (record.taxons - known_taxon_content_ids).empty?
-      record.errors[:taxons] << I18n.t('tag_import.errors.invalid_taxons_found')
```
carvil: Could you please also delete this I18n entry?

tijmenb: Good catch! Amended.

PR #178, opened 2016-09-30
Use linkables for tag migrations dropdown

tijmenb:
This changes the way we populate the select-a-taxon dropdown in the tag migration screen. Linkables is the appropriate class / API call for this, since it's already used for other dropdowns. This means that it uses the `internal_name` property (which disambiguates taxons with the same name) and shows the taxons alphabetically.

### DIFF
--- a/spec/support/content_item_helper.rb
+++ b/spec/support/content_item_helper.rb
@@ -7,4 +7,17 @@ module ContentItemHelper
       document_type: "guidance",
     ).merge(other_fields)
   end
+
+  def build_linkable(hash)
+    default = {
+      content_id: SecureRandom.uuid,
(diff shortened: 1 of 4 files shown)


spec/support/content_item_helper.rb
```diff
@@ -7,4 +7,17 @@ def basic_content_item(title, other_fields: {})
       document_type: "guidance",
     ).merge(other_fields)
   end
+
+  def build_linkable(hash)
```
tijmenb: Note that this only returns random content by default. I think that's better than hard coded values, since it requires the tests to be explicit about what they expect the helper method to produce. Perhaps `basic_content_item` should be rewritten to to something like it.

PR #179, opened 2016-09-30
Add pagination to taxons

carvil:
This PR adds pagination to the taxons' index page.

Trello: https://trello.com/c/0fUIlIQ2/202-add-pagination-to-taxon-index-page

It looks like this:

screen shot 2016-09-30 at 13 40 07


Gemfile
```diff
@@ -19,6 +19,9 @@ gem 'select2-rails', '~> 3.5.9'
 
 gem 'govuk_sidekiq', '~> 0.0.4'
 
+gem 'kaminari'
```
tijmenb: Can you pin these?

carvil: Yup, I've pushed the change that includes pinned versions of these.

PR #180, opened 2016-09-30
Use `ContentItem.find` for fetching taxons in tag migration

tijmenb:
`RemoteTaxons` currently has a bug, so using the existing `ContentItem` class is better.


tijmenb:
Closed as @carvil has a PR for the same thing.

PR #188, opened 2016-10-03
Remove callout style from new/edit taxon pages

tijmenb:
The taxon new & edit pages are currently displayed with the "callout" component from `govuk_admin_template`. This component is meant to be used as a warning to the user, not as a layout choice. We've not used this in any other application.

Removing it will 1) allow us to improve the display of all pages more easily, 2) remove a lot of code, 3) make the app consistent with other applications and 4) use the callout style for its original purpose.

This was originally part of https://trello.com/c/Erx6NIwe. 

Extracting this out to make that PR easier to review.


tijmenb:
cc @carvil

PR #186, opened 2016-09-30
Improve taxon deletion workflow for taxons

Davidslv:
Trello: https://trello.com/c/ODbysu1S/178-content-tagger-improve-taxon-deletion-workflow-for-taxons

### Taxon with child taxons and tagged content
screen shot 2016-10-04 at 15 16 53

### Taxon with no child taxons or tagged content
screen shot 2016-10-03 at 11 19 21

PR #190, opened 2016-10-07
Revert "Improve taxon deletion workflow for taxons"

tijmenb:
Reverts alphagov/content-tagger#186.

There's a bug in this workflow that prevents deleting the taxons:

https://errbit.integration.publishing.service.gov.uk/apps/56605a3d6578633ef4030000/problems/57f7c3d56578635dc9800000


carvil:
👍

PR #191, opened 2016-10-10
Dont show taxons without internal name

carvil:
We need to filter out content items without an internal name (both `nil` and empty string).

Trello: https://trello.com/c/JNJ4nYcL/230-bug-not-able-to-add-parent-taxon


tijmenb:
Background to this: because of a bug in publishing-api it's returning incorrect data.

Publishing Platform story: https://trello.com/c/zCwGn8vo

After it's fixed we should remove this `select` from our code base, because we shouldn't be this defensive when receiving data from the publishing-api. Rather, we should crash with an appropriate exception when publishing-api starts sending incorrect data.

PR #192, opened 2016-10-10
Solves bug related with taxon deletion workflow

Davidslv:
This was reverted here: https://github.com/alphagov/content-tagger/pull/190

The problem was that the wrong method was being called
(taxon.child_taxons instead of children here: https://github.com/alphagov/content-tagger/compare/solve-taxons-deletion-workflow?expand=1#diff-8258490cf5fbd5fae1b039d4d51ded74R1)

### DIFF
--- a/app/views/taxons/_tagged_content.html.erb
+++ b/app/views/taxons/_tagged_content.html.erb
@@ -1,9 +1,11 @@
+<h1>Tagged Content</h1>
+
 <table class="table queries-list table-bordered table-striped" data-module="filterable-table">
   <thead>
     <tr class="table-header">
       <th>Title</th>
-      <th></th>
-      <th></th>
+      <th>Format</th>
+      <th>Actions</th>
     </tr>
 
     <%= render partial: 'shared/table_filter' %>
@@ -12,9 +14,9 @@
   <tbody>
     <% tagged.each do |content_item| %>
       <tr>
-        <td><%= content_item["title"] %> (<%= content_item["document_type"].humanize %>)<br/></td>
-        <td><%= link_to "View on site", website_url(content_item["base_path"]) %></td>
-        <td><%= link_to I18n.t('views.taxons.update_tags'), tagging_url(content_item["content_id"]) %></td>
+        <td><%= link_to content_item["title"], website_url(content_item["base_path"]) %></td>
+        <td><%= content_item["document_type"].humanize %></td>
+        <td><%= link_to 'Update tags', tagging_url(content_item["content_id"]) %></td>
       </tr>
     <% end %>
   </tbody>
--- a/app/views/taxons/confirm_delete.html.erb
+++ b/app/views/taxons/confirm_delete.html.erb
@@ -1,0 +1,12 @@
+<h1>You are about to delete "<%= taxon.title %>"</h1>
+
+<%= render partial: 'deletion_warning',
+  locals: { taxon: taxon, children: children, tagged: tagged } %>
+
+<% if children.any? %>
+  <%= render partial: 'children', locals: { children: children } %>
+<% end %>
+
+<% if tagged.any? %>
+  <%= render partial: 'tagged_content', locals: { tagged: tagged } %>
+<% end %>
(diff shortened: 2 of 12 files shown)


app/views/taxons/_tagged_content.html.erb
```diff
@@ -12,9 +14,9 @@
   <tbody>
     <% tagged.each do |content_item| %>
       <tr>
-        <td><%= content_item["title"] %> (<%= content_item["document_type"].humanize %>)<br/></td>
-        <td><%= link_to "View on site", website_url(content_item["base_path"]) %></td>
-        <td><%= link_to I18n.t('views.taxons.update_tags'), tagging_url(content_item["content_id"]) %></td>
+        <td><%= link_to content_item["title"], website_url(content_item["base_path"]) %></td>
+        <td><%= content_item["document_type"].humanize %></td>
+        <td><%= link_to 'Update tags', tagging_url(content_item["content_id"]) %></td>
```
carvil: Could you please use I18n here?

app/views/taxons/confirm_delete.html.erb
```diff
@@ -0,0 +1,12 @@
+<h1>You are about to delete "<%= taxon.title %>"</h1>
+
+<%= render partial: 'deletion_warning',
+  locals: { taxon: taxon, children: children, tagged: tagged } %>
+
+<% if children.any? %>
```
carvil: What do you think of adding a message in case there aren't any children to display?

app/views/taxons/confirm_delete.html.erb
```diff
@@ -0,0 +1,12 @@
...
+  <%= render partial: 'children', locals: { children: children } %>
+<% end %>
+
+<% if tagged.any? %>
```
carvil: What do you think of adding a message in case there isn't content to display?

spec/features/delete_taxon_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+  include ContentItemHelper
+  include PublishingApiHelper
+
+  scenario "a taxon with no children" do
+    given_a_taxon_with_no_children
+    when_i_visit_the_taxon_page
+    and_i_click_to_delete_the_taxon
+    then_i_expect_to_be_informed_that_im_about_to_delete_the_taxon
+    and_i_should_see_the_action_buttons
+    and_i_click_delete
```
carvil: What do you think of renaming this to `when_i_click_delete`, as it's an action we are performing rather than an assertion?

spec/features/delete_taxon_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+    then_i_expect_the_request_to_have_been_made
+  end
+
+  scenario "a parent taxon with children" do
```
carvil: What do you think of including a couple of content items as well? That way we would be able to assert that they are also on the page together with the children.

spec/features/delete_taxon_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+    then_i_should_see_a_warning_message
+    and_i_should_see_the_action_buttons
+    and_i_click_delete
+    then_i_expect_the_request_to_have_been_made
+  end
+
+  def given_a_taxon_with_no_children
+    @content_id = SecureRandom.uuid
+
+    @taxon = {
```
carvil: There is a helper method to generate this for us in https://github.com/alphagov/content-tagger/blob/master/spec/support/content_item_helper.rb#L2

spec/features/delete_taxon_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+      content_id: @content_id,
+      expanded_links: {
+        documents: []
+      }
+    )
+  end
+
+  def a_child_taxon
+    @child_taxon_content_id = SecureRandom.uuid
+    @child_taxon = {
```
carvil: We could use the helper method I mentioned above to generate this: https://github.com/alphagov/content-tagger/blob/master/spec/support/content_item_helper.rb#L2

spec/features/delete_taxon_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+
+    publishing_api_has_expanded_links(
+      content_id: @content_id,
+      expanded_links: {
+        documents: []
+      }
+    )
+  end
+
+  def a_child_taxon
```
carvil: What do you think of renaming this step to something like `a_child_taxon_exists`?

PR #194, opened 2016-10-11
Make naming of taxons clearer

Davidslv:
Trello: https://trello.com/c/jZtVkf4H/241-use-internal-name-everywhere-in-content-tagger

screen shot 2016-10-12 at 10 32 36
screen shot 2016-10-12 at 10 29 34

### DIFF
--- a/app/views/taxons/_form_fields.html.erb
+++ b/app/views/taxons/_form_fields.html.erb
@@ -1,9 +1,17 @@
-<%= f.input :title, input_html: { class: 'form-control' }, label: 'Title' %>
-<%= f.input :internal_name, input_html: { class: 'form-control' } %>
+<%= f.input :internal_name, input_html: { class: 'form-control' },
+  label: I18n.t('views.taxons.internal_name'),
+  hint: I18n.t('views.taxons.internal_name_hint') %>
+
+<%= f.input :title, input_html: { class: 'form-control' },
+  label: I18n.t('views.taxons.external_name'),
+  hint: I18n.t('views.taxons.displayed_on_govuk') %>
+
 <%= f.input :description,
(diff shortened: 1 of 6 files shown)


app/views/taxons/_form_fields.html.erb
```diff
@@ -1,4 +1,6 @@
-<%= f.input :title, input_html: { class: 'form-control' }, label: 'Title' %>
+<%= f.input :title, input_html: { class: 'form-control' },
+  label: 'Title (External name)',
+  hint: 'This is the name that the public will see' %>
```
carvil: Could you please move these strings to I18n?

lib/tasks/copy_taxon_title.rake
```diff
@@ -0,0 +1,12 @@
+desc "copy taxons title field to internal name"
+task copy_taxons_title: :environment do
```
carvil: This looks good. I was under the impression the Publishing API did this already. 

It looks like they do it for [linkables](https://github.com/alphagov/publishing-api/blob/012cb3f1ceb3b18e7059a367cc4030aa0763afb4/app/presenters/queries/linkable_presenter.rb#L10), but I can't see it being done for `get_content_items`: 

https://github.com/alphagov/publishing-api/blob/012cb3f1ceb3b18e7059a367cc4030aa0763afb4/app/presenters/queries/content_item_presenter.rb

It might be good to ask them why that is before we run this rake task.

Davidslv:
I've updated this PR

lib/tasks/copy_taxon_title.rake
carvil: 👍

PR #195, opened 2016-10-11
Add feature to move content items between taxons

carvil:
This feature allows us to move content items between taxons. It is similar to bulk tagging, however it deletes the link of the source taxon, keeping only the new link.

If we have tagged both `Content Item 1` and `Content Item 2` with `Taxon A`, when we move the content items over to `Taxon B` using this feature, the original link to `Taxon A` will be deleted.

Trello: https://trello.com/c/YRnivAny/214-content-tagger-allow-ias-to-move-content-from-taxon-to-taxon

Thanks @klssmith for pairing with me on this

Here is how this looks, screen by screen:

1) First we select the taxon we want from the index page

screen shot 2016-10-11 at 15 13 30

2) We then click `Move Content`

screen shot 2016-10-11 at 15 13 41

3) In the next screen, we select the destination taxon and which content items we wish to move across

screen shot 2016-10-11 at 15 13 51

4) After reading the warning message, we then start tagging

screen shot 2016-10-11 at 15 13 58

5) On success, it looks like a normal tag migration

screen shot 2016-10-11 at 15 14 09

6) When going to those 2 content items' pages, we can see only the new link is kept

screen shot 2016-10-11 at 15 14 30

and

screen shot 2016-10-11 at 15 14 39

### DIFF
--- a/app/controllers/taxon_migrations_controller.rb
+++ b/app/controllers/taxon_migrations_controller.rb
@@ -1,0 +1,39 @@
+class TaxonMigrationsController < ApplicationController
+  def new
+    unless params[:source_content_id]
+      redirect_to taxons_path
+      return
+    end
+
+    source_content_item = ContentItem.find!(params[:source_content_id])
+
+    expanded_links = BulkTagging::FetchTaggedContent.call(
+      tag_content_id: source_content_item.content_id,
+      tag_document_type: source_content_item.document_type,
+    )
+
+    render :new, locals: {
+      tag_migration: TagMigration.new(source_content_id: source_content_item.content_id),
+      taxons: Linkables.new.taxons,
+      expanded_links: expanded_links,
+      source_content_item: source_content_item,
+    }
+  end
+
+  def create
+    source_content_item = ContentItem.find!(params[:tag_migration][:source_content_id])
+
+    tag_migration = BulkTagging::BuildTagMigration.call(
+      source_content_item: source_content_item,
+      taxon_content_ids: params[:taxons],
+      content_base_paths: params[:content_base_paths],
+    )
+    tag_migration.delete_source_link = true
+    tag_migration.save!
+
+    redirect_to tag_migration_path(tag_migration)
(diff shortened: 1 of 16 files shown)


mobaig:
@carvil can you reformat the commit messages to 72 char width? It's a fairly widespread convention, to the degree that Github expects this when formatting and displaying commits. Any longer and you get weird looking text wrapping which isn't great for readability.

app/services/publish_links.rb
```diff
@@ -27,7 +31,14 @@ def links_to_update
...
     links_to_update.merge(existing_links) do |_, new_links, old_links|
-      (old_links || []).concat(new_links).uniq
+      old_links_for_type =
+        if tag_mapping.delete_source_link?
+          old_links - [source_content_id]
+        else
+          old_links
+        end
+
+      (old_links_for_type || []).concat(new_links).uniq
```
jackscotti: can `old_links_for_type` be `nil`?

app/controllers/taxon_migrations_controller.rb
```diff
@@ -0,0 +1,39 @@
+class TaxonMigrationsController < ApplicationController
+  def new
+    unless params[:source_content_id]
+      redirect_to taxons_path
+      return
```
carvil: @jackscotti Yup, see http://api.rubyonrails.org/classes/ActionController/Base.html

```
If you need to redirect on the condition of something, then be sure to add “and return” to halt execution.
```

jackscotti: I see, I thought the redirection included an implicit return statement.

app/services/publish_links.rb
carvil: @jackscotti I take that back, it isn't a bug. The reason `old_links_for_type` can be nil because when we are tagging a given content item for the first time and there are no links to return. The links hash is empty and therefore `old_links` is nil. Does that make sense?

carvil:
@mobaig I've reformatted the message now.

jackscotti:
looks good to me, @mobaig ?

PR #196, opened 2016-10-12
Paginate bulk tag search

carvil:
When searching for tags in the bulk tagging feature, the results will
now be paginated instead of showing a warning messages about multiple
results.

Trello:
https://trello.com/c/HkbPffzv/242-content-tagger-add-pagination-to-bulk-tag-search

### DIFF
--- a/app/models/empty_search_response.rb
+++ b/app/models/empty_search_response.rb
@@ -1,0 +1,17 @@
+class EmptySearchResponse
+  def results
+    []
+  end
+
+  def current_page
+    1
+  end
+
+  def total_pages
+    0
+  end
+
+  def limit_value
+    5
+  end
+end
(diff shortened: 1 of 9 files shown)


app/models/empty_search_response.rb
```diff
@@ -0,0 +1,17 @@
+class EmptySearchResponse
```
jackscotti: why is this not inheriting from `SearchResponse` or vice-versa?

carvil: This class is exists to show the initial state of that page. Making it inherit from `SearchResponse` wouldn't make us write less code since we would need to override all the methods anyway (apart from the last one). By keeping it separate there is less code to browse when looking around. Does that make sense @jackscotti ?

PR #197, opened 2016-10-12
Use taxon internal name

Davidslv:
Trello: https://trello.com/c/jZtVkf4H/241-use-internal-name-everywhere-in-content-tagger

When the internal_name is not available it will fallback to use the title.
This shouldn't be a problem after running the rake task introduced in 2d56cf8d57d417247a396848971de8ea8f8807b0

Although it will fallback to use the title when `expanded links` is requested from Publishing API,
that happens on [ExpandedTaxonomy](https://github.com/alphagov/content-tagger/blob/master/app/models/expanded_taxonomy.rb#L28)

Paired with @klssmith

### DIFF
--- a/app/models/taxon.rb
+++ b/app/models/taxon.rb
@@ -31,6 +31,10 @@ class Taxon
     'taxons'
   end
 
+  def internal_name
+    @internal_name || @title
+  end
+
   def notes_for_editors
     @notes_for_editors || ""
   end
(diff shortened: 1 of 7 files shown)


app/models/taxon.rb
```diff
@@ -30,4 +30,8 @@ def base_path
   def link_type
     'taxons'
   end
+
+  def internal_name
```
carvil: Could you please add a test case to these model about this behaviour? Just in case it gets changed in the future.

Davidslv: @carvil I've added it. Thanks 👍

carvil:
Nice one @Davidslv

PR #201, opened 2016-10-12
Handle error messages from Publishing API

klssmith:
If Publishing API returns an error, then a message is now provided to the
user to show why it failed. Previously, the error caused Content-Tagger to
 crash.

[Trello card](https://trello.com/c/VBe1sCVb/238-handle-error-messages-from-publishing-api)

screen shot 2016-10-12 at 14 55 57

### DIFF
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -47,7 +47,9 @@ class TaxonsController < ApplicationController
   def destroy
     response_code = Services.publishing_api.unpublish(params[:id], type: "gone").code
 
-    redirect_to taxons_path, flash: destroy_flash_message(response_code)
+    redirect_to taxon_path(params[:id]), flash: destroy_flash_message(response_code)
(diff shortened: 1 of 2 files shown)


tijmenb:
Interesting. 

Two points:

- This message doesn't tell the user what to do about the error, so maybe we _should_ crash, so that a developer can take a look at it?
- The specific error "Cannot unpublish with draft present" is odd. We always send a content item and then call publish, so there should never be a draft present (if this is what that means). Is this a bug in the publishing-api?

Davidslv:
I've noticed this yesterday when I tried to delete a taxon: `/taxons/b304ec93-5dc1-4799-9127-2f4dd947154a`

I'm not sure if that's a bug in Publishing API, maybe we should ask them

tijmenb:
Talked about this IRL. We don't want to show the errors to the user, because they won't know what to do with it. We also need to be warned when this kind of error occurs, since in most cases it's something we did wrong, rather than the user.

PR #202, opened 2016-10-13
Don't display unpublished (deleted) taxons on index page

mobaig:
A user wouldn't expect deleted taxons to still appear in this view or in
search results.

https://trello.com/c/7CDYowYT/240-remove-unpublished-taxons-from-the-taxons-index-page


app/models/taxon_search_results.rb
```diff
@@ -8,9 +8,12 @@ def initialize(search_response)
   def taxons
     @taxons ||=
       begin
-        search_response['results'].map do |taxon_hash|
+        results = search_response['results'].map do |taxon_hash|
+          next if taxon_hash['publication_state'] == 'unpublished'
```
carvil: What do you think of writing a test case on `taxon_search_results_spec.rb` to make sure this line doesn't get deleted in the future?

mobaig: The feature spec exercises this behaviour, but yeah a unit test makes sense. Adding...

mobaig: @carvil updated.

PR #207, opened 2016-10-18
Text changes to tagmigration show page

klssmith:
Changes made to the text on the tagmigration show page as part of [this Trello card.](https://trello.com/c/rYqgodpD/249-content-tagger-text-changes-to-tagmigration-show-page)

Paired with @Davidslv

### DIFF
--- a/app/controllers/tag_migrations_controller.rb
+++ b/app/controllers/tag_migrations_controller.rb
@@ -43,7 +43,7 @@ class TagMigrationsController < ApplicationController
   def show
     render :show, locals: {
       tag_migration: tag_migration,
-      current_tagged_taxon: tag_migration.tag_mappings.first.link_title,
+      current_tagged_taxon: tag_migration.source_title,
       aggregated_tag_mappings: presented_aggregated_tag_mappings,
       completed_tag_mappings: aggregated_tag_mappings.sum(&:completed_tag_mappings),
       total_tag_mappings: aggregated_tag_mappings.sum(&:total_tag_mappings),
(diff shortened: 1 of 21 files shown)


jackscotti:
👏 well done!

app/controllers/tag_migrations_controller.rb
```diff
@@ -43,6 +43,7 @@ def create
   def show
     render :show, locals: {
       tag_migration: tag_migration,
+      current_tagged_taxon: tag_migration.tag_mappings.first.link_title,
```
carvil: Reading through this I think we could change a couple of things a bit to make this easier (and avoid passing another param to the view or have too much indirection). That involves refactoring things a little bit.

I think we have 2 options:

## Fetch the source content item via the API

In this case, we could:
1) remove the `source_description` from `TagMigration`
2) add a new method to `TagMigration`, which would look like this

```
def source_content_item
  @source_content_item ||= Services.publishing_api.get_content(source_content_id)
end
```

3) and turn `source_description` into a method in the `TagMigration` class so we keep supporting it, and change it to:

```
def source_description
  "#{source_content_item.title} (#{source_document_type.humanize})"
end
```
4) finally, add a method for the title:

```
def source_title
  source_content_item.title
end
```

## Add extra attributes to `TagMigration` so we store that information when we build the object

Currently, a `TagMigration` has a `source_content_id` and a `source_description`. The `source_description` attributes looks a lot more like a presenter method rather than something we should store in the database.

What do you think of:
1) create a migration that adds both `source_title` and `source_document_type`
2) change BulkTagging::BuildTagMigration so we set those new fields with the correct data
3) remove `source_description` from `TagMigration` and convert it into a method like suggested in the other option

Regardless of the solutions above (or others you might think of), we could then refer to `source_title` in the view via the `tag_migration` object.

Sorry for the long comment :) What do you think?

klssmith:
@carvil thanks for your comments, I've addressed it by using your suggested 2nd option.

carvil:
Nice one @klssmith it looks great! As discussed IRL, we should probably write a rake task to migrate existing TagMigrations across to the new format (by populating the new fields).

Davidslv:
@carvil as discussed there's only one TagMigration that happens to have a `source_description` with a `nil` value. We decided there was no need to create a rake task for such.

Thanks 👍

PR #210, opened 2016-10-20
Move "Bulk tag by upload" into "Bulk Tag"

Davidslv:
Trello: https://trello.com/c/XZGZNtHb/213-content-tagger-updates-to-page-titles-and-labelling

We belive this functionality should move into the Bulk Tag tab.

Although having it directly within the page would make it confusing when
you search for content to be tagged and also see the list that comes
from the uploaded content. That's why I've created a "button" to point
to the "Bulk tag by upload"

Also the reason to make it a button is that it stands out in the eye of
the user instead of a link, which looks tiny and reads weird next to
"Bulk tag history".

screen shot 2016-10-20 at 11 58 08

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -13,11 +13,7 @@
     <%= link_to t('navigation.tag_search'), new_tag_search_path %>
   </li>
 
-  <li class='<%= active_navigation_item == 'tagging_spreadsheets' ? 'active' : nil %>'>
-    <%= link_to t('navigation.tag_importer'), tagging_spreadsheets_path %>
-  </li>
-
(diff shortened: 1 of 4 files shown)


tijmenb:
Should https://github.com/alphagov/content-tagger/pull/210/commits/931e30627a8c00062ebe2afad65ebf4ff8bcd123 be in a separate PR?

Davidslv:
@tijmenb Indeed, I've move it here https://github.com/alphagov/content-tagger/pull/211

carvil:
@Davidslv this is ok to merge, please rebase it first

Davidslv:
Thanks @carvil, just rebased

PR #214, opened 2016-10-21
Update taxonomy visualisation

mobaig:
This commit updates the front end code and underlying data structure of
the taxonomy visualisation feature. Previously we could only see a basic
tree structure detailing the child taxons and immediate parents of a
given taxon. The updated design allows us to see all ancestors and all
successors in a clearer layout.

- Update ExpandedTaxonomy so that it now provides an abstraction over
  two trees - the fully expanded set of parents and the fully expanded
  set of children. Also includes convenience methods for querying data
  about the root node and its immediate parents/children.
- Rework the HTML and CSS to accommodate the new design and additional
  data being displayed.
- Update the viewing_taxons feature spec to reflect the new behaviour.
- Some minor naming clarification in the TreeNode class.

See individual commits for details of other, smaller changes.

https://trello.com/b/GBkshanv/taxonomy-team

Below are some screenshots of different shapes the visualisation can take depending on the structure of the taxonomy. There are some differences from the original design, largely to make the markup more robust when dealing with wildly different structures and multiple levels of nesting.

## Multiple parents

screen shot 2016-10-21 at 10 08 44

## Single parent

screen shot 2016-10-21 at 10 09 07

## Large imbalance in the number of child nodes

screen shot 2016-10-21 at 10 13 56

### DIFF
--- a/app/views/taxons/_taxonomy_tree.html.erb
+++ b/app/views/taxons/_taxonomy_tree.html.erb
@@ -1,22 +1,52 @@
 <div class="taxonomy-tree">
-  <% taxonomy_tree.each do |tree_node| %>
-    <div class="taxon-level taxon-depth-<%= tree_node.node_depth %>">
-      <span class="taxon-level-title">
-        <%= link_to tree_node.title, taxon_path(tree_node.content_id) %>
-      </span>
-
-      <% parent_taxons = tree_node.parent_taxons %>
-
-      <% if parent_taxons.present? && (parent_taxons.count > 1 || tree_node.node_depth == 0) %>
-        <br/><strong>Parents:</strong>
-        <div class="taxon-level-parent-links">
-          <% parent_taxons.each do |parent_taxon| %>
-            <p>
-              ↰ <%= link_to parent_taxon.internal_name, taxon_path(parent_taxon.content_id) %>
-            </p>
-          <% end %>
+  <div class="taxon-parents">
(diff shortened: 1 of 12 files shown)


app/views/taxons/_taxonomy_tree.html.erb
```diff
@@ -1,22 +1,52 @@
...
+              <%= link_to parent_ancestor.title, taxon_path(parent_ancestor.content_id) %>
+            </span>
+          </div>
+        <% end %>
+      </div>
+    <% end %>
+  </div>
 
-      <% parent_taxons = tree_node.parent_taxons %>
+  <% parents_class = taxonomy_tree.immediate_parents.present? ? "taxon-focus--has-parents" : "" %>
```
mobaig: It's a reasonably common CSS convention to place modifiers after a `--`, see SUIT CSS and BEM, for example. No mention of anything like this in the GOV.UK CSS styleguide so I figured it's legit.

app/views/taxons/confirm_delete.html.erb
```diff
@@ -1,4 +1,4 @@
-<h1>You are about to delete "<%= taxon.internal_name %>"</h1>
+<h1>You are about to delete "<%= taxon.title %>"</h1>
```
mobaig: Hasty code changes in the related controller. Just seen we have access to a proper taxon object with internal name on it. Updated.

PR #223, opened 2016-10-26
Better handling of circular dependencies

mobaig:
Make the taxonomy visualisation more robust to errors caused by circular relationships between taxons. Specifically, by:

1. Adding a history of expanded nodes when building out the tree of child taxons in ExpandedTaxonomy.
2. Some basic validation and UI changes to prevent a taxon accidentally being made a parent of itself.

Further details in each commit. 

https://trello.com/c/FI9NYgiz/264-correctly-handle-circular-dependencies-when-linking-taxons

### DIFF
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -21,7 +21,21 @@ class TaxonsController < ApplicationController
   end
 
   def create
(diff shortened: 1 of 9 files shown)


carvil:
Good one @mobaig !

You might want to rebase e2a6b52

Other than that, good to 🚢

PR #225, opened 2016-10-28
Rework styling on taxonomy visualisation

mobaig:
- Make the ends of branches fold inwards when displaying taxons at the
  very left or very right of the visualisation. This removes the
  impression that the taxonomy appears 'cut-off' and incomplete either
  side.
- Remove the unnecessarily long border above or below the taxon in focus
  when there is only one parent or child node.

screen_shot_2016-10-28_at_11_19_51

screen shot 2016-10-28 at 11 25 44


mobaig:
reopening this on a different branch.

PR #227, opened 2016-10-28
[DO NOT MERGE] Use internal name in taxonomy visualisation

mobaig:
For consistency with the taxons index page, use the internal name field
when displaying a taxonomy visualisation.

Needs https://github.com/alphagov/publishing-api/commit/93917e6993fa64bdf7a1479d0a9191ea908ad0bd deployed first.


tijmenb:
This was done instead: https://github.com/alphagov/content-tagger/pull/231

PR #228, opened 2016-10-28
Limit the number of fields from get_content_items

carvil:
We have seen quite a few timeouts in integration when searching for
content items. That's partly due to the fact that we were asking for all
fields.

This change aims at speeding up the Bulk Tag queries so that we avoid
getting timeouts.

### DIFF
--- a/app/services/bulk_tagging/search.rb
+++ b/app/services/bulk_tagging/search.rb
@@ -26,7 +26,8 @@ module BulkTagging
       Services.publishing_api.get_content_items(
         document_type: document_type,
         page: page,
-        q: query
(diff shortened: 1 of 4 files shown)


app/services/bulk_tagging/search.rb
```diff
@@ -26,7 +26,8 @@ def gds_response
       Services.publishing_api.get_content_items(
         document_type: document_type,
         page: page,
-        q: query
+        q: query,
+        fields: [:content_id, :document_type, :title, :base_path]
```
dougdroper: 👍

PR #230, opened 2016-10-31
Rename all instances of "Tag Search" to "Bulk Tag"

klssmith:
"Bulk Tag" is used for the UI, so this commit ensures that Bulk Tag is
always used in the code too.

Part of [this Trello card.](https://trello.com/c/Erx6NIwe/151-rename-code-in-content-tagger)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -28,8 +28,8 @@ Rails.application.routes.draw do
     get  'progress'
   end
 
-  resource :tag_search, only: [:new] do
-    get 'results' => 'tag_searches#results', as: "results_of"
+  resource :bulk_tag, only: [:new] do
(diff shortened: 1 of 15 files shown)


config/routes.rb
```diff
@@ -28,8 +28,8 @@
     get  'progress'
   end
 
-  resource :tag_search, only: [:new] do
-    get 'results' => 'tag_searches#results', as: "results_of"
+  resource :bulk_tag, only: [:new] do
+    get 'results' => 'bulk_tags#results', as: "results_of"
```
mobaig: The helper method 'results_of_bulk_tag_path' is a little misleading. Reckon it'd be better to change the `as: ` value to `search_results_for`.

klssmith: Thanks @mobaig - have renamed the path now.

PR #231, opened 2016-10-31
Use internal name in taxonomy visualisation

mobaig:
Now that the publishing API provides the details hash in taxon expanded links, we have access to the internal name.

This PR includes the switch to using the details hash and some refactoring that aims to keep naming consistent within the app.

https://trello.com/c/Y7HZ6DQ3/263-get-the-details-hash-into-taxon-expanded-links


app/controllers/taxonomies_controller.rb
```diff
@@ -7,7 +7,7 @@ def show
       format.csv do
         send_data(
           CsvTreePresenter.new(@taxonomy.child_expansion).present,
-          filename: @taxonomy.root_node.title + ".csv"
+          filename: @taxonomy.root_node.content_item.title + ".csv"
```
tijmenb: Do we still need this when we're delegating `content_id` from `TreenNode`?

PR #235, opened 2016-11-29
Make basepath visible

carolinegreen:
Make base path visible read only field on editing page.

Before:

screen shot 2016-11-29 at 17 07 05

After:

screen shot 2016-11-29 at 17 07 27


mobaig:
lgtm 👍

tijmenb:
Wooo, welcome to Ruby-land @carolinegreen! 👍

PR #236, opened 2016-12-02
Fix 'addEventListener' used before it was defined

Davidslv:
In JavaScript, prior to ES6, variable and function declarations are hoisted to the top of a scope,
so it's possible to use identifiers before their formal declarations in code.

This can be confusing and some believe it is best to always declare variables and functions before using them.

Source: http://eslint.org/docs/rules/no-use-before-define


tijmenb:
Seems legit!

PR #237, opened 2016-12-06
Render 404 page for invalid taxons

rubenarakelyan:
Currently, attempting to access an invalid taxon displays an application error. This commit adds exception handling to catch such requests and display a 404 error page.

Trello: https://trello.com/c/3oEY4cUZ/354-fix-crash-in-content-tagger

### DIFF
--- a/spec/controllers/taxons_controller_spec.rb
+++ b/spec/controllers/taxons_controller_spec.rb
@@ -15,6 +15,17 @@ RSpec.describe TaxonsController, type: :controller do
     end
   end
 
+  describe "#show" do
+    it "renders 404 for unknown taxons" do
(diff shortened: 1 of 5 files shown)


spec/controllers/taxons_controller_spec.rb
```diff
@@ -15,6 +15,17 @@
     end
   end
 
+  describe "#show" do
+    it "renders 404 for unknown taxons" do
+      stub_request(:get, "https://publishing-api.test.gov.uk/v2/content/does-not-exist")
```
tijmenb: Why doesn't this use `publishing_api_does_not_have_item` like below?

rubenarakelyan: Ah, I copied the tests from the the tagging set and they were inconsistent 😔

PR #244, opened 2016-12-14
Add Jenkinsfile

suzannehamilton:
This adds the Jenkins configuration to allow pipeline builds on the Jenkins 2 CI server.

Currently blocked while we add qmake to the CI servers (alphagov/govuk-puppet#5273).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,84 @@
+#!/usr/bin/env groovy
+
+REPOSITORY = 'content-tagger'
+
+node {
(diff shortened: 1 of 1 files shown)


suzannehamilton:
The Jenkinsfile PRs I'm raising are all quite similar (but not identical - for example, this one does Sass linting). I've discussed it with @tijmenb, and the plan is to migrate them all separately, and then find the best way to commonise them at the end.

carvil:
👍

surminus:
🎉

PR #245, opened 2016-12-14
[WIP]  Proof of concept of a way to manage ordered tags in the tagging form

MatMoore:
We're moving tagging of ordered related items from panoption to tagger. These are the links to other pages on GOV.UK that sometimes get shown in a sidebar.
https://trello.com/c/D6TywAem/294-replace-panopticon-s-related-artefact-functionality-with-content-tagger

If possible we want to show this in the same form as all the other tags, so that the user can easily see all the tags for a content item and understand how the content will be linked to/from on GOV.UK.

This commit enhances the fixed length list of inputs added in the previous commit when javascript is enabled.

To get this to work, I've implemented a new widget with:

- Drag and drop to reorder
- A way to look up a related item by base path and add it to the links if it exists
- A button to remove each tag
- Similar look and feel to the other tagging widgets
- A way to preview what the link will look like

**This is nowhere near done.** I'd like this to be reviewed for the overall approach taken. Please do not review the code itself yet. Can also go through this at my desk if it's easier.

# What it looks like

## With javascript
screen shot 2016-12-14 at 17 43 16

(please ignore the weird icon in the input, that's lastpass)

## Without javascript
screen shot 2016-12-14 at 17 48 16

# What panopticon looked like
screen shot 2016-12-14 at 17 47 34

#  How the widget is intended to work
- The ordered related links are rendered in a `ul` (on reflection, this should probably be a `ol`)
- Each list item contains two divs:
    1. A text input that holds the `base_path` values passed back to the backend
    2. A div representing the actual tag, like what we show in the `select2` widgets
- If javascript is disabled, the user sees the text inputs
- If javascript is enabled the user sees the tags
- Drag and drop reorders the whole list item in the DOM
- All the text inputs have the same name, and rails reads them into an array

## Why I can't use the existing widget content tagger uses for other tag types
- It's built on select2, and assumes you have a fixed list of options. For related items you should be able to tag to any content item.
- It doesn't have an ordering to it, and we need to be able to set the order links will appear in the sidebar.

 ## Why I didn't reuse the related artefacts widget from panopticon
- It's still not a free choice of any GOV.UK content. It autocompletes from a huge list, and this is quite unresponsive in panopticon.
- We decided for this ticket that autocomplete was not needed, which is why there is a separate field for looking up new items.
- It's not hugely obvious that the tags are orderable - we have to explain it in text
- I wasn't aware of it when we started, and the panopticon one does a different thing (progressive enhancement of a comma separated value field instead of multiple text boxes). I'm not sure at this point how to reuse any part of the panopticon solution without keeping the whole thing.

# What's left
## Not done yet
- When there is an error in the form, we need to refetch the titles of all related links since we only submit the base paths. Note that in this case we may have a mix of valid and invalid base paths.
- Problems looking up a tag are currently ignored - we should show errors if there are errors.
- On the backend, we should be restricting this tagging to formats that show related links in the rendering app.
- Non GOV.UK URLs can be normalised to valid GOV.UK base paths - this is a bug.

## Needs improving:
- The CSS needs to be extracted into classes
- Some things that should be full width, aren't, because I wasn't sure how to do the CSS
- There are multiple places responsible for rendering the same HTML; I think we should create a shared template for a related link, that the javascript can use for dynamically add
- All references to artefacts/panopticon stuff should be removed
- There is duplication of base path lookup functionality between `/taggings/lookup` and `/taggings/<content_id>`
- The templates should use `govuk_admin_templates` stuff for showing the flash in a standard way and including analytics
- On the backend, there are a lot of different Structs and value classes used for similar things, which could be simplified
- It looks weird when there are no related links tagged to a content item.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,7 @@ gem 'airbrake', '~> 4.3.1'
 gem 'sass-rails', '~> 5.0'
 gem 'uglifier', '>= 1.3.0'
 gem 'select2-rails', '~> 3.5.9'
+gem 'jquery-ui-rails', '6.0.1'
 
 gem 'govuk_sidekiq', '~> 0.0.4'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -134,6 +134,8 @@ GEM
       rails-dom-testing (>= 1, < 3)
       railties (>= 4.2.0)
       thor (>= 0.14, < 2.0)
+    jquery-ui-rails (6.0.1)
+      railties (>= 3.2.16)
     json (1.8.3)
     json-schema (2.5.2)
       addressable (~> 2.3.8)
@@ -362,6 +364,7 @@ DEPENDENCIES
   govuk_admin_template (~> 4.4)
   govuk_sidekiq (~> 0.0.4)
   headless
+  jquery-ui-rails (= 6.0.1)
   kaminari (~> 0.17)
   logstasher (= 0.6.2)
   pg
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,6 +1,7 @@
 //= require_tree .
 //= require select2
 //= require_self
+//= require jquery-ui
 
 $(document).ready(function() {
   $(".select2").select2({ allowClear: true });
--- a/app/assets/javascripts/views/taggings_show.js
+++ b/app/assets/javascripts/views/taggings_show.js
@@ -1,0 +1,5 @@
+$(document).ready(function() {
+  "use strict";
+
+  $(".sortable-inputs").sortable();
+});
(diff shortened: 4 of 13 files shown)


carolinegreen:
It looks good to me too so far.  Really nice you are mentioning the semantics of markup ul, vs ol.  

I wonder about whether we should have kept the mention in the bracket of draggable links from Pantopicon as I think it’s supposed to be a good idea not just to rely on one way of representing things but @alextea would be better placed to answer that.

I think you’ve covered these in your things to do, but separating the inline styles from the html, and turning things into templates sounds like a good idea.  The css classes, could do with being renamed as they don’t have the best names describing their functionality and are still carrying things over from the widget you’ve based this on.  You’ll probably be able to remove some classes as well.  I think you probably already know and are planning to do this, but I am mentioning it in case.

tijmenb:
Continued in https://github.com/alphagov/content-tagger/pull/261.

PR #246, opened 2016-12-16
Add ordered related links to tagging form (version without JS)

MatMoore:
@tijmenb and I discussed deploying this on its own before https://github.com/alphagov/content-tagger/pull/245 is ready. See that PR for screenshots of what it will look like when done.

The downside is that it will be harder to reorder related links, and the inputs will take up a lot of vertical space.

If we decide against deploying this I think it would still be useful to code review this separately.

Screenshot:
screen shot 2016-12-16 at 11 51 05

5 blank fields are included each time you edit the form. There isn't a way to add more without javascript so if you want to add more links you need to save the form to do the next 5.

### DIFF
--- a/app/controllers/taggings_controller.rb
+++ b/app/controllers/taggings_controller.rb
@@ -28,14 +28,23 @@ class TaggingsController < ApplicationController
     content_item = ContentItem.find!(params[:content_id])
     tag_types = content_item.allowed_tag_types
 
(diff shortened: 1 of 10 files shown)


MatMoore:
We've decided to deploy this behind a feature toggle, so the form won't be shown in production yet.

PR #249, opened 2016-12-16
Don't reference specific tag types

MatMoore:
The activemodel fields are dynamically generated from an array defined
in ContentItemExpandedLinks.

Currently we're varying that list as a way of feature-toggling the new ordered
related items fields, so we shouldn't rely on any particular one being defined.

### DIFF
--- a/app/forms/tagging_update_form.rb
+++ b/app/forms/tagging_update_form.rb
@@ -2,27 +2,38 @@ class TaggingUpdateForm
   include ActiveModel::Model
   attr_accessor :content_id, :previous_version
 
-  attr_accessor(*ContentItemExpandedLinks::TAG_TYPES)
+  TAG_TYPES = ContentItemExpandedLinks::TAG_TYPES
+  attr_accessor(*TAG_TYPES)
 
   # The number of extra empty form fields to add to a link section when the link
   # section shows an individual form input for each value. This allows users to
   # append new links to the end of the existing list.
   EXTRA_TEXT_FIELD_COUNT = 5
 
-  validate :related_item_paths_should_be_valid
+  if TAG_TYPES.include?(:ordered_related_items)
+    validate :related_item_paths_should_be_valid
+  end
 
   # Return a new LinkUpdate object with topics, mainstream_browse_pages,
   # organisations and content_item set.
   def self.from_content_item_links(content_item_links)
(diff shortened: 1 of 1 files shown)


app/forms/tagging_update_form.rb
```diff
@@ -2,27 +2,38 @@ class TaggingUpdateForm
...
-  validate :related_item_paths_should_be_valid
+  if TAG_TYPES.include?(:ordered_related_items)
+    validate :related_item_paths_should_be_valid
+  end
 
   # Return a new LinkUpdate object with topics, mainstream_browse_pages,
   # organisations and content_item set.
   def self.from_content_item_links(content_item_links)
+    tags = TAG_TYPES.each_with_object({}) do |tag_type, current_tags|
+      if tag_type == :ordered_related_items
```
carvil: Since this branch will only happen once, what do you think of:

```
def self.from_content_item_links(content_item_links)
 tags = TAG_TYPES.each_with_object({}) do |tag_type, current_tags|
     current_tags[tag_type] = extract_content_ids(
       content_item_links.send(tag_type)
     )
 end

 tags[:ordered_related_items] =
   pad_with_empty_items(tags[:ordered_related_items])

 new(
   content_id: content_item_links.content_id,
   previous_version: content_item_links.previous_version,
   **tags
 )
end
```

MatMoore: Wouldn't that make a call like this when related items is disabled?

```ruby
new(
   ...
  ordered_related_items: nil
)
```

carvil: Yes, that's correct. We would need to add the same conditional as above to the end of that line:

```
if TAG_TYPES.include?(:ordered_related_items)
  tags[:ordered_related_items] =
     pad_with_empty_items(tags[:ordered_related_items])
end
```

I think keeping it separate makes it easier to read, but also happy to leave it as is.

MatMoore:
I'm not sure this can be unit tested, the way it's currently written. It would be more of an integration test, and I'm not sure it's worth the effort vs just rolling back the changes and rebasing the branch in the new year. The only reason there is a conditional at all was as a quick hack to keep testing the "future" behaviour, because a whole bunch of tests change in a minor way when you change the available tags.

I haven't given this much thought, the only motivation was to avoid keeping the branch open and falling behind master.

PR #250, opened 2016-12-19
Fix menubar selection highlighting

klssmith:
- Previously, when viewing some pages there was no menu item that
appeared to be active.
- Every page now has one menu item highlighted.

Trello: https://trello.com/c/gifTCGLV/374-content-tagger-lots-of-microcopy-changes-and-some-button-layout-tweaks


tijmenb:
👍 nice!

PR #258, opened 2016-12-22
Remove feature flag for related links tagging

tijmenb:
We're finishing this feature now. Because of the deploy freeze this is safe to merge now.

https://trello.com/c/D6TywAem


rubenarakelyan:
🚢

PR #259, opened 2016-12-22
Add more testing for related item tagging

tijmenb:
This extracts the test for related items from the main spec. It improves readability (at the cost of duplication). Also adds a specific test for the happy path.

https://trello.com/c/D6TywAem


rubenarakelyan:
👍

PR #260, opened 2016-12-22
Refactor tagging form handling

tijmenb:
This refactors the way we create the tagging form and handle the user input.

It builds upon https://github.com/alphagov/content-tagger/pull/241 and ideas from https://github.com/alphagov/content-tagger/pull/238.

It attempts to improve the readability of the controller and `TaggingUpdateForm`.

The basic story is:

- The `show` action in the controller instantiates a `TaggingUpdateForm` object. This is an ActiveModel-compliant object and it's only responsibility is to allow us to build the form.
- The `update` action uses the `TaggingUpdatePublisher` class, which has the task of parsing the user input and sending it to the publishing-api.

This also makes sure that we use just one instance variable in the controllers, making us a little more Sandi-compliant (https://robots.thoughtbot.com/sandi-metz-rules-for-developers).


app/controllers/taggings_controller.rb
```diff
@@ -15,32 +15,22 @@ def find_by_slug
...
     content_item = ContentItem.find!(params[:content_id])
-    tag_types = content_item.allowed_tag_types
-
-    if tagging_update_form.valid?
-      Services.publishing_api.patch_links(
-        tagging_update_form.content_id,
-        links: tagging_update_form.links_payload(tag_types),
-        previous_version: tagging_update_form.previous_version.to_i,
-      )
+    tagging_update = Tagging::TaggingUpdatePublisher.new(content_item, params[:tagging_tagging_update_form])
```
rubenarakelyan: `tagging_update_form`? (plus in 3 other places)

tijmenb: I've renamed it to `publisher` to differentiate it from the form object.

rubenarakelyan: Ah sorry, I meant the `tagging_tagging_update_form` :)

tijmenb: That's actually correct - now that that class lives inside the `Tagging` module, Rails generates this param name for it.

rubenarakelyan: Ah cool

PR #261, opened 2016-12-22
Add javascript UI for related links

tijmenb:
This is the continuation of https://github.com/alphagov/content-tagger/pull/245 rebased on top of the refactors in https://github.com/alphagov/content-tagger/pull/260.

Trello: https://trello.com/c/D6TywAem

### DIFF
--- a/app/assets/javascripts/views/related-content-tagger.js
+++ b/app/assets/javascripts/views/related-content-tagger.js
@@ -1,0 +1,73 @@
+(function (Modules) {
+  "use strict";
+
+  Modules.RelatedContentTagger = function () {
+    this.start = function () {
+
+      $(".ordered-related-items").on("click", ".select2-search-choice-close", removeItem);
+      $("#related_item_new_base_path").on("keypress", lookUpBasePathOnEnterPress);
+      $("#lookup_base_path_button").on("click", lookUpBasePath);
+
+      $(".related-content-item-entry .title").show();
+      $(".related-content-item-entry .value").hide();
+      $(".add-sortable-input").show();
+      $(".sortable-inputs").sortable();
+
+
+      function removeItem() {
+        $(this).parents("li").remove();
+        return false;
+      }
+
+      function lookUpBasePathOnEnterPress(event) {
+        if (enterKeyPressed(event)) {
+          lookUpBasePath();
+
+          // Prevent form submission
+          return false;
+        }
+      }
+
+      function enterKeyPressed(event) {
+        return event.keyCode === 13;
+      }
+
+      function lookUpBasePath() {
+        var $basePathInput = $("#related_item_new_base_path");
+        var basePath = $basePathInput.val();
+
+        $.getJSON({
+          url: "/taggings/lookup-urls?base_path=" + encodeURIComponent(basePath),
+          success: onTagLookupSuccess,
+          error: onTagLookupError
+        });
+
+        function onTagLookupSuccess(lookup) {
+          var $templateTag = $(".related-item-template > li").first();
+
+          var $newTag = $templateTag
+            .clone()
+            .appendTo(".ordered-related-items ul");
+
+          $newTag
+            .find("input")
+            .attr("value", lookup.base_path);
+
+          $newTag
+            .find(".js-artefact-name")
+            .text(lookup.title + " (" + lookup.base_path + ")")
+            .attr("href", "https://www.gov.uk" + lookup.base_path);
+
+          $basePathInput.val("");
+          $basePathInput.removeClass("has-error");
+          $(".related-item-error-message").hide();
+        }
+
+        function onTagLookupError(error) {
+          $basePathInput.addClass("has-error");
+          $(".related-item-error-message").show();
+        }
+      }
+    };
+  };
+})(window.GOVUKAdmin.Modules);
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/views/taggings_show.js
```diff
@@ -0,0 +1,51 @@
...
+  // a new related link.
+  $("#related_item_new_base_path").on("keypress", function() {
+    if (event.keyCode === 13) {
+      $("#lookup_base_path_button").click();
+      return false;
+    }
+  });
+
+  $("#lookup_base_path_button").on("click", function() {
+    var basePathInput = $("#related_item_new_base_path");
```
Davidslv: Could you change the name of the variables that are jQuery objects to have `$` as prefix please?
Seems to be a common practice https://github.com/alphagov/styleguides/blob/master/js.md#jquery

app/views/taggings/_related_item_tagging_entry.html.erb
```diff
@@ -0,0 +1,23 @@
+<li class="related-content-item-entry">
+  <% unless base_path.blank? || @tagging_update.related_item_errors[base_path] %>
+    <div class="title select2-search-choice ui-sortable-handle ">
+      <a href="https://www.gov.uk<%= base_path %>" class="js-artefact-name">
```
Davidslv: is there a way of not having the url half hardcoded?

MatMoore:
This has still has some bugs so I'm going to focus on that before the cleanup stuff. Added a failing feature test to illustrate the problem.

app/controllers/taggings_controller.rb
```diff
@@ -14,6 +14,22 @@ def find_by_slug
...
+      render json: {
+        base_path: content_item.base_path,
+        content_id: content_item.content_id,
+        title: content_item.title
+      }
+    else
+      render json: { errors: content_lookup.errors }, status: 404
+    end
+  end
+
```
MatMoore: I'm still not sure about the relationship between these two objects in

```ruby
    publisher = Tagging::TaggingUpdatePublisher.new(content_item, params[:tagging_tagging_update_form])

    if publisher.save_to_publishing_api
      redirect_to :back, success: "Tags have been updated!"
    else
      @tagging_update = Tagging::TaggingUpdateForm.from_content_item(content_item)
      @tagging_update.related_item_errors = publisher.errors
      @tagging_update.update_attributes_from_form(params[:tagging_tagging_update_form])
```

- It seems like it assumes that the publisher's errors are all to do with related links, which is technically true but could easily change
- All the data extracted from `#from_content_item` is overriden by `#update_attributes_from_form` except for `links`, which is used to expand base paths later on. This feels a bit weird to me as I'd expect to pass the request params to the form at initialisation time.
- `#title_for_related_link` doesn't work when you submit an invalid form with valid (but new) related items in it

Some naming things
- There's redundancy in `Tagging::TaggingUpdateForm` and `Tagging::TaggingUpdatePublisher`
- TaggingUpdateForm is still named form but it's separate from `app/forms`

MatMoore:
Having to hand this over again as I'm on 2nd line tomorrow.

I think the backend part is working OK now (there was a bug with failed form validations that I've fixed). Hoping that we can just focus on cleaning up the front end stuff now (there's a few comments from Caroline in https://github.com/alphagov/content-tagger/pull/245 as well as this one).

app/views/taggings/_related_item_tagging_entry.html.erb
MatMoore: Should be, but haven't looked into this yet. Could always add a helper if one doesn't exist.

app/assets/javascripts/views/related-content-tagger.js
```diff
@@ -0,0 +1,76 @@
+(function(Modules) {
+  "use strict";
+
+  Modules.RelatedContentTagger = function() {
+    this.start = function() {
+      $(document).ready(function() {
+        "use strict";
```
tijmenb: Are the `$.ready` and `use strict` redundant here?

tijmenb:
I've squashed all commits by me, @MatMoore & @alecgibson into one.

app/services/tagging/tagging_update_form.rb
```diff
@@ -40,6 +41,24 @@ def related_item_errors
...
+
+      if link.nil?
+        # links is populated from the existing links. When we submit a form and
+        # it fails, we may have additional related links we want to know the
+        # titles of.
+        # In practice this shouldn't actually be visible most of the time,
+        # because with javascript enabled, the user can't enter invalid data,
+        # and without it we don't display the titles. It is still possible to
+        # see errors due to tagging conflicts though.
+        # TODO: attempt to fetch missing links on demand.
```
MatMoore: Are we ok with the way this is done? Maybe it would be nicer to explicitly return nil here and handle it in the view?

alecgibson: I don't know if this is relevant, but I'm about to start adding the feature to manually edit taxon URLs (https://trello.com/c/5cBp6Ryt/303-allow-taxon-urls-to-be-edited). 

This will require detection of URL collision (and appropriate error handling), so I could look into this on that card? (Although it also means there could be impending merge conflicts?)

tijmenb: I'm okay with doing it as-is. Not sure if @alecgibson's card will affect this particular code (but keep it in mind).

PR #262, opened 2017-01-03
Reorganise everything into namespaces

tijmenb:
Big commit, sorry.

This organises most of the code into one of three namespaces: Tagging, BulkTagging and Taxonomy.

The goal of this is make the application easier to understand to newcomers and to avoid it becoming a big ball of code. The three features of this application are pretty well-defined, and should not have to talk to each other.

Some notes:

- I considered inverting the pattern/namespace (so that it reads `app/taxonomy/models`, `app/bulk_tagging/services`). That seems a little outside of the mainstream, but might be worth talking about later.
- There are a couple of classes that are used by multiple modules. I've kept these at the root level. The rules of this separation are that no module can talk to each other, but that each module can talk to the top level. The top level should also not talk to the modules, but that requires refactoring outside the scope of this PR.

## How to review

Probably best to clone the branch and review it in your editor.

Part of: https://trello.com/c/Erx6NIwe/151-content-tagger-rename-code-in-content-tagger


app/workers/initial_tagging_import.rb
```diff
@@ -1,11 +1,12 @@
+# TODO: put this safely into BulkTagging module
```
MatMoore: What's stopping this?

app/models/remote_taxons.rb
```diff
@@ -1,6 +1,7 @@
+# TODO: Split this up and move it into the BulkTagging & Taxonomy namespaces.
```
MatMoore: Which part of this would live in `BulkTagging`? This seems like it could all go in `Taxonomy` to me, but I'm not familiar with how it's modelled.

It seems like `#parents_for_taxon` is something you would want to call on `Taxon`, and taxon searching could also live in the taxonomy namespace.

tijmenb: Not sure yet! That's why I wanted to do this in a separate PR.

app/workers/initial_tagging_import.rb
tijmenb: We need to keep the old class, in cases there are in-flight jobs during deploy. That seemed too much for this PR.

tijmenb:
@MatMoore Thanks! I'm planning another PR to address some of the naming things.

PR #263, opened 2017-01-04
Reload page when bulk tag operation is complete

Davidslv:
User Research showed IAs were lacked confidence about when a bulk
tagging operation was complete.

This commit reloads the page to provide a clear visual feedback that the
operation has been complete.

screen shot 2017-01-04 at 15 01 05

Trello:
https://trello.com/c/PCMxxmq9/371-content-tagger-provide-clear-visual-feedback-when-bulk-tagging-operations-are-complete

### DIFF
--- a/app/assets/javascripts/js.cookies.js
+++ b/app/assets/javascripts/js.cookies.js
@@ -1,0 +1,156 @@
+/*!
+ * JavaScript Cookie v2.1.3
+ * https://github.com/js-cookie/js-cookie
+ *
+ * Copyright 2006, 2015 Klaus Hartl & Fagner Brack
+ * Released under the MIT license
+ */
+;(function (factory) {
+	var registeredInModuleLoader = false;
+	if (typeof define === 'function' && define.amd) {
+		define(factory);
+		registeredInModuleLoader = true;
+	}
+	if (typeof exports === 'object') {
+		module.exports = factory();
+		registeredInModuleLoader = true;
+	}
+	if (!registeredInModuleLoader) {
+		var OldCookies = window.Cookies;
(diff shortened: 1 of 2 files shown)


MatMoore:
What's the reloaded cookie for?

Davidslv:
@MatMoore the cookie is to avoid a loop reloading the page once it's completed, so we set a cookie to as a flag so the script doesn't reload the page when the page has been reloaded and the process complete.

MatMoore:
Ah ok, I understand the reasoning, but it seems a bit wrong to do it with cookies - in theory you could go to the URL from a different browser and it would still have the same problem.

Could you just avoid the reload if the thing you're refreshing for is not on the page?

Or another way to do it would be to redirect to a separate "completed" URL when done (you could still reuse the templates).

Davidslv:
We tried to use `window.location.replace` and `window.location.assign` with a `#complete` in the end of the url but that didn't reloaded the page and required an extra line `window.location.reload` which caused the specs to fail, so we have decided to keep the implementation with cookies, even though that means using an external library (which is just use for ease of setting and getting a cookie without writing two extra functions as it can be seen in this prototype: https://github.com/alphagov/content-tagger/commit/490fb122001fb8301137c2a98ac27ab6b19b28eb

app/assets/javascripts/js.cookies.js
```diff
@@ -0,0 +1,156 @@
+/*!
```
tijmenb: Can you put this file in `vendor/assets`?

http://guides.rubyonrails.org/asset_pipeline.html#asset-organization

PR #266, opened 2017-01-11
Use separate link to link to site

tijmenb:
When we link the base path, it's not entirely clear where we are sending the user. This removes any ambiguity.

https://trello.com/c/CW3lMGRf

## Before

screen shot 2017-01-11 at 17 42 26

## After

screen shot 2017-01-11 at 17 41 11


rubenarakelyan:
🎉

PR #268, opened 2017-01-12
Split base path into prefix and slug

alecgibson:
Add the ability to edit the base path when creating or updating a taxon.

The base path is edited as two components:
* Prefix: the base path of a top-level taxon, selected from a hard-coded static list (to be replaced by an API call in future)
* Slug: the customisable part of the path, which defaults to the document title (if present)

taxon_base_path_editing

Note that we currently have no special handling of update errors due to a base path collision. This is waiting on the Publishing API to improve error reporting so that we can determine error sources programmatically.

Trello: https://trello.com/c/5cBp6Ryt/303-allow-taxon-urls-to-be-edited

This Pull Request also includes the rake task for updating existing Education URLs. Note that anything _not_ Education-related (ie Driving?) will have to be removed before running the task.

Trello: https://trello.com/c/XIDoBBqJ/312-migrate-existing-taxon-urls-to-the-new-url-structure


app/models/theme.rb
```diff
@@ -0,0 +1,7 @@
+class Theme
+  EDUCATION_THEME_BASE_PATH = '/education-training-and-skills'.freeze
+
+  def self.taxon_path_prefixes
+    [EDUCATION_THEME_BASE_PATH]
```
tijmenb: Can we add `/alpha-taxonomy` here too? Then the existing taxons won't be invalid.

app/models/taxon.rb
```diff
@@ -24,14 +26,35 @@ def content_id
...
 
+  def path_prefix
+    @path_prefix ||= ''
+  end
+
+  def path_slug
+    @path_slug ||= (title.blank? ? '' : '/' + title.parameterize)
+  end
+
+  def base_path=(base_path)
```
tijmenb: Would `@path_prefix, @path_slug = base_path.split('/')` work here?

app/models/taxon.rb
```diff
@@ -24,14 +26,35 @@ def content_id
     @content_id ||= SecureRandom.uuid
   end
 
+  def path_prefix
+    @path_prefix ||= ''
```
tijmenb: Could we do something to get rid of these setters?

app/controllers/taxons_controller.rb
```diff
@@ -93,6 +103,10 @@ def destroy_flash_message(response_code)
     end
   end
 
+  def path_prefixes_for_select
+    Theme.taxon_path_prefixes
```
tijmenb: Might be easier to put `Theme.taxon_path_prefixes` into the view directly, so avoid having to pass them around in the views.

app/models/taxon.rb
```diff
@@ -13,7 +15,7 @@ class Taxon
 
   include ActiveModel::Model
 
-  validates_presence_of :title, :internal_name
+  validates_presence_of :title, :internal_name, :path_prefix, :path_slug
   validates_with CircularDependencyValidator
```
tijmenb: We'll probably have to validate that the `path_prefix` is one of our allowed prefixes. The `path_slug` also needs some validation (a-z 0-9 and dashes?).

app/models/taxon.rb
alecgibson:
```
'/foo/bar/baz'.split('/')
 => ["", "foo", "bar", "baz"] 
```

app/models/theme.rb
alecgibson: I've just written a rake task that will rename them all. Is that sufficient?

app/models/taxon.rb
alecgibson: Like what?

app/models/taxon.rb
tijmenb: `_, @path_prefix, @path_slug = base_path.split('/')`

app/models/theme.rb
tijmenb: Yeah, I guess so, but it would have to go out at the same time as this PR.

app/controllers/taxons_controller.rb
alecgibson: It used to be, but `taxonomy_editing_spec.rb` now contains tests that rely on a valid prefix, so changing them may cause confusing failures.

app/models/taxon.rb
alecgibson: This loses the delimiter, which would require putting the getters back in to put the '/' back, or else fiddling around a bit more in this setter. Not convinced this is much cleaner?

app/controllers/taxons_controller.rb
alecgibson: As discussed, will leave for now and refactor later.

app/models/taxon.rb
tijmenb: Yup, agreed, it's fine like this.

lib/tasks/taxonomy/fixup_taxon_urls.rake
```diff
@@ -0,0 +1,37 @@
+namespace :taxonomy do
+  desc "fixup taxon URLs"
+  task fixup_taxon_urls: :environment do
+    total = RemoteTaxons.new.search.search_response["total"]
+    taxons = RemoteTaxons.new.search(per_page: total).taxons
```
tijmenb: Would it be more efficient to use

```
      Services.publishing_api.put_content(content_id, payload)
      Services.publishing_api.publish(content_id, "minor")
```

instead of `Taxonomy::PublishTaxon`?

carvil:
Looks good. Let's try this locally first and make sure the links between taxons are preserved. We can then go ahead and test it on integration.

alecgibson:
@carvil @tijmenb is this good to go now?

app/assets/javascripts/base-path-preview.js
```diff
@@ -0,0 +1,23 @@
...
+  Modules.BasePathPreview = function () {
+    this.start = function() {
+
+      var basePathWrapper = $('.js-base-path');
+      var basePath = $('.js-base-path input');
+      var pathPrefix = $('select.js-path-prefix');
+      var pathSlug = $('input.js-path-slug');
+
+      pathPrefix.change(updateBasePathPreview);
+      pathSlug.change(updateBasePathPreview);
```
carvil: What do you think of using ```pathSlug.on('keydown', updateBasePathPreview);````

That way the view-only field gets updated as we type the slug.

carvil: What do you think of using ```pathSlug.on('keydown', updateBasePathPreview);```

That way the view-only field gets updated as we type the slug.

Gemfile
```diff
@@ -43,7 +43,7 @@ end
 group :test do
   gem 'webmock', '~> 1.22'
   gem 'capybara'
-  gem 'capybara-webkit'
+  gem 'poltergeist'
```
tijmenb: What's the reason for switching to poltergeist?

app/views/taxons/_form_fields.html.erb
```diff
@@ -1,3 +1,22 @@
+<%= f.input :path_prefix,
+  input_html: { class: 'js-path-prefix form-control' },
+  label: I18n.t('views.taxons.path_prefix'),
+  collection: path_prefixes_for_select %>
+
+<%= f.input :path_slug,
+  input_html: { class: 'js-path-slug form-control' },
+  label: I18n.t('views.taxons.path_slug') %>
+
+<%= f.input :base_path,
```
tijmenb: How will this work without JS? Won't it submit an empty `base_path`?

app/views/taxons/edit.html.erb
```diff
@@ -2,12 +2,20 @@
   breadcrumbs: [:taxons, link_to(taxon.internal_name, taxon_path(taxon.content_id)), "Edit"] %>
 
 <%= simple_form_for taxon, url: taxon_path(taxon.content_id), method: :patch do |f| %>
+  <% if flash[:error].present? %>
```
tijmenb: I think if you use `flash[:danger]` a flash message will automatically appear:

https://github.com/alphagov/govuk_admin_template#flash-messages

config/locales/en.yml
```diff
@@ -7,7 +7,7 @@ en:
     tag_migration: 'Bulk tag history'
     tag_importer: 'Bulk tag by upload'
   errors:
-    invalid_taxon: There was a problem with your request and we were notified about it. We will fix this issue as soon as possible
+    invalid_taxon: There was a problem with your request. The chosen slug may already be in use. We have been notified of the issue and will fix it as soon as possible.
```
tijmenb: I think the message might have been incorrect. Are we actually being notified about this? (Would have been via Airbrake).

lib/tasks/taxonomy/fixup_taxon_urls.rake
```diff
@@ -0,0 +1,43 @@
+namespace :taxonomy do
+  desc "fixup taxon URLs"
+  task fixup_taxon_urls: :environment do
+    total = RemoteTaxons.new.search.search_response["total"]
+    taxon_ids = RemoteTaxons.new.search(per_page: total).taxons.map(&:content_id)
```
tijmenb: Is this doing 2 search requests?

carvil:
When I run the task, I get:

```
Finished with 16 errors
The following taxons could not be updated:
07fdd985-f3ec-4f4e-a316-3f4fd491bd64 /education-training-and-skills/tests
8a2141ae-2493-491c-8e5f-2bd23c1a6afe /education-training-and-skills/science
b0fcf462-e6b9-44e1-9145-48f66d53c81f /education-training-and-skills/programmes-of-study
9104dfd2-4da9-4309-98dc-dc54f7d8d2db /education-training-and-skills/maths
dc1d6a10-7f7b-464c-bd1f-3bdb66d096d0 /education-training-and-skills/english
1ddd7a85-6c64-4b1e-9e54-55b50a0ae3f3 /education-training-and-skills/assessments
9adcdd2a-2a92-44ea-87c9-3f20e05bf352 /education-training-and-skills/phonics
9dae94c5-858b-479b-b509-a461d9a2d87d /education-training-and-skills/further-education-skills-and-vocational-training
ea6f0880-91b2-4e35-89de-8f0cb416a316 /education-training-and-skills/funding-and-finance-for-students
b2b74951-cbdb-439a-a097-38ddd5db5b47 /education-training-and-skills/careers-guidance-in-schools
de30706b-620c-42a4-bf56-c345bbe11847 /education-training-and-skills/school-inspections
ea5be62f-bd87-47f9-a986-56238e3b4ff4 /education-training-and-skills/support-for-education-work-in-other-countries
2b1e8cb2-5e59-4b7e-af8d-99d1e833149f /education-training-and-skills/university-and-higher-education
719e8ad1-9307-4e66-b5ce-6b3d182c971f /education-training-and-skills/early-years-pupil-premium
87dad742-39c6-4f07-8114-5bb13b7bd455 /education-training-and-skills/children-s-centres
85395f92-3a84-4dcf-93ea-48f09b3dcaf9 /education-training-and-skills/childminding
```

Which is the issue of similarly named taxons that share the same base path with this change.

I'm ok if we say we have to fix these manually in content tagger for now if people agree. Going forward, content tagger will validate these things anyway, so we won't have to do it again. What do you think?

alecgibson:
@carvil as discussed - I'll run the Rake task as-is, and copy the duplicate paths into a Trello card for someone from Product to pick up

Gemfile
alecgibson: The JavaScript tests were falling over with Webkit, and I couldn't figure out why. Works fine with poltergeist - if you want me to switch it back can you give me a hand debugging webkit?

lib/tasks/taxonomy/fixup_taxon_urls.rake
alecgibson: Yes, but it's a one-off task, so I thought there was no need to over-optimise.

tijmenb: 👍

app/views/taxons/_form_fields.html.erb
alecgibson: This might be something I don't understand about Ruby/Rails, but given that this field is read-only, should it never set anything in the form? The actual model (`taxon.rb`) has a getter for `base_path` that concatenates the prefix and the slug. I can't actually test this locally right now, though because I've got an ES migration going (will test later).

tijmenb: To prevent confusion, if you're never expecting this field to be submitted, it might be better to give it another name and/or not use the form helper like this.

alecgibson: You were 100% right - this was getting submitted.

config/locales/en.yml
alecgibson: I think so? There's an Airbrake notification in `publish_taxon.rb`?

PR #272, opened 2017-01-19
Add Appsignal configuration

tijmenb:
This adds Appsignal to the app. We're evaluating to see if it is a good replacement for Errbit, while also adding performance tracking.

It's only activated if `APPSIGNAL_API_KEY` is set, which is currently only in integration (https://github.com/alphagov/govuk-puppet/pull/5376 https://github.gds/gds/deployment/pull/1215).

cc @carvil 

https://trello.com/c/J3kDeZD7


carvil:
Merging now that puppet ran on the backend servers in integration and the key is set.

PR #275, opened 2017-01-23
Rake task for removing unused taxons

MatMoore:
Unpublish old taxon content items that are not linked to the current
education taxonomy.

Anything that is not reachable from the top level of the education
taxonomy is no longer needed and clutters up content tagger.

None of these have been visible to users so don't bother with redirects.

This should be run with ROOT_TAXON_CONTENT_ID=c58fdadd-7743-46d6-9629-90bb3ccc4ef0 (education, training and skills).

Part of https://trello.com/c/4AJLWbUW/427-content-tagger-clear-out-all-the-test-data

There is a separate task for just printing the orphaned taxons so we
can see what will be removed.

### DIFF
--- a/lib/tasks/taxonomy/remove_orphans.rake
+++ b/lib/tasks/taxonomy/remove_orphans.rake
@@ -1,0 +1,48 @@
+namespace :taxonomy do
+  task show_orphans: :environment do
+    orphaned_taxons.each do |taxon|
+      puts "#{taxon.base_path} (#{taxon.content_id})"
+    end
+  end
+
+  task remove_orphans: :environment do
+    content_ids_to_remove = orphaned_taxons.map(&:content_id)
+
+    content_ids_to_remove.each do |content_id|
+      puts "Removing #{content_id}"
+      Services.publishing_api.unpublish(content_id, type: 'gone')
+    end
+  end
+end
+
+def orphaned_taxons
+  Enumerator.new do |yielder|
+    1.step do |page|
+      results = RemoteTaxons.new.search(page: page).taxons
+
+      break if results.empty?
+
+      results.each do |result|
+        yielder << result unless taxonomy_includes?(result.content_id)
+      end
+    end
+  end
+end
+
+def connected_taxonomy
+  @taxonomy ||= begin
+    root_taxon = ENV['ROOT_TAXON_CONTENT_ID']
+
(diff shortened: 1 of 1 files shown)


app/services/taxonomy/publish_taxon.rb
```diff
@@ -25,6 +25,14 @@ def publish
       raise(InvalidTaxonError, I18n.t('errors.invalid_taxon'))
     end
 
+    def unpublish(type:, alternative_path: nil)
```
carvil: What do you think of using a different class for this? This is called `PublishTaxon`, and having an `unpublish` method could be a bit confusing.

MatMoore: I think a separate class for every action is too much boilerplate. I'd rather rename this one to something that is appropriate for any any publishing api actions on a taxon. But I'm having trouble with words today...

TaxonPublishingApiRequestMakerThing?? 🆘 😞

carvil: I can see how it sounds a bit too much, but we don't have that many service classes at the moment. I think this pattern simplifies unit testing (by not needing `describe` and `context` blocks for example) and makes it transparent in terms of what it's expected from the class. By using a class method like `call` or similar people are not encourage to instantiate the class and play with its data, which I think is good. But it's just a suggestion :) Just brain dumping some of the discussions we had a long time ago when building `content-tagger`

tijmenb: If this is a one-off, perhaps just use `publishing_api.unpublish` directly, like in the controller: 

https://github.com/alphagov/content-tagger/blob/master/app/controllers/taxons_controller.rb#L82

MatMoore: Hmm. I'll think about this tomorrow. Inlining seems most reasonable for now, because if I was to create something more generic I'd expect to use it differently in the future anyway (redirecting or withdrawing instead of just `gone`-ing it).

@carvil brain dumping is definitely welcome. I'm sure I'm missing a lot of context here, because I have almost the complete opposite view about the number of services (I count 14, and I probably couldn't tell you what most of them are for)

MatMoore:
I've inlined the unpublish call into the rake task as this wasn't particularly useful in the PublishTaxon class.

I'll check the list of taxons before unpublishing on production/staging because my dev environment is not up to date enough.

PR #276, opened 2017-01-24
Rename education base path to /education

MatMoore:
Top level themes are the root taxon of a tree. Every other taxon in the tree
has that initial URL as a prefix.

Because of https://github.com/alphagov/content-tagger/pull/268, the taxon
"Primary curriculum, key stage 2" would have a URL like this:
https://www.gov.uk/education-training-and-skills/primary-curriculum-key-stage-2

In https://github.com/alphagov/content-tagger/pull/275, we removed everything that is not under the education theme.

This commit renames the prefix to /education.

Some taxons are not valid at the moment (due to missing descriptions) but the rake task ignores this so we can update the base paths independently.

For now if anything fails to be updated, skip it.

We can remove /alpha-taxonomy, and /education-training-and-skills in future commits, so that the validation still works if we can't migrate all taxons at once.

Trello: https://trello.com/c/CbJ3kzQC/439-rename-the-base-path-of-education-training-and-skills

I need to test this some more - On dev I have some content items that conflict with each other but I'm not sure how they got that way.

### DIFF
--- a/app/models/theme.rb
+++ b/app/models/theme.rb
@@ -1,8 +1,11 @@
 class Theme
-  ALPHA_TAXONOMY = '/alpha-taxonomy'.freeze # TODO: remove once all base paths have been updated
-  EDUCATION_THEME_BASE_PATH = '/education-training-and-skills'.freeze
+  # TODO: remove these once no longer used in base paths
+  ALPHA_TAXONOMY = '/alpha-taxonomy'.freeze
+  OLD_EDUCATION_THEME_BASE_PATH = '/education-training-and-skills'.freeze
+
+  EDUCATION_THEME_BASE_PATH = '/education'.freeze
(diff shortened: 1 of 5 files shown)


MatMoore:
I ran this on integration, and the problem is caused by Key Stage 1 & Key Stage 2:

These taxons can't be renamed in this way, because there are two of each:
- tests
- science
- programmes of study
- maths
- english
- assessments

I suggest we name these like

- /education/key-stage-1-tests
- /education/key-stage-2-tests

I'll add in a workaround for these particular cases.

MatMoore:
I've pushed a new version that has a special case for the ambiguous names. I extracted a class so I can test this handles the three basic cases, but we can always throw this code away after we've run it on prod.

PR #277, opened 2017-01-25
UI tweaks for changing taxon URLs

tijmenb:
A few tweaks to the form to change taxon URLs, based on observing Graeme and Sam using the current interface.

## Before

screen shot 2017-01-25 at 09 53 55

## After

screen shot 2017-01-25 at 09 53 24

https://trello.com/c/MczGaTgT/440-ui-tweaks-for-changing-taxon-urls


alecgibson:
I'd thought about doing something like this - is it alright to pick a fixed width for the prefix? Are we assuming that it won't ever be too long, or that people won't need to increase font size (eg if they're visually impaired)?

tijmenb:
@alecgibson the UI works pretty well with all screen sizes:

screen shot 2017-01-25 at 12 26 15
screen shot 2017-01-25 at 12 26 25

I've spoken to Sam, who thinks this would be good to test with Graeme on integration, so this is ready for review & merge.

PR #282, opened 2017-01-25
Remove unused base path prefixes

MatMoore:
This removes the option of using the prefixes `/alpha-taxonomy` and `/education-training-and-skills` when creating or editing taxons.

Now that we've renamed the slugs of all the education taxons to use `/education` (https://github.com/alphagov/content-tagger/pull/276), and deleted the rest (https://github.com/alphagov/content-tagger/pull/275), we don't need this code any more.


alecgibson:
You can probably also delete `fixup_taxon_urls.rake` and `remove_non_education_taxons.rake`

PR #283, opened 2017-01-26
Update outdated gems

Davidslv:
This PR has been inspired on [Thoughtbot's article](https://robots.thoughtbot.com/keep-your-gems-up-to-date) and was created with the help of [bummr gem](https://github.com/lpender/bummr). 

The intention is to discuss a possible implementation that would do this automatically for us on all the ruby repositories of alphagov to avoid that any ruby live application fall too behind the latest released gem.

### Why?

- I believe that it would make us less vulnerable to security vulnerabilities
- I also believe that by doing this periodically, any broken changes that might occur have a good chance of being easily fixed because it will be a small change instead of a big change that will take too long to be fixed by any of the developers.

### How?

I think that the easiest way would be:

1.  add the `bummr` into the Gemfile and issue a PR against master
2. create a script inside the repository `update-gems.sh` that creates a new branch, runs `bummr` and has capacity of answering the prompt, commit, push and issue a pull request
3. this script would then ran every night on CI 

I might have missed something, let me know what you think by leaving your comment.

Thanks


tijmenb:
The updated gems here look fine, thanks for updating. As for making something automatic, I believe Paul is working on adding Snyk, which does much the same thing (see https://github.com/alphagov/signon/pull/532). I expect that he'll come with a decision on that pretty soon so let's wait for that for a bit.

PR #285, opened 2017-01-30
Add tagging to user needs

cbaines:
This is work that @binaryberry, @issyl0 and I are working on as part of the preparatory work for the migration of Publisher (mainstream publisher) to the Publishing API.

Currently needs are stored in the Need API, and managed in Maslow. We are working on making needs content, and moving them to the Publishing API, which enables associating content and the user needs it meets using the links stored by the Publishing API. These changes bring tagging user needs to content tagger. 

content-tagger-user-needs

### DIFF
--- a/app/models/linkables.rb
+++ b/app/models/linkables.rb
@@ -16,6 +16,10 @@ class Linkables
     @organisations ||= for_document_type('organisation')
   end
 
+  def needs
+    @needs ||= for_document_type('need', include_draft: false)
+  end
+
   def mainstream_browse_pages
     @mainstream_browse_pages ||= for_nested_document_type('mainstream_browse_page')
   end
--- a/app/models/linkables.rb
+++ b/app/models/linkables.rb
@@ -22,8 +22,11 @@ class Linkables
 
 private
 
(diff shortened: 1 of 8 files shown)


app/models/linkables.rb
```diff
@@ -16,14 +16,21 @@ def organisations
     @organisations ||= for_document_type('organisation')
   end
 
+  def needs
+    @needs ||= for_document_type('need', include_draft: false)
```
tijmenb: What's the reason for excluding drafts here?

cbaines: When they were included, there were a lot more as there are currently lots of draft/proposed needs in Maslow/Need API. So excluding draft content reduces the number to select from.

Also, if it were working, I believe the info frontend only shows validated (published) needs when they are associated with content, and this will be true for the Publishing API as well, as I believe link expansion will ignore draft content (in this case the need) when presenting to the live content store.

I'm not fussed either way on this, as its more of a product decision in an area I still don't have a lot of knowledge about.

tijmenb: Cool, seems fine!

PR #286, opened 2017-02-01
Remove the link type column in spreadsheet uploads

MatMoore:
It's no longer useful, because we only allow taxons now.

This caused some of the errors in https://trello.com/c/rC2btscW/448-content-tagger-show-appropriate-error-messages-when-uploading-untaggable-content

Paired with @klssmith @Davidslv


tijmenb:
Do we have to update a sample spreadsheet somewhere?

MatMoore:
Yes we do, I've put a note on the ticket to update it after we deploy.

PR #288, opened 2017-02-08
Catch lock version conflicts in PublishLinks and retry

rubenarakelyan:
This commit changes the `PublishLinksWorker` to catch `GdsApi::HTTPConflict` exceptions when calling `patch_links`, and to retry such jobs without reporting the exceptions to errbit.

Trello: https://trello.com/c/JYnN3C7L/466-don-t-report-merge-conflicts-in-content-tagger-to-errbit

### DIFF
--- a/app/workers/publish_links_worker.rb
+++ b/app/workers/publish_links_worker.rb
@@ -11,8 +11,21 @@ class PublishLinksWorker
     return if tag_mapping.blank?
 
     if tag_mapping.valid?(:update_links)
-      BulkTagging::PublishLinks.call(tag_mapping: tag_mapping)
-      tag_mapping.mark_as_tagged
+      attempts = 0
+      begin
+        attempts += 1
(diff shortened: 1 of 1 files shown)


tijmenb:
Those conflicts are caused by a classic concurrency issue.

```
links = read_links
links << new_links
write_links(links)
```

So this can happen:

```
links = read_links
links << new_links
# somebody else edits links in publishing-api
write_links(links) # now fails
```

It might make more sense to retry the `BulkTagging::PublishLinks call something like 5 times, instead of using Sidekiq to do this. Thoughts?

rubenarakelyan:
Yup, I've used the pattern from https://github.com/alphagov/rummager/blob/eb1fde14bde10f22d24e2183643c6a4c16576421/lib/services.rb so that we immediately retry up to 5 times and then defer back to Sidekiq to do its thing.

PR #291, opened 2017-02-13
Include internal taxon names in bulk tagging search

rubenarakelyan:
This commit uses the new `search_in` parameter for `get_content_items` to extend the search to the internal taxon name when searching for taxons in the bulk tagging section.

Depends on https://github.com/alphagov/publishing-api/pull/771.

Trello: https://trello.com/c/B7Xj7NFu/370-content-tagger-allow-search-on-internal-taxon-names-including-brackets

### DIFF
--- a/app/services/bulk_tagging/search.rb
+++ b/app/services/bulk_tagging/search.rb
@@ -27,7 +27,8 @@ module BulkTagging
         document_type: document_type,
         page: page,
         q: query,
-        fields: [:content_id, :document_type, :title, :base_path]
+        fields: [:content_id, :document_type, :title, :base_path],
+        search_in: [:title, :base_path, :'details.internal_name']
       )
     end
   end
--- a/spec/services/bulk_tagging/search_spec.rb
+++ b/spec/services/bulk_tagging/search_spec.rb
@@ -10,7 +10,8 @@ module BulkTagging
         q: 'tax',
         page: 1,
         document_type: 'taxon',
-        fields: [:content_id, :document_type, :title, :base_path]
+        fields: [:content_id, :document_type, :title, :base_path],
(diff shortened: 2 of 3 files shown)


app/services/bulk_tagging/search.rb
```diff
@@ -27,7 +27,8 @@ def gds_response
         document_type: document_type,
         page: page,
         q: query,
-        fields: [:content_id, :document_type, :title, :base_path]
+        fields: [:content_id, :document_type, :title, :base_path],
+        search_in: 'title,base_path,details.internal_name'
```
MatMoore: I think this should be an array for consistency?

rubenarakelyan: https://github.com/alphagov/publishing-api/blob/master/doc/api.md#get-v2content currently shows `search_in` as a comma-separated string whereas `fields` is an array, and `gds-api-adapters` passes all the params straight through so I don't think we can make this an array here.

rubenarakelyan: After a discussion with @tijmenb I've raised https://github.com/alphagov/publishing-api/pull/771 to make this consistent since we're the only ones using this feature for now.

MatMoore: We could change it in the publishing api though if we're the only ones using it. I think it's worth doing as it's a bit surprising for it not to work like all other array fields.

tijmenb: @MatMoore yeah, ruben has done that in https://github.com/alphagov/publishing-api/pull/771.

MatMoore: Ah cool. I missed the previous comment when submitting there.

PR #292, opened 2017-02-14
Add rake task for exporting taxon base paths

alecgibson:
This adds a Rake task to export the base paths of all taxons. This should make it easier to update the [navigation prototype](https://github.com/alphagov/govuk-content-navigation), for example.

Trello: https://trello.com/c/bv9RRshW/401-update-the-prototype-with-the-latest-tagged-data

### DIFF
--- a/lib/tasks/taxonomy/export_tagged_content.rake
+++ b/lib/tasks/taxonomy/export_tagged_content.rake
@@ -6,7 +6,7 @@ namespace :taxonomy do
     taxon as a starting point. The script will print to STDOUT base paths for
     all content tagged to the chosen taxon and all of its children.
   DESC
-  task :export_base_paths, [:taxon_id] => :environment do |_, args|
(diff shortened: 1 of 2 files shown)


alecgibson:
@carvil Do you think it's worth renaming the `export_base_paths` Rake task?

carvil:
@alecgibson I think that is a good idea, as the name is a bit misleading.

alecgibson:
@carvil what's it actually exporting? The content tagged to each taxon? So maybe `taxonomy:export_content_tags` ?

carvil:
@alecgibson it traverses the taxonomy and for each taxon fetches content tagged to it, outputting Taxon base path/Content Item base path pairs to output.

What do you think of `taxonomy:export_tagged_content`?

alecgibson:
@carvil done

carvil:
@alecgibson don't think so: https://github.com/search?q=org%3Aalphagov+export_base_paths&type=Code

👍

PR #295, opened 2017-02-23
Temporarily increase publishing api timeout, and cache responses

MatMoore:
Publishing api response time is currently very variable for the linkables
request and is regularly timing out. Since content tagger makes a lot of
linkables requests in succession when tagging a page (to populate the
autocomplete widgets), content tagger has become unusable for actually tagging
content.

This happens on all environments, but I'm not able to reliably reproduce it in
dev at the moment.  Until we get to the bottom of it, it might mitigate the
problem if content tagger waited for the publishing api calls to complete. The
nginx read_proxy_timeout is set to 15 so we should be able to wait longer than
the default 4 seconds.

This is still unbearably slow, but the app is barely working at all at the
moment.  I've also switched on caching. Publishing API uses memcached to cache
the result sets, but getting that data onto the page is still really
inefficient because we make a separate request for each of the 7 tag types.

### DIFF
--- a/lib/services.rb
+++ b/lib/services.rb
@@ -5,6 +5,10 @@ module Services
     @publishing_api ||= GdsApi::PublishingApiV2.new(
       Plek.new.find('publishing-api'),
       disable_cache: true,
+
+      # TODO: revisit this when get_linkables reliably responds in under 4 seconds
+      timeout: 15.seconds,
+
       bearer_token: ENV['PUBLISHING_API_BEARER_TOKEN'] || 'example',
     )
   end


lib/services.rb
```diff
@@ -4,7 +4,11 @@ module Services
   def self.publishing_api
     @publishing_api ||= GdsApi::PublishingApiV2.new(
       Plek.new.find('publishing-api'),
-      disable_cache: true,
+      cache_ttl: 5.minutes,
```
tijmenb: Wondering if we can do this for just the linkables. I'm worried that responses for `get_content_item` will also, be cached. That could cause confusion if the user sees the old version after editing a taxon.

MatMoore: Good point - do you know how publishing apps handle this? I don't see a mechanism for caching just certain calls in gds-api-adapters, but will look into it.

tijmenb: You can wrap it in a cache yourself like this:

```ruby
Rails.cache.fetch("some-key", expires_in: 15.minutes) do
  # get_linkable()
end
```

MatMoore: Ah. That's much better - will rework this. Publishing platform team are also looking at improving the request itself, so hopefully there will be less need for this soon.

MatMoore:
@tijmenb have reworked this to use the rails cache, just for the linkables. I've shortened the time a bit because I think the full 15m could be inconvenient in some cases and there is some upstream caching as well.

MatMoore:
Have removed the caching from this PR so we can potentially deploy it tomorrow morning.

I'll need to look at the caching again as adding this in broke other functionality in content tagger, so we need to isolate it to this form.

cbaines:
> Have removed the caching from this PR so we can potentially deploy it tomorrow morning.

I've got a patch for the Publishing API which will hopefully address the performance issues with the linkables endpoint, so this might not be necessary if that can be deployed promptly.

MatMoore:
I'll close this since https://github.com/alphagov/publishing-api/pull/797 looks like it will do the job, and I'm not going to have time to deploy anything today.

PR #296, opened 2017-02-24
Use `get_content_items` to populate linkables

carvil:
Previously we were using `get_linkables` to fetch all types of links from the Publishing API to populate dropdowns and allow tagging of pages.

These links include:
- taxons
- organisations
- topics
- mainstream browse pages
- needs

These API calls have become increasingly slow which means that in most environments the requests to this particular page timeout.

This commit replaces calls to `get_linkables` to `get_content_items` in the hope to speed the load page of that tagging page and prevent timeouts in all environments.

It also makes the linkable items instances of `ContentItem` rather than hashes, which hopefully makes things more consistent.

What do people think?

cc/ @MatMoore @tijmenb

### DIFF
--- a/app/controllers/copy_taxons_controller.rb
+++ b/app/controllers/copy_taxons_controller.rb
@@ -6,6 +6,6 @@ class CopyTaxonsController < ApplicationController
 private
 
   def taxons
-    Services.publishing_api.get_linkables(format: 'taxon')
+    Linkables.new.get_tags_of_type(:taxon)
   end
 end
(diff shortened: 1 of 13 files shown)


MatMoore:
I think get_linkables might be the faster one now. It's no longer timing out on prod, and there is a further improvement to it here https://github.com/alphagov/publishing-api/pull/802

MatMoore:
As discussed offline let's close this for now.

We think that there's a case for removing get_linkables altogether so we should talk to publishing platform about this and agree what to do with these endpoints.

PR #298, opened 2017-03-02
Add taxon undelete functionality

rubenarakelyan:
This commit adds the ability to undelete previously deleted taxons. Deleted taxons are listed on a separate “trash” page, from where they can be restored.

The taxon deletion process is also changed as follows:

* The delete confirmation page now makes it clear that deleted taxons can be restored
* The wording of the “delete” button is now less scary, since deletions can be reversed

Users attempting to create a new taxon with the same slug as an existing one (live or deleted) now reveive a more specific error message to this effect.

As a bonus, error messages when creating/editing taxons are now displayed correctly, and the values of form fields are not lost if there is an error when attempting to publish the taxon.

Trello: https://trello.com/c/qC6zPIO2/516-bug-taxons-cannot-be-created-in-content-tagger-that-use-the-same-slug-as-a-previously-deleted-taxon

The taxons index page has a button linking to the trash page:

screen shot 2017-03-02 at 11 26 26

This message is displayed when a deleted taxon is restored:

screen shot 2017-03-02 at 11 26 19

This message is displayed if a user attempts to create a new taxon with a slug that is already in use:

screen shot 2017-03-02 at 11 57 23

This message is displayed for all other errors received from the publishing-api:

screen shot 2017-03-02 at 11 57 36

The trash page, with a button back to the taxons index page. Deleted taxons are displayed with a link to restore them:

screencapture-content-tagger-dev-gov-uk-taxons-trash-1488454009880

### DIFF
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -3,14 +3,34 @@ class TaxonsController < ApplicationController
     search_results = remote_taxons.search(
       page: params[:page],
       per_page: params[:per_page],
-      query: query
+      query: query,
+      states: ['published']
     )
 
-    render :index, locals: {
+    locals = {
       taxons: search_results.taxons,
       search_results: search_results,
       query: query,
     }
+
+    render :index, locals: locals
+  end
+
+  def trash
+    search_results = remote_taxons.search(
+      page: params[:page],
+      per_page: params[:per_page],
+      query: query,
+      states: ['unpublished']
+    )
+
+    locals = {
+      taxons: search_results.taxons,
+      search_results: search_results,
+      query: query,
+    }
+
+    render :trash, locals: locals
   end
 
   def new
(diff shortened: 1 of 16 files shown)


app/controllers/taxons_controller.rb
```diff
@@ -93,13 +93,47 @@ def confirm_delete
...
+    Taxonomy::PublishTaxon.call(taxon: taxon)
+    redirect_to taxons_path, flash: flash_message(type: 'restore', response_code: response_code)
+  rescue Taxonomy::PublishTaxon::InvalidTaxonError => e
+    redirect_to taxons_trash_path, flash: { danger: e.message }
+  end
+
 private
 
-  def destroy_flash_message(response_code)
+  def search_and_render_taxons(deleted: false)
```
tijmenb: I think it might be easier to read if we don't extract this code out into a method, but accept a bit of duplication. This method already contains two `if` statements, which we would get rid of putting it back into the action methods. 

The same goes for the flash messages - if they were inlined there would be less code to parse.

app/models/remote_taxons.rb
```diff
@@ -25,7 +25,7 @@ def taxon_content_items(page:, per_page:, query:)
         q: query || '',
         page: page || 1,
         per_page: per_page || 50,
-        states: ["published"],
+        states: states || ['published'],
```
tijmenb: Can we be explicit in the caller instead of the default?

config/routes.rb
```diff
@@ -1,9 +1,13 @@
 Rails.application.routes.draw do
   root to: 'taggings#lookup'
 
+  get '/taxons/trash', to: 'taxons#trash', as: 'taxons_trash'
+
   resources :taxons do
     get :confirm_delete
+    get :restore
```
tijmenb: I think you can do something like `get :trash, on: :collection` for the `trash` method.

app/controllers/taxons_controller.rb
rubenarakelyan: I've split these back out

app/models/remote_taxons.rb
rubenarakelyan: Yup, I've made this an empty array and the default is in the caller

config/routes.rb
rubenarakelyan: Yup, much cleaner now :)

rubenarakelyan:
I've added a restore acceptance test

PR #299, opened 2017-03-08
Find nested tags

alecgibson:
This rake task will process all content tagged under a root taxon, and find any content tagged to a taxon where the content is also tagged to an ancestor of that taxon.

### Trello

https://trello.com/c/cuRxpiUD/465-create-a-list-of-all-content-pages-tagged-both-at-a-leaf-taxon-and-at-another-taxon-up-in-that-same-tree

### DIFF
--- a/lib/tasks/taxonomy/find_nested_tags.rake
+++ b/lib/tasks/taxonomy/find_nested_tags.rake
@@ -45,7 +45,12 @@ namespace :taxonomy do
 
       next unless already_tagged
 
+      content_links = Services.publishing_api.get_expanded_links(content['content_id']).to_h
+      organisations = content_links.dig('expanded_links', 'organisations') || []
+      organisation_titles = organisations.map { |organisation| organisation['title'] }
+
       nested_tag_warnings << {
+        organisation: organisation_titles.join('; '),
         content_base_path: content_base_path,
         content_title: content['title'],
         content_id: content['content_id'],
@@ -66,7 +71,7 @@ namespace :taxonomy do
 
   def clone_hash_of_hashes(hash_of_hashes)
     hash_of_hashes.each_with_object({}) do |(key, hash), clone|
-      clone.merge!(key => hash.clone)
+      clone[key] = hash.clone
     end
   end
 
@@ -81,6 +86,7 @@ namespace :taxonomy do
 
       unique_nested_tags.each do |tags|
         row = [
+          content_warnings.first[:organisation],
           content_warnings.first[:content_id],
           content_warnings.first[:content_base_path],
           content_warnings.first[:content_title],
--- a/lib/tasks/taxonomy/find_nested_tags.rake
+++ b/lib/tasks/taxonomy/find_nested_tags.rake
@@ -1,0 +1,108 @@
+require 'set'
+require 'csv'
+
+namespace :taxonomy do
+  desc "Find any content tagged at one taxon level, then tagged again at one of those taxon's ancestors"
+  task :find_nested_tags, [:root_taxon_id] => [:environment] do |_, args|
+    root_taxon_id = args.fetch(:root_taxon_id)
+    root_taxon = OpenStruct.new Services.publishing_api.get_content(root_taxon_id).to_h
+    taxonomy = Taxonomy::ExpandedTaxonomy.new(root_taxon.content_id)
+    taxonomy.build
+
+    nested_tag_warnings = find_content_tagged_at_multiple_levels(taxonomy.child_expansion)
+
+    output = format_warnings_as_rows(nested_tag_warnings)
+
+    output.each do |row|
+      puts row.to_csv
+    end
+  end
+
+  def find_content_tagged_at_multiple_levels(taxon, content_already_tagged = {})
+    taxon_base_path = taxon.base_path
+
+    taxon_content = Services.publishing_api.get_linked_items(
+      taxon.content_id,
+      link_type: "taxons",
+      fields: %w(base_path content_id title)
+    ).to_a
+
+    nested_tag_warnings = []
+
+    taxon_content.each do |content|
+      content_base_path = content['base_path']
+      already_tagged = false
+
+      if content_already_tagged.key?(content_base_path)
+        already_tagged = true
+      else
+        # Use a hash rather than a set, since sets do not guarantee ordering, but hash keys enumerate in insertion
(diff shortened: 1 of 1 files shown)


lib/tasks/taxonomy/find_nested_tags.rake
```diff
@@ -0,0 +1,55 @@
...
+
+    errors = []
+
+    taxon_content.each do |content|
+      content_base_path = content['link']
+
+      if content_already_tagged.key?(content_base_path)
+        errors << "#{content_base_path} tagged at #{taxon_base_path}, but also tagged at ancestors: #{content_already_tagged[content_base_path].to_a}"
+      else
+        content_already_tagged[content_base_path] = Set.new
```
suzannehamilton: You can avoid doing this by initializing the hash using `Hash.new { |h, k| h[k] = Set.new }` rather than `{}`.

lib/tasks/taxonomy/find_nested_tags.rake
```diff
@@ -0,0 +1,55 @@
...
+namespace :taxonomy do
+  desc "Find any content tagged at one taxon level, then tagged again at one of those taxon's ancestors"
+  task :find_nested_tags, [:root_taxon_base_path] => [:environment] do |_, args|
+    root_taxon_base_path = args[:root_taxon_base_path]
+
+    content_store = GdsApi::ContentStore.new(Plek.find('content-store'))
+
+    root_taxon = content_store.content_item(root_taxon_base_path).to_h
+
+    errors = find_content_tagged_at_multiple_levels(root_taxon)
```
suzannehamilton: I wonder if there's a better name than "errors". There's nothing in the task's name or description which hints that content tagged at multiple levels is an error, and when I first saw it I assumed it was rake task error handling rather than the expected result.

lib/tasks/taxonomy/find_nested_tags.rake
```diff
@@ -0,0 +1,55 @@
...
+require 'set'
+
+namespace :taxonomy do
+  desc "Find any content tagged at one taxon level, then tagged again at one of those taxon's ancestors"
+  task :find_nested_tags, [:root_taxon_base_path] => [:environment] do |_, args|
+    root_taxon_base_path = args[:root_taxon_base_path]
+
+    content_store = GdsApi::ContentStore.new(Plek.find('content-store'))
+
+    root_taxon = content_store.content_item(root_taxon_base_path).to_h
```
carvil: You should probably get it from the publishing API to avoid another dependency, as this is a publishing app. That means either passing in a content_id, or do a search by base_path.

An example with content ID would be:

```
Services.publishing_api.get_content(taxon_content_id).to_h
```

lib/tasks/taxonomy/find_nested_tags.rake
```diff
@@ -0,0 +1,55 @@
...
+      content_already_tagged[content_base_path] << taxon_base_path
+    end
+
+    puts "#{taxon_base_path}: #{taxon_content.size} items with #{errors.size} errors"
+
+    child_taxons = taxon.dig('links', 'child_taxons')
+
+    if child_taxons
+      child_taxons.each do |child_taxon|
+        errors += find_content_tagged_at_multiple_levels(child_taxon, content_already_tagged.clone)
```
suzannehamilton: Just to check - does `clone` create copies of the sets in the hash?

lib/tasks/taxonomy/find_nested_tags.rake
alecgibson: I'm not sure I understand?

alecgibson: I recognise that this would be nice, but the niceness isn't as nice with the new clone method I need.

lib/tasks/taxonomy/find_nested_tags.rake
alecgibson: Done

lib/tasks/taxonomy/find_nested_tags.rake
alecgibson: Done

lib/tasks/taxonomy/find_nested_tags.rake
alecgibson: Good catch - fixed

alecgibson:
@suzannehamilton @carvil I've addressed your comments - can you please re-review?

I've run it on int here: https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/1327/console

carvil:
@alecgibson could you please output this in a CSV format? It would be nice to copy/paste the errors into a spreadsheet.

Could you please speak with Holly or Graeme to see what the best format would be for this?

alecgibson:
@carvil I've updated it to output CSV, but still waiting on Holly and/or Graeme's input.

alecgibson:
@carvil @suzannehamilton this is now good to go. It's outputting CSV. After all that, I don't think there are actually any content items with triple-nested taxons (which would be pretty crazy anyway?).

https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/1330/console

alecgibson:
Updated to include organisations: https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/1342/console

lib/tasks/taxonomy/find_nested_tags.rake
```diff
@@ -0,0 +1,124 @@
...
+  desc "Find any content tagged at one taxon level, then tagged again at one of those taxon's ancestors"
+  task :find_nested_tags, [:root_taxon_id] => [:environment] do |_, args|
+    root_taxon_id = args.fetch(:root_taxon_id)
+    root_taxon = OpenStruct.new Services.publishing_api.get_content(root_taxon_id).to_h
+    taxonomy = Taxonomy::ExpandedTaxonomy.new(root_taxon.content_id)
+    taxonomy.build
+
+    nested_tag_warnings = find_content_tagged_at_multiple_levels(taxonomy.child_expansion)
+
+    output = format_warnings_as_rows nested_tag_warnings
```
carvil: Could you please wrap arguments in parenthesis? It' makes it easier to read and consistent with other method calls.

alecgibson: Done

alecgibson:
@suzannehamilton could you please re-review?

PR #300, opened 2017-03-08
Measure patch links, lookups and linkables

tijmenb:
This adds statsd timers for patch links, lookups and linkables. This will allow us to make a dashboard that shows how the app is performing.


suzannehamilton:
LGTM 👍

PR #301, opened 2017-03-08
Add rake task for finding taxon-tagged content by document type

suzannehamilton:
This will be used to check our assumptions about whether content with a given document type is really 'guidance' and should appear on taxon pages.

Trello: https://trello.com/c/m3sc4i5N/11-notices-appearing-as-guidance-content

### DIFF
--- a/lib/tasks/taxonomy/export_tagged_content.rake
+++ b/lib/tasks/taxonomy/export_tagged_content.rake
@@ -15,7 +15,7 @@ namespace :taxonomy do
     taxonomy.build
 
     taxons = taxonomy.child_expansion.map do |node|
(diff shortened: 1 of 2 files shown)


tijmenb:
Wondering if this data isn't already accessible via the search API?

https://www.gov.uk/api/search.json?filter_part_of_taxonomy_tree[]=c58fdadd-7743-46d6-9629-90bb3ccc4ef0&filter_content_store_document_type=notice

suzannehamilton:
@tijmenb You're right - I didn't realise we could do this via search. 🙂  Is it worth reverting this change so we don't have another rake task to maintain?

tijmenb:
Yeah, good idea!

PR #302, opened 2017-03-08
Fix broken rake tasks

MatMoore:
`content_id` and `base_path` are directly accessible from taxons using the new taxonomy gem.


suzannehamilton:
LGTM 👍

PR #303, opened 2017-03-08
Add ordered related items overrides

MatMoore:
Add new field for related items overrides. Depends on https://github.com/alphagov/govuk-content-schemas/pull/560

This will be used for the beta navigation we are deploying tomorrow for education. It allows publishers to override the automatically generated related content.

`A` group users will see the existing `ordered_related_items`, for mainstream content only.  `B` group users will see `ordered_related_items_overrides` if set, otherwise automatically generated links. The ordered_related_items will be ignored.

We thought about combining the views for `ordered_related_items` and `ordered_related_items_overrides` but couldn't think of a clean way to do it so decided to leave them separate.

Trello https://trello.com/c/vM6Kaw9B/505-ias-and-content-designers-can-curate-related-links-for-any-content-type
Paired with @rubenarakelyan  and @klssmith

### DIFF
--- a/app/assets/javascripts/views/related-content-tagger.js
+++ b/app/assets/javascripts/views/related-content-tagger.js
@@ -2,16 +2,23 @@
   "use strict";
 
   Modules.RelatedContentTagger = function () {
-    this.start = function () {
+    this.start = function (fieldset) {
+      var $fieldset = $(fieldset);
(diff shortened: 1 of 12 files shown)


tijmenb:
🤔 the tests for this should not be green until https://github.com/alphagov/govuk-content-schemas/pull/560 is deployed. Can we add a failing schema test for this?

PR #306, opened 2017-03-14
Taxonomy analysis

alecgibson:
Added a new Rake task for analysing the taxons tagged to each content item. It will output some information on the content item (ID and base path), as well as the number of taxons tagged to that content, and their IDs.

This change also adds a new `analysis` namespace to the Rake tasks.


alecgibson:
I've run this on int, and the results are here: https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/1367/console

PR #307, opened 2017-03-15
Rake task to upload the parenting and childcare taxonomy

carvil:
This rake task takes a CSV file containing taxon titles, descriptions
and parents, and creates them in content tagger.

The reason for this task is so we can upload the new taxonomy to staging
and perform a data dump for the prototype. It will also let us bulk tag
all pieces of content related to the parenting and childcare branch, so
they are available in search. This way, we can point the prototype to
staging and be able to show both scoped search results and related
links.

**NOTE**: this rake task is temporary and only needed until we can
create draft taxons live. Since the prototype will be tested before
that's available, we need it as a temporary step.

There is no need to merge this branch, as it can be deployed to integration and staging as it is.

### DIFF
--- a/app/services/parenting_childcare/build_taxon.rb
+++ b/app/services/parenting_childcare/build_taxon.rb
@@ -1,0 +1,20 @@
+module ParentingChildcare
+  class BuildTaxon
+    def self.build(title:, description:, prefix:)
+      taxon_params = {
+        path_prefix: prefix,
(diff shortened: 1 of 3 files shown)


tijmenb:
Rebased on https://github.com/alphagov/content-tagger/pull/316. Merging now, so that we can update https://github.com/alphagov/content-tagger/pull/318 and have this work with the draft workflow.

PR #314, opened 2017-03-21
Show list of draft taxons

tijmenb:
This shows the user a list of draft taxons. Also changes the interface to navigate between the lists of taxons in live/draft/deleted state.

Note that it's currently not possible for users to create draft taxons, they're instantly published on creation. Functionality to create drafts and publish them will be implemented in a separate PR.  

## Before

screen shot 2017-03-21 at 11 22 24

## After

screen shot 2017-03-21 at 11 22 15

screen shot 2017-03-21 at 11 24 15

https://trello.com/c/8YH5l9kF

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -49,3 +49,9 @@ $red: #b10e1e;
 .left-space {
   margin-left: 5px;
 }
+
+// taxons page
+
+.nav.nav-tabs {
+  margin-bottom: 20px;
+}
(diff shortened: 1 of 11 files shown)


MatMoore:
FYI I think this line is a bug https://github.com/alphagov/content-tagger/blob/master/app/models/content_item.rb#L21

```
-    @state = data.fetch('state', nil)
+    @state = data.fetch('publication_state', nil)
```

I noticed this when working on draft tagging but had to drop this and forgot to raise a PR. Can you address this here (assuming it is a bug as I thought)?

tijmenb:
@MatMoore Hmm, good catch there. I'll fix that in a separate PR though, I think we're not using `ContentItem` in this flow (maybe there's some refactoring to be done to fix this).

MatMoore:
Oops sorry, I thought you were but it was `Taxon`.

PR #315, opened 2017-03-21
Add rake task to register a taxonomy email signup page

mobaig:
Create a content item at the base path /taxonomy-email-signup. The
email-alert-frontend app will be updated to serve signup pages at this
endpoint, allowing users to subscribe to content changes in the
taxonomy.

https://trello.com/c/lPlSbZnX/166-3-build-3-types-of-sign-up-pages-backend

### DIFF
--- a/lib/tasks/taxonomy/register_email_signup_page.rake
+++ b/lib/tasks/taxonomy/register_email_signup_page.rake
@@ -1,0 +1,46 @@
+namespace :taxonomy do
+  desc <<-DESC
+    Register a content item representing the email-alert-frontend endpoint that
+    handles taxonomy email alert subscriptions
+  DESC
+  task register_email_signup_page: :environment do
+    def base_path
+      '/taxonomy-email-signup'
+    end
+
+    def routes
+      [
+        { path: base_path, type: "exact" },
+        { path: "#{base_path}/confirm", type: "exact" },
+      ]
+    end
+
+    # The base_path and routes specified here are the only important values -
+    # we want to ensure the specified path is reserved correctly via the
+    # publishing-api. All other fields are populated with nominal values,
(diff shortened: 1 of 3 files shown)


mobaig:
cc: @tijmenb

tijmenb:
Maybe we should get a 2nd opinion on the URL, maybe "taxonomy" is too internal-sounding? We could check with Holly or Rob.

lib/tasks/taxonomy/register_email_signup_page.rake
```diff
@@ -0,0 +1,46 @@
...
+        { path: "#{base_path}/confirm", type: "exact" },
+      ]
+    end
+
+    # The base_path and routes specified here are the only important values -
+    # we want to ensure the specified path is reserved correctly via the
+    # publishing-api. All other fields are populated with nominal values,
+    # particularly the details hash which needs to satisfy the schema
+    # definition for an email_alert_signup.
+    email_signup_page = {
```
tijmenb: Can you extract this payload and test it against the schema? Otherwise it might get out of date.

lib/tasks/taxonomy/register_email_signup_page.rake
```diff
@@ -0,0 +1,46 @@
...
+      base_path: base_path,
+      document_type: 'email_alert_signup',
+      schema_name: 'email_alert_signup',
+      title: 'Get email alerts',
+      description: '',
+      locale: 'en',
+      need_ids: [],
```
tijmenb: This is likely to be removed in the future, I think we can omit this here.

mobaig:
@tijmenb Comments addressed, ready for another review. Spoke to Rob, opted for `/email-alert` instead.

PR #319, opened 2017-03-21
Use edition level links for "parent_taxons"

tijmenb:
This moves the `parent_taxons` links to the edition-level. This means that the links will be updated in the same call as the content. It saves a call to the publishing-api and ties in nicely with the new draft taxon workflow.

Needs a corresponding change in govuk-content-schemas: https://github.com/alphagov/govuk-content-schemas/pull/587, until that is deployed this PR will fail.

Trello: https://trello.com/c/8YH5l9kF

### DIFF
--- a/app/services/taxonomy/build_taxon_payload.rb
+++ b/app/services/taxonomy/build_taxon_payload.rb
@@ -23,6 +23,9 @@ module Taxonomy
           internal_name: internal_name,
           notes_for_editors: notes_for_editors,
         },
+        links: {
+          parent_taxons: taxon.parent_taxons.select(&:present?)
+        },
         routes: [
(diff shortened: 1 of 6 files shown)


app/services/taxonomy/publish_taxon.rb
```diff
@@ -33,6 +29,7 @@ def publish(validate: true)
       )
 
       if existing_content_item.nil?
+        Rails.logger.info(e.message)
```
rubenarakelyan: Is this a debugging call or intentionally writing to the log?

tijmenb: Removed that commit.

PR #322, opened 2017-03-22
Refactor taxon controller & views

tijmenb:
The `TaxonsController` has become too big and has contains a lot of logic. This refactors the controller and views.

The guiding principle is that any view should only be passed one object from the controller. See Sandi's rule No. 4:

> "Only instantiate one object in the controller"

https://robots.thoughtbot.com/sandi-metz-rules-for-developers#only-instantiate-one-object-in-the-controller

To do this we create 3 view objects, IndexPage, ShowPage, and EditPage. These objects know the data to render the page. It simplifies the controller a lot. We could probably also simplify the views a bit, but I haven't gotten to that in this PR.

Commit also inlines some HTML in the deletion warning template to make it easier to change.

cc @carvil


carvil:
👍

PR #329, opened 2017-03-23
Fix bug with specifying parents

tijmenb:
In https://github.com/alphagov/content-tagger/pull/319 we switched the taxon system to use "edition" level links. But we forgot to port over the reading of the links. Turns out that if you ask the publishing-api for links, the edition level links won't be returned.

By fixing this bug we can also remove a lot of stubbing!

https://trello.com/c/8YH5l9kF


tijmenb:
Closing because I'm reverting the original PR.

PR #332, opened 2017-03-24
Add discarding of drafts

tijmenb:
This adds functionality so users can discard drafts.

Code by @klssmith.

## Screens

screen shot 2017-03-24 at 10 44 20

---

screen shot 2017-03-24 at 10 44 26

---

screen shot 2017-03-24 at 10 44 32

https://trello.com/c/8YH5l9kF

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -109,6 +109,7 @@ en:
       destroy_alert: It was not possible to delete the taxon
       restore_success: You have sucessfully restored the taxon
       restore_alert: It was not possible to restore the taxon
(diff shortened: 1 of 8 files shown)


config/locales/en.yml
```diff
@@ -109,6 +109,7 @@ en:
       destroy_alert: It was not possible to delete the taxon
       restore_success: You have sucessfully restored the taxon
       restore_alert: It was not possible to restore the taxon
+      discard_draft_success: You have successfully discarded the draft taxon
```
rubenarakelyan: Maybe it's better to call this "deleted" rather than "discarded" to keep everything consistent

PR #337, opened 2017-04-05
Add fallback OAuth ID and secret

rubenarakelyan:
This commit adds a fallback OAuth ID and secret. This will allow the signon-munging script for development to configure signon for this app locally.

It also renames the file to make it consistent with other apps and allow the munging script to find it.


rubenarakelyan:
We could keep it with the underscore but then I'd have to include an exception in the signon script in the dev VM since all the other apps use `gds-sso.rb`...

PR #342, opened 2017-04-19
Inline taxonomy helpers gem

tijmenb:
This inlines the relevant classes from the https://github.com/alphagov/govuk_taxonomy_helpers gem.

The taxonomy gem was part of an effort to avoid duplication between applications dealing with the taxonomy. Unfortunately we ran out of time last quarter, which left us with an API that was still in flux
(alphagov/govuk_taxonomy_helpers#11).

To avoid confusion, I think it's best to inline the code in the places where we use the taxonomy helpers. We're going to be doing work around the taxonomy in the next couple of months and I think it will be easier to extract code than trying to nail the API right now.

Trello: https://trello.com/c/FVfbiKxn

### DIFF
--- a/lib/govuk_taxonomy_helpers/linked_content_item.rb
+++ b/lib/govuk_taxonomy_helpers/linked_content_item.rb
@@ -1,0 +1,127 @@
+require 'forwardable'
+
+module GovukTaxonomyHelpers
+  # A LinkedContentItem can be anything that has a content store representation
+  # on GOV.UK.
+  #
+  # It can be used with "taxon" content items (a topic in the taxonomy) or
+  # other document types that link to taxons.
+  #
+  # Taxon instances can have an optional parent and any number of child taxons.
+  class LinkedContentItem
+    extend Forwardable
+    attr_reader :title, :content_id, :base_path, :children, :internal_name
+    attr_accessor :parent
+    attr_reader :taxons
+    def_delegators :tree, :map, :each
+
+    # Extract a LinkedContentItem from publishing api response data.
+    #
+    # @param content_item [Hash] Publishing API `get_content` response hash
+    # @param expanded_links [Hash] Publishing API `get_expanded_links` response hash
+    # @return [LinkedContentItem]
+    # @see http://www.rubydoc.info/gems/gds-api-adapters/GdsApi/PublishingApiV2#get_content-instance_method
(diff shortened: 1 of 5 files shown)


lib/govuk_taxonomy_helpers/publishing_api_response.rb
```diff
@@ -0,0 +1,91 @@
+module GovukTaxonomyHelpers
```
whoojemaflip: feels like this should have unit tests

lib/govuk_taxonomy_helpers/linked_content_item.rb
```diff
@@ -0,0 +1,126 @@
...
+
+module GovukTaxonomyHelpers
+  # A LinkedContentItem can be anything that has a content store representation
+  # on GOV.UK.
+  #
+  # It can be used with "taxon" content items (a topic in the taxonomy) or
+  # other document types that link to taxons.
+  #
+  # Taxon instances can have an optional parent and any number of child taxons.
+  class LinkedContentItem
```
whoojemaflip: tests?

tijmenb:
@whoojemaflip I'm not sure if it would add much. The copied code is used in `ExpandedTaxonomy`, which has a lot of tests:

https://github.com/alphagov/content-tagger/blob/master/spec/models/taxonomy/expanded_taxonomy_spec.rb

Some of it is also covered in the integration tests:

https://github.com/alphagov/content-tagger/blob/master/spec/features/viewing_taxons_spec.rb

I could copy over https://github.com/alphagov/govuk_taxonomy_helpers/tree/master/spec, but I think it may not add much value right now. I suspect we'll end up doing the things copied here differently anyway in the coming months. What do you think?

whoojemaflip:
I'm not familiar with the plan for the taxonomy helpers gem, but it strikes me that any iteration of the API will likely happen inline with application code. In this case, it will benefit our future selves to have the tests alongside the code to aid refactoring.

tijmenb:
On second thought, let's leave the taxonomy gem where it is. I wanted to inline it to make sure that there wasn't any confusion about the status (it shouldn't be used). But I don't think it's worth it to drag in 500 lines of questionable specs into this project. And perhaps the inlining will be confusing too.

As part of https://trello.com/c/rNpmwHt8 we'll probably need to change the UI of this app, which will give us a opportunity to rewrite the taxonomy tree-management code (in gem or not).

PR #343, opened 2017-04-20
Upgrade gems

klssmith:
This upgrades most of the gems in the Gemfile in preparation for upgrading to Rails 5. 
(web-console and kaminari cannot be upgraded before the Rails upgrade.) 

[Trello card](https://trello.com/c/UvadVLm8/72-upgrade-dependencies-on-our-applications-including-rails)


Gemfile
```diff
@@ -42,7 +41,7 @@ group :development do
 end
 
 group :test do
-  gem 'webmock', '~> 1.22'
+  gem 'webmock', '~> 3.0'
```
whoojemaflip: is there value in pinning the versions of `test` gems?

klssmith: I don't know - I can remove the version here.

whoojemaflip: :+1:

klssmith: @whoojemaflip Now fixed.

PR #348, opened 2017-04-27
Add task to create a new taxonomy tree

whoojemaflip:
Run this `rake` task with the content_id of a taxon node as an argument.

```
$ rake taxonomy:promote_taxon_to_taxonomy_root[8a2141ae-2493-491c-8e5f-2bd23c1a6afe]
```

It will patch this content_item, setting the Homepage `/` as it's parent
taxon. Causing the chosen taxon to become the root node of a top level
taxonomy tree.

[Trello](https://trello.com/c/wHZQz5Yk/64-build-joe-s-designs-to-allow-publishers-to-tag-across-multiple-themes-in-whitehall)

Paired with @klssmith

### DIFF
--- a/lib/tasks/taxonomy/promote_taxon_to_taxonomy_root.rake
+++ b/lib/tasks/taxonomy/promote_taxon_to_taxonomy_root.rake
@@ -1,0 +1,28 @@
+namespace :taxonomy do
+  desc "Promote a taxon to become a taxonomy root"
+  task :promote_taxon_to_taxonomy_root, [:taxon_id] => :environment do |_, args|
+    taxon_id = args.fetch(:taxon_id)
+    homepage_content_id = "f3bbdec2-0e62-4520-a7fd-6ffd5d36e03a"
+
+    puts "Promoting content with ID: #{taxon_id} to a taxonomic root node."
+
+    begin
+      homepage_links = Services.publishing_api.get_links(
+        homepage_content_id
+      )
+
+      root_taxons = homepage_links['links'].fetch('root_taxons', [])
+      root_taxons << taxon_id
+
+      Services.publishing_api.patch_links(
+        homepage_content_id,
+        links: { root_taxons: root_taxons.uniq },
+      )
+
+      puts '✅  OK'
+    rescue GdsApi::BaseError => e
+      puts "❌  FAILURE #{e.code}"
+      exit(1)
+    end
+  end
+end


tijmenb:
This might not work for a couple of reasons.

1. In the new navigation the `parent_taxons` are used for the breadcrumbs. For example [this navigation page](https://www.gov.uk/education/funding-for-different-types-of-schools-and-settings) and [this content page](https://www.gov.uk/government/publications/academies-and-the-dedicated-schools-grant-2014-to-2015). This is done by [loading the content item](https://github.com/alphagov/government-frontend/blob/db8de5071bd5b0412ece244ed4083d57fb33dd0e/app/presenters/content_item_presenter.rb#L15) into the navigation helpers [and generating a breadcrumbs payload](https://github.com/alphagov/government-frontend/blob/d0c0da319ef131576076308469ccc1d85292728f/app/views/shared/_breadcrumbs.html%2Bnew_navigation.erb#L18-L20) by recursively following the parent taxons.

    This means putting the homepage in the tree will probably cause the homepage to show up in the breadcrumbs twice.

2. Since the homepage isn't a taxon, so updating a root taxon like education from the content tagger will possibly remove what we do in this rake task (because the select box will be empty).

3. The content tagger interface builds a [tree of the taxonomy](https://content-tagger.publishing.service.gov.uk/taxons/90e8a226-8d40-4c1f-91e4-36a712b34e57) based on `parent_taxons`. This means we'd also show the homepage in that tree. Since we currently expect `parent_taxons` to be a taxon, some things like links to the taxon page might break.

Perhaps the safest way to achieve this would be to create a new `root_taxons` link type for the homepage.

whoojemaflip:
Interesting points.

Regarding /1 above, I suspect we will actually make the breadcrumbs code a little simpler by instigating `/` as the taxonomy/navigation root. 

For example, here: https://github.com/alphagov/govuk_navigation_helpers/blob/master/lib/govuk_navigation_helpers/taxon_breadcrumbs.rb#L17

whoojemaflip:
2 and 3 could be solved potentially by making the homepage into a `Taxon`

whoojemaflip:
Taking all this on board, and deciding it's probably easier and less confusing to:

- create a new schema type for  `/`
- create a new link type in that schema for `taxon_roots`

Schema created here:
https://github.com/alphagov/govuk-content-schemas/pull/610

PR #349, opened 2017-04-28
Blacklist related item tagging based on document type

andrewgarner:
As formats are migrated their rendering apps change. This switches the logic to decide whether or not to blacklist tagging of ordered related items from the rendering app to the document type. We shouldn't expect a rendering app to render all the formats it supports in the same way, for example, [government-frontend](github.com/alphagov/government-frontend) renders travel advice pages that [support related items](https://github.com/alphagov/government-frontend/blob/master/app/presenters/travel_advice_presenter.rb#L47-L63), and also speeches that don't.

This is a more robust fix to #347

The list of document types that have related links was hand crafted by consulting the example pages linked from https://docs.publishing.service.gov.uk/document-types.html

See https://trello.com/c/fnqWUB1w and https://trello.com/c/D2ocCxfD and https://twitter.com/kentbeck/status/250733358307500032


spec/models/content_item_spec.rb
```diff
@@ -1,124 +1,75 @@
...
 
 RSpec.describe ContentItem do
-  let(:content_item) do
-    ContentItem.new(
-      content_item_params
-    ).tap do |content_item|
-      content_item.link_set = Tagging::ContentItemExpandedLinks.new(link_set_params)
-    end
-  end
+  subject(:content_item) { described_class.new(data, blacklist: blacklist) }
```
tijmenb: Hmm, not sure about this. We're not using `subject` other specs in this app, and I'd probably prefer be consistent (but on the hand these specs weren't great to begin with, so I'd like to refactor it some day anyway).

andrewgarner: Using `subject` allows us to make more concise assertions such as:

```rb
it { is_expected.to include(:ordered_related_items_overrides) }
```

instead of:

```rb
it "should include :ordered_related_items_overrides" do
  expect(content_item.blacklisted_tag_types).to include(:ordered_related_items_overrides)
end 
```

IMHO it is now easier to see what changes between each test context and the effect that has on the method under test.

tijmenb: Yeah, and I dislike test contexts too :) But that's all a different discussion. We'll have a discussion in a different PR.

PR #350, opened 2017-05-08
Add support for setting the visibility flag

klssmith:
UI and presenter code to support updating the
`visible_to_departmental_editors` flag on a Taxon.

Setting this flag on a draft taxon will allow editors
to see this taxon, and tag content to it.

It will not affect the public visibility however.

### Edit taxon page with new checkbox:
screen shot 2017-05-09 at 12 00 03

Paired with @whoojemaflip 

[Trello](https://trello.com/c/3xt6aNdb/73-make-certain-taxons-visible-as-drafts-to-whitehall-editors)

### DIFF
--- a/app/views/taxons/_form_fields.html.erb
+++ b/app/views/taxons/_form_fields.html.erb
@@ -31,12 +31,14 @@
   hint: I18n.t('views.taxons.displayed_on_govuk'),
   input_html: { class: 'form-control' } %>
 
-<%= f.input :visible_to_departmental_editors,
-  as: :boolean,
-  input_html: { class: 'form-control' },
-  checked_value: 'true',
-  label: I18n.t('views.taxons.visible_to_departmental_editors'),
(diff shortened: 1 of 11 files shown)


klssmith:
Tests are currently failing because the change to the taxon schema in govuk-content-schemas needs to be deployed.

app/views/taxons/_form_fields.html.erb
```diff
@@ -31,6 +31,17 @@
...
   input_html: { class: 'form-control' } %>
 
+<% if page.show_visibilty_checkbox? %>
+  <%= f.input :visible_to_departmental_editors,
+    as: :boolean,
+    input_html: { class: 'form-control' },
+    checked_value: 'true',
+    label: I18n.t('views.taxons.visible_to_departmental_editors'),
+    hint: I18n.t('views.taxons.visible_to_departmental_editors_hint')  %>
+<% else %>
```
whoojemaflip: are we sure we want this setting to persist between taxon state changes? It might make for some unexpected behaviour at some point...

PR #351, opened 2017-05-09
Analyse the number of mainstream items tagged at and below taxons

alecgibson:
In order to support design work on surfacing mainstream content, we need
to collect information at how much mainstream content is tagged at a
given level, or below that level.

This change adds a Rake task that will perform this analysis and output
a CSV.

### Trello

https://trello.com/c/k9BUzkDP/114-identify-the-number-of-mainstream-content-items-at-and-below-each-level-of-the-taxonomy

### DIFF
--- a/lib/tasks/taxonomy/analysis/mainstream.rake
+++ b/lib/tasks/taxonomy/analysis/mainstream.rake
@@ -1,0 +1,123 @@
+require 'csv'
+require 'set'
+
+namespace :taxonomy do
+  namespace :analysis do
+    desc 'Determine how much mainstream content is tagged at and below each taxon'
+    task :mainstream_analysis, [:root_taxon_id] => [:environment] do |_, args|
+      root_taxon_id = args.fetch(:root_taxon_id)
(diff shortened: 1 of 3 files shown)


alecgibson:
Results from running on Integration are here: https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/1895/console

lib/tasks/taxonomy/analysis/mainstream.rake
```diff
@@ -0,0 +1,123 @@
...
+    def number_of_mainstream_items_in(content_items)
+      content_items.count do |content_item|
+        mainstream_document_types.include? content_item['document_type']
+      end
+    end
+
+    def number_of_guidance_items_in(content_items)
+      content_items.count do |content_item|
+        supertypes = GovukDocumentTypes.supertypes(document_type: content_item['document_type'])
+        supertypes['navigation_document_supertype'] == 'guidance'
```
tijmenb: `navigation_document_supertype` is also available in search, so you shouldn't have to use the document type gem:

https://www.gov.uk/api/search.json?fields[]=navigation_document_supertype

PR #355, opened 2017-05-16
Allow a Taxon to be saved without a linked parent taxon

whoojemaflip:
It should be possible to save a Taxon in content-tagger without
a parent taxon. Previously, publishing-api would error on the patch-links call.

This makes it possible to save a taxon without a linked parent taxon.

### DIFF
--- a/app/services/taxonomy/update_taxon.rb
+++ b/app/services/taxonomy/update_taxon.rb
@@ -19,10 +19,8 @@ module Taxonomy
       end
 
       Services.publishing_api.put_content(content_id, payload)
-      Services.publishing_api.patch_links(
-        content_id,
-        links: { parent_taxons: Array(parent) }
-      )
+      Services.publishing_api.patch_links(content_id, links: links)
+
     rescue GdsApi::HTTPUnprocessableEntity => e
       # Since we cannot easily differentiate the reasons for getting a 422
       # error code, we do a lookup to see if a content item with the slug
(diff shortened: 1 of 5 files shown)


spec/features/delete_taxon_spec.rb
```diff
@@ -144,7 +143,6 @@ def then_the_taxon_is_deleted
 
   def then_the_taxon_is_restored
     expect(@put_content_request).to have_been_made
-    expect(@patch_links_request).to have_been_made
```
tijmenb: Why is this removed?

app/services/taxonomy/update_taxon.rb
```diff
@@ -19,10 +19,13 @@ def publish(validate: true)
       end
 
       Services.publishing_api.put_content(content_id, payload)
-      Services.publishing_api.patch_links(
-        content_id,
-        links: { parent_taxons: Array(parent) }
-      )
+
+      if parent.present?
```
tijmenb: Will this still allow the user to _remove_ the parent?

whoojemaflip: This class handles the user's form submission, so if there is no parent taxon submitted in the UI then `taxon.parent` will be an empty string.

spec/features/delete_taxon_spec.rb
whoojemaflip: lazy programmer

app/services/taxonomy/update_taxon.rb
whoojemaflip: ... so: yes

spec/features/delete_taxon_spec.rb
whoojemaflip: resolved

app/services/taxonomy/update_taxon.rb
whoojemaflip: OIC

PR #358, opened 2017-05-30
Allow users to download tagged content as CSV

tijmenb:
This adds a button to the taxon show page that allows users to download the tagged content as a CSV. The resulting CSV has been checked by using the Upload feature in Google Sheets.

https://trello.com/c/ezC0aq3a

### DIFF
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -106,6 +106,12 @@ class TaxonsController < ApplicationController
     redirect_to taxons_path, success: t("controllers.taxons.discard_draft_success")
   end
 
+  def download_tagged
+    export = Taxonomy::TaxonomyExport.new(taxon.content_id)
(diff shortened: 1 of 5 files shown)


app/controllers/taxons_controller.rb
```diff
@@ -106,6 +106,12 @@ def discard_draft
     redirect_to taxons_path, success: t("controllers.taxons.discard_draft_success")
   end
 
+  def download_tagged
```
whoojemaflip: would this be better expressed as `GET /taxons.csv` or is the functionality so different?

tijmenb: The functionality is different yeah, this downloads the content tagged to a taxon, not the taxons itself.

tijmenb:
Updated to include content `description` as well, as requested by @alextea.

PR #359, opened 2017-06-07
Update taxons_tagged_to_content.rake to output data

theKHutDeveloper:
Update taxons_tagged_to_content.rake to output data that does not require pivoting in Google Sheets

Removed fields that are not necessary.

### DIFF
--- a/lib/tasks/taxonomy/analysis/taxons_tagged_to_content.rake
+++ b/lib/tasks/taxonomy/analysis/taxons_tagged_to_content.rake
@@ -32,8 +32,6 @@ namespace :taxonomy do
           .join('; ')
 
         csv_row = [
-          organisations,
-          content_item['content_id'],
           content_item['link'],
           content_item['taxons'].length,
         ]
--- a/lib/tasks/taxonomy/analysis/taxons_tagged_to_content.rake
+++ b/lib/tasks/taxonomy/analysis/taxons_tagged_to_content.rake
@@ -3,7 +3,7 @@ require 'csv'
 
 namespace :taxonomy do
   namespace :analysis do
-    desc "Produces a CSV list of content items under a root taxon ID, and the taxons they are tagged to"
+    desc "Produces a CSV list of links under a root taxon ID, and the base paths they are tagged to"
     task :taxons_tagged_to_content, [:root_taxon_id] => [:environment] do |_, args|
       root_taxon_id = args.fetch(:root_taxon_id)
 
@@ -13,7 +13,7 @@ namespace :taxonomy do
       loop do
         results = GdsApi::Rummager.new(Plek.find('rummager')).search(
           filter_part_of_taxonomy_tree: root_taxon_id,
-          fields: %w(link taxons content_id organisations),
+          fields: %w(link taxons),
           start: page * results_per_page,
           count: results_per_page,
         )['results']
@@ -24,20 +24,33 @@ namespace :taxonomy do
         page += 1
       end
 
-      puts %w(organisations content_id base_path number_of_taxons).to_csv
+      content_items.each do |item|
+        details = []
+        item["taxons"].each do |taxon|
+          resp = Services.publishing_api.get_content(taxon)
+          details << resp["base_path"]
+        end
+        item["base_paths"] = details
+      end
+
+      puts %w(Link Number_of_taxons Base_path_1 Taxon_id_1 Base_path_2
+        Taxon_id_2 Base_path_3 Taxon_id_3).to_csv
 
       content_items.each do |content_item|
-        organisations = content_item.fetch('organisations', [])
-          .map { |org| org['title'] }
-          .join('; ')
-
         csv_row = [
           content_item['link'],
           content_item['taxons'].length,
         ]
 
-        content_item['taxons'].each { |taxon_id| csv_row << taxon_id }
+        max = [content_item['base_paths'].length, content_item['taxons'].length].max
 
+        for i in 0...max
+          content_item['base_paths'][i].nil? ? nil
+            : csv_row.push(content_item['base_paths'][i])
+
+          content_item['taxons'][i].nil? ? nil
+            : csv_row.push(content_item['taxons'][i])
+        end
         puts csv_row.to_csv
       end
     end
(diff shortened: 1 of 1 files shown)


theKHutDeveloper:
Trello: https://trello.com/c/7ONnsVc2/192-update-taxonstaggedtocontent-to-output-data-that-doesnt-require-pivoting-in-google-sheets

@carvil 1st part of the Trello card completed - removed unnecessary fields

lib/tasks/taxonomy/analysis/taxons_tagged_to_content.rake
```diff
@@ -24,20 +24,33 @@ namespace :taxonomy do
...
+      content_items.each do |content_item|
         csv_row = [
           content_item['link'],
           content_item['taxons'].length,
         ]
 
-        content_item['taxons'].each { |taxon_id| csv_row << taxon_id }
+        max = [content_item['base_paths'].length, content_item['taxons'].length].max
+
+        for i in 0...max
```
carvil: This is completely fine, but not widely used in Ruby. One nice method on `Integer` is [upto](https://ruby-doc.org/core-2.2.0/Integer.html#method-i-upto). In this case, you could write is as:

```
0.upto(max) do |i|
  ...
end
```

lib/tasks/taxonomy/analysis/taxons_tagged_to_content.rake
```diff
@@ -24,20 +24,33 @@ namespace :taxonomy do
...
+          content_item['base_paths'][i].nil? ? nil
+            : csv_row.push(content_item['base_paths'][i])
 
+          content_item['taxons'][i].nil? ? nil
```
carvil: Instead of looping through all taxons and base paths, we could use a few methods on `Array` to help us do this. Here is an example of how to use them:

```
[3] pry(main)> [1, 2, 3].zip(['a', 'b','c']).flatten
=> [1, "a", 2, "b", 3, "c"]
[4] pry(main)>
```

- [zip](https://ruby-doc.org/core-2.2.0/Array.html#method-i-zip)
- [flatten](https://ruby-doc.org/core-2.2.0/Array.html#method-i-flatten)

What do you think?

lib/tasks/taxonomy/analysis/taxons_tagged_to_content.rake
```diff
@@ -24,22 +24,33 @@ namespace :taxonomy do
         page += 1
       end
 
-      puts %w(organisations content_id base_path number_of_taxons).to_csv
+      content_items.each do |item|
```
alecgibson: I think this loop can be happily merged into the loop below, which also iterates over `content_items`? Then you don't have to actually mutate the `content_item` (which can be a little surprising).

You'd also then be able to directly use the `publishing_api` response when constructing your `csv_row`. For example:

```ruby
# Set up csv_row (omitted)
csv_row = [...]

# Fetch taxon data
taxons = item['taxons'].map do |taxon_id|
  Services.publishing_api.get_content(taxon_id)
end

# Reformat data for extra columns
taxon_columns = taxons.each_with_object([]) do |taxon, columns|
  columns << taxon['content_id']
  columns << taxon['base_path']
end

# Add extra columns
csv_row += taxon_columns
```

alecgibson: (If you're unfamiliar with [`Array#map`](https://ruby-doc.org/core-2.2.0/Array.html#method-i-map) and [`Enumerable#each_with_object`](https://ruby-doc.org/core-2.4.1/Enumerable.html#method-i-each_with_object), I'd highly recommend reading about them and experimenting a bit)

theKHutDeveloper: @alecgibson Thanks Alec, I don't know many of the methods for arrays or hashes. A lot of them seem to be interchangeable :/ Yeah a lot of reading will have to be done

alecgibson: Sure. Given that this would basically be a complete rewrite, it's fine to not make these changes, but do try to read around the extra methods you have available to you, because it should hopefully lead to cleaner code.

lib/tasks/taxonomy/analysis/taxons_tagged_to_content.rake
```diff
@@ -3,7 +3,7 @@ require 'csv'
 
 namespace :taxonomy do
   namespace :analysis do
-    desc "Produces a CSV list of content items under a root taxon ID, and the taxons they are tagged to"
+    desc "Produces a CSV list of links under a root taxon ID, and the base paths they are tagged to"
```
alecgibson: I think removing the word "taxons" from this description is a bit confusing - possibly "...the taxon base paths they are..." ?

lib/tasks/taxonomy/analysis/taxons_tagged_to_content.rake
```diff
@@ -24,21 +24,28 @@ namespace :taxonomy do
...
-          organisations,
-          content_item['content_id'],
           content_item['link'],
           content_item['taxons'].length,
         ]
 
-        content_item['taxons'].each { |taxon_id| csv_row << taxon_id }
+        taxon_details = content_item['base_paths'].zip(content_item['taxons']).flatten
+
+        taxon_details.each do |taxon_detail|
```
alecgibson: You could just write:

```ruby
csv_row += taxon_details
```

theKHutDeveloper: @alecgibson: Wouldn't this add taxon_details as an array? I want the details to appear in separate columns

alecgibson: Adding two arrays concatenates them, as you can see in the docs: https://docs.ruby-lang.org/en/2.0.0/Array.html#method-i-2B

theKHutDeveloper:
@alecgibson: Made the change you suggested

alecgibson:
@theKHutDeveloper looks good, although why did you change the description? "Link" is a bit of a tricky term for us (it can mean anything from a URL on a page to the links that exist between content items), and technically things are tagged to taxons, not to base paths (you could change the wording to taxon base path, if you find that clearer?).

theKHutDeveloper:
@alecgibson: I agreed the description with Carlos yesterday, but I'm happy to change it, as others may find it confusing.

theKHutDeveloper:
@alecgibson: Changed the description

alecgibson:
Looks good! Go ahead and merge 🚀

PR #364, opened 2017-06-13
Add associated taxons input form

davidbasalla:
For https://trello.com/c/fkKqW1rA/206-allow-taxons-to-be-associated-to-each-other-in-content-tagger

This PR adds the ability to add associated taxons to taxons. It's part of the work we're doing around world locations to be rendered as taxon pages. This PR adds a new input field and shows them on the show page (if present).

When deleting a taxon that is being associated from another taxon, the publishing api automatically updates to remove the taxon accordingly, so there's no need to cater for this explicitly in the ContentTagger app.

TODO

- [x]  Add explanation to README

-------------------------
# Screenshots

## Before - edit/new

screen shot 2017-06-13 at 14 37 22

## After - edit/new 
screen shot 2017-06-13 at 14 36 59

-------------------------

## Before - show
screen shot 2017-06-13 at 14 36 10

## After - show
screen shot 2017-06-13 at 14 36 33

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -425,6 +425,9 @@ SpecialGlobalVars:
   Description: 'Avoid Perl-style global variables.'
   Enabled: true
 
+Style/VariableNumber:
+  Enabled: false
+
 UnneededCapitalW:
   Description: 'Checks for %W when interpolation is not needed.'
   Enabled: true
--- a/README.md
+++ b/README.md
@@ -22,6 +22,12 @@ searches for a collection, if the search is too broad, we will only show the top
 results. Instead, we ask the user to be more specific in the search query in
 order to find the collection they are looking for.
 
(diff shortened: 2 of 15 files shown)


tijmenb:
This looks good, but I'm unsure of the `let` blocks. The lazy loading in this case is not necessary in these test cases, and in some cases we explicitly don't want it (hence the `let!`). Using ivars in this case guarantees us that the code is executed, and makes the tests a little less magic. I would prefer to keep the acceptance tests free from `let` statements.

davidbasalla:
Thanks @tijmenb, have removed that last commit with the `let` block change. I've had to disable a rubocop instead since it wasn't happy about the instance variable naming of `@taxon_3` (and others) 🙄

tijmenb:
@davidbasalla great!

PR #366, opened 2017-06-13
Add new columns to tagging events

tijmenb:
- `navigation_document_supertype` is the a supertype which is used to differentiate guidance and other content
- `base_path` is useful to link out to the tagged page

https://trello.com/c/GRPCKmbU


whoojemaflip:
Will the import script handle these new columns?

tijmenb:
Yup, it just uses the CSV column names:

https://github.com/alphagov/content-tagger/blob/master/lib/tasks/import_tagging_events.rake#L3-L5

PR #367, opened 2017-06-13
Bare bones taxonomy analytics page

whoojemaflip:
Adds a new page and top level navigation item, for the purposes of exposing the performance of the GOV.UK taxonomy. 

The page shows a list of terms, ordered by the amount of content currently tagged to them.

screen shot 2017-06-14 at 11 47 46

### DIFF
--- a/app/services/analytics/index_page.rb
+++ b/app/services/analytics/index_page.rb
@@ -1,0 +1,19 @@
+module Analytics
+  class IndexPage
+    attr_reader :params
+
+    def initialize(params)
+      @params = params
+    end
+
+    def taxons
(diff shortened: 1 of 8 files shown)


app/services/analytics/index_page.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+
+    def taxons
+      TaggingEvent.content_counts_by_taxon
+    end
+
+  private
+
+    def remote_taxons
+      @remote_taxons ||= RemoteTaxons.new
```
andrewgarner: Is this being used?

whoojemaflip: it was not

PR #377, opened 2017-06-15
Commonize sortable form element for taxon tagging

alecgibson:
We have two extremely similar sortable fields, used for tagging related
content to content items. We will soon be adding a third of these fields
(for external related links), so in preparation for this, this change
separates the repeated HTML from the content of the form elements, so
that it can be reused without copying and pasting.

### Trello

https://trello.com/c/42clJNoN/184-add-external-link-sidebar-overrides-for-the-new-navigation


alecgibson:
It turns out that external links cannot be curated in `content-tagger` without significant work, so we won't need this after all.

PR #381, opened 2017-06-23
Add task to create worldwide taxons

davidbasalla:
For https://trello.com/c/6S8yxQpD/248-automate-creation-of-worldwide-taxons

We need to create a lot of new taxons (229 countries * 10 sub taxons = 2290 taxons). It would take too long to create these by hand, so this rake tasks allows us to create them programmatically. We should probably remove this code once we've created the taxons.

TODO
- [x] Split out code into separate class/es
- [x] Make one call for taxons instead of search for each taxon in `country_parent_taxon`
- [x] Add content_ids for generic taxons
- [x] Check that taxons don't get published, only updated
- [x] Check that multiple taxons don't get created on multiple runs of the rake task


lib/tasks/taxonomy/create_worldwide_taxons.rake
```diff
@@ -0,0 +1,368 @@
...
+    end
+
+    def create_taxon(country, taxon)
+      {
+        path_prefix: "/world",
+        path_slug: slug(country, taxon),
+        internal_name: internal_name(country, taxon),
+        title: title(country, taxon),
+        description: taxon[:description],
+        visible_to_departmental_editors: true, # find out what this is used for
```
tijmenb: This determines whether or not the taxon is visible in Whitehall when it's in draft state. `true` is fine here.

PR #382, opened 2017-06-23
Add a task to export a taxonomy to a JSON array

alecgibson:
The navigation prototype needs to display the Parenting & Childcare
taxonomy, but this taxonomy is still only available in the draft
content store. Since the draft content store is in a VPN that the
prototype cannot access, we need to manually export the taxonomy so
that we can manually mock out a content store response.

### Trello

https://trello.com/c/i3YMQLaT/242-spike-download-draft-taxons-from-draft-content-store-and-serve-them-in-the-prototype

### DIFF
--- a/lib/tasks/taxonomy/export_taxonomy_to_json.rake
+++ b/lib/tasks/taxonomy/export_taxonomy_to_json.rake
@@ -1,0 +1,59 @@
+require 'json'
+
+namespace :taxonomy do
+  desc <<-DESC
+    Exports an expanded taxonomy to a single JSON array
+  DESC
+  namespace :export do
+    task :json, [:root_taxon_id] => [:environment] do |_, args|
+      root_taxon_id = args.fetch(:root_taxon_id)
+      root_taxon = OpenStruct.new(Services.publishing_api.get_content(root_taxon_id).to_h)
+      taxonomy = Taxonomy::ExpandedTaxonomy.new(root_taxon.content_id)
+      taxonomy.build
+
+      flattened_taxonomy = flatten_taxonomy(taxonomy.child_expansion)
+
+      puts JSON.generate(flattened_taxonomy)
+    end
+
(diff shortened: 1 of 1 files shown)


lib/tasks/taxonomy/export_taxonomy_to_json.rake
```diff
@@ -0,0 +1,59 @@
+require 'json'
+
+namespace :taxonomy do
+  desc <<-DESC
+    Exports an expanded taxonomy to a single JSON array
+  DESC
+  namespace :export do
+    task :json, [:root_taxon_id] => [:environment] do |_, args|
+      root_taxon_id = args.fetch(:root_taxon_id)
+      root_taxon = OpenStruct.new Services.publishing_api.get_content(root_taxon_id).to_h
```
carvil: Could you please add `()` around the parameter for consistency?

lib/tasks/taxonomy/export_taxonomy_to_json.rake
```diff
@@ -0,0 +1,59 @@
...
+      if taxon.children
+        taxon.children.each do |child_taxon|
+          flattened_taxonomy += flatten_taxonomy(child_taxon)
+        end
+      end
+
+      flattened_taxonomy
+    end
+
+    def convert_to_content_item(taxon, recursion_direction = nil)
```
carvil: The service [BuildTaxon](https://github.com/alphagov/content-tagger/blob/master/app/services/taxonomy/build_taxon.rb) builds a taxon object from a payload. Would it make sense to have a `to_h` in the class `Taxon` method that would return the content item hash and call to it instead?

alecgibson: As discussed, adding a `to_h` on the `Taxon` model would be more work than it's worth.

PR #395, opened 2017-07-03
Clean up some code

tijmenb:
This PR contains a few acts of refactoring and cleanup. See individual commits for details.


app/models/application_record.rb
```diff
@@ -1,3 +0,0 @@
-class ApplicationRecord < ActiveRecord::Base
```
whoojemaflip: on the other hand, this is where Rails is at now. Don't fight the tide?

tijmenb: Some tides are worth fighting.

https://en.wikipedia.org/wiki/Delta_Works

whoojemaflip: good answer

PR #397, opened 2017-07-17
Database config use database url in dev

cbaines:


tijmenb:
What's not working with the current set up?

cbaines:
> What's not working with the current set up?

For me, it's getting content-tagger to use the port and database name that I'd like it to on my laptop. This patch solves that problem for me, but I thought I'd suggest changing the configuration, as that avoids it inconveniencing someone else in the future.

PR #406, opened 2017-07-24
Import missing tagged content

whoojemaflip:
Extracts related_topics for a Three Level Taxonomy page

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,5 +2,5 @@
 
 node {
   def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
-  govuk.buildProject()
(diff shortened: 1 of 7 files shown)


Jenkinsfile
```diff
@@ -2,5 +2,5 @@
 
 node {
   def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
-  govuk.buildProject()
+  govuk.buildProject(sassLint: false)
```
tijmenb: What's the reason for removing sass linting?

whoojemaflip: It was enabled as a new default option in CI, but failed catastrophically when it ran against this codebase. Think there was an error in selecting for vendored content, and that may have been fixed now.

PR #413, opened 2017-07-25
Add Bulk publish taxons button

koetsier:
Adds a 'publish tree' button on the view taxon page of any draft taxon. When pressed, this taxon and all its children will be set to published. A sidekiq job per taxon is used to publish the taxons.

https://trello.com/c/cGImL1qz/47-a-way-to-publish-taxons-in-bulk

screen shot 2017-07-25 at 12 01 59
screen shot 2017-07-25 at 14 34 03
screen shot 2017-07-25 at 14 35 04

### DIFF
--- a/app/services/taxonomy/bulk_update_taxon.rb
+++ b/app/services/taxonomy/bulk_update_taxon.rb
@@ -1,0 +1,19 @@
+module Taxonomy
+  class BulkUpdateTaxon
+    def initialize(root_taxon_content_id)
+      @root_taxon_content_id = root_taxon_content_id
+    end
+
+    def self.call(root_taxon_content_id)
+      new(root_taxon_content_id).bulk_publish
+    end
+
+    def bulk_publish
+      linked_content_item = GovukTaxonomyHelpers::LinkedContentItem.from_content_id(content_id: @root_taxon_content_id,
+                                                                                    publishing_api: Services.publishing_api)
+      linked_content_item.each do |taxon|
+        PublishTaxonWorker.perform_async(taxon.content_id)
+      end
+    end
+  end
+end
--- a/app/workers/publish_taxon_worker.rb
+++ b/app/workers/publish_taxon_worker.rb
@@ -1,0 +1,9 @@
+class PublishTaxonWorker
+  include Sidekiq::Worker
+
+  def perform(taxon_content_id)
+    Services.publishing_api.publish(taxon_content_id)
+  rescue GdsApi::HTTPConflict => ex # Ignore attempts to publish already published content
+    puts "409 #{ex.message}"
+  end
+end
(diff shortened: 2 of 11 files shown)


app/workers/publish_taxon_worker.rb
```diff
@@ -0,0 +1,10 @@
+class PublishTaxonWorker
+  include Sidekiq::Worker
+
+  def perform(taxon_content_id)
+    Services.publishing_api.publish(taxon_content_id)
+  rescue GdsApi::HTTPClientError => ex
```
whoojemaflip: There's probably a more specific error class you can `rescue` from here.

whoojemaflip: Edit: Surprised to see there isn't, so I raised [a PR](https://github.com/alphagov/publishing-api/pull/980) against the `publishing-api`. Feel free to add to that thread if you have strong feelings about the HTTP semantics of this response.

app/models/taxon.rb
```diff
@@ -19,7 +19,7 @@ class Taxon
 
   validates_presence_of :title, :description, :internal_name, :path_prefix
   validates :path_prefix, inclusion: { in: Theme.taxon_path_prefixes }
-  validates :path_slug, allow_blank: true, format: { with: %r{\A/[a-zA-Z0-9\-]+\z} }
+  validates :path_slug, allow_blank: true, format: { with: %r{\A/[a-zA-Z0-9\-\/]+\z} }
```
whoojemaflip: This could really do with being in it's own commit, so there's a historical record of why this change was made.

app/workers/publish_taxon_worker.rb
koetsier: changed it so it now catches 409 errors instead of 400

app/models/taxon.rb
koetsier: done

app/workers/publish_taxon_worker.rb
whoojemaflip: uh, you could just rescue the `GdsApi::HTTPConflict` error now. No need to inspect the contents of the error message.

app/models/taxon.rb
whoojemaflip: Could you explain why this needed to be done in the commit message

app/services/taxonomy/bulk_update_taxon.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    def bulk_publish
+      linked_content_item = LinkedContentItemApi.linked_content_item(@root_taxon_content_id)
+      linked_content_item.each do |taxon|
+        PublishTaxonWorker.perform_async(taxon.content_id)
+      end
+    end
+
+    class LinkedContentItemApi
```
whoojemaflip: is there any need for this to be contained in it's own class? I feel this obscures the intention and makes this code harder to read.

For test purposes, wouldn't it be just as easy to stub the response from `GovukTaxonomyHelpers::LinkedContentItem.from_publishing_api` to return your test doubles?

koetsier: moved the method in LinkedContentItemApi to govuk_taxonomy_helpers

PR #417, opened 2017-07-31
Move taxonomy publishing to Sidekiq for retry

whoojemaflip:
Publishing the legacy taxonomy prototype would error out and fail to complete almost every time it was run, due to random API timeouts and unreproducible errors.

This moves all of the API based operations required for publishing into separate workers, so that Sidekiq can manage the error handling and retries.

Each of the jobs should be idempotent, so that it can fail at any point in the process and be retried without corrupting data.


app/workers/legacy_taxonomy/taxonomy_publisher.rb
```diff
@@ -0,0 +1,31 @@
...
+      content_id = taxon_data.content_id
+
+      Client::PublishingApi.put_content(content_id, taxon_for_publishing_api(taxon_data))
+      Client::PublishingApi.publish(content_id)
+
+      if parent_taxon_id
+        Client::PublishingApi.patch_links(content_id, links: { parent_taxons: [parent_taxon_id] })
+      end
+
+      taxon_data.tagged_pages.each do |taggable|
```
koetsier: could potentially filter first on taggable['content_id']

whoojemaflip: :+1:

PR #420, opened 2017-08-07
Force the use of generated expanded links to avoid stale results

whoojemaflip:
Recent changes to publishing-api.expanded-links mean that the
tagging interface will return an error, due to stale data.

Resulting in this error message when trying to change the tags assigned:

image

This change should ensure that the tagging interface will always
receive the freshest data.

[Trello](https://trello.com/c/sOMPe5Od/142-error-while-tagging-content-in-content-tagger)


tijmenb:
This will now work because https://github.com/alphagov/publishing-api/pull/989 adds the `version` attribute back in the payload when `generate=true`.

PR #425, opened 2017-08-10
Tagging projects add basic search

cbaines:
search-demo

This currently uses a very crude query, which probably needs changing to a proper full text search ranking with a rank threshold.

https://trello.com/c/FgwahQeG/103-filter-table-with-search

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -4,7 +4,8 @@ class ProjectsController < ApplicationController
   end
 
   def show
-    render :show, locals: { project: project }
+    render :show, locals: { project: project,
+                            project_content_items_to_display: project_content_items_to_display }
   end
 
   def new
@@ -30,9 +31,22 @@ private
     @_project ||= Project.find(params[:id])
   end
 
+  def project_content_items_to_display
+    @_project_content_items_to_display = begin
+      items = project.content_items.uncompleted
+      items = items.matching_search(search_query) if search_query
+
+      items
+    end
+  end
+
   def new_project_params
     params
       .fetch(:new_project_form)
       .permit(:name, :remote_url, :taxonomy_branch)
+  end
+
+  def search_query
+    params[:query]
   end
 end
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -42,15 +42,14 @@ private
       items = project.content_items
 
       tagged_state_filter = filter_params[:tagged_state]
-      if tagged_state_filter && tagged_state_filter != "all"
-        done = { tagged: true, not_tagged: false }[tagged_state_filter.to_sym]
 
-        if done.nil?
+      if tagged_state_filter && tagged_state_filter != TAGGED_STATE_ALL
+        unless TAGGED_STATES.include? tagged_state_filter
           raise ActionController::BadRequest,
                 "The value \"#{tagged_state_filter}\" is an invalid tagging state."
         end
 
-        items = items.where(done: done)
+        items = items.where(done: tagged_state_filter == TAGGED_STATE_TAGGED)
       end
 
       query = filter_params[:query]
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -33,8 +33,22 @@ private
 
   def project_content_items_to_display
     @_project_content_items_to_display = begin
-      items = project.content_items.uncompleted
-      items = items.matching_search(search_query) if search_query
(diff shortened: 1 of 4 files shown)


app/controllers/projects_controller.rb
```diff
@@ -30,9 +31,36 @@ def project
     @_project ||= Project.find(params[:id])
   end
 
+  def project_content_items_to_display
+    @_project_content_items_to_display = begin
+      items = project.content_items
+
+      tagged_state_filter = filter_params[:taggedState]
```
koetsier: is this supposed to be camelcase?

app/models/project_content_item.rb
```diff
@@ -19,4 +19,8 @@ def proxied_url
   end
 
   scope :uncompleted, -> { where(done: false) }
+
+  def self.matching_search(query)
```
koetsier: could this be a scope?

app/controllers/projects_controller.rb
```diff
@@ -30,9 +31,36 @@ def project
     @_project ||= Project.find(params[:id])
   end
 
+  def project_content_items_to_display
```
koetsier: this bit might need a unit test

app/views/projects/show.html.erb
```diff
@@ -1,39 +1,113 @@
...
+             value="<%= params[:query] %>"
+             placeholder="Filter by title...">
+
+      <div class="radio">
+        <label>
+          <input type="radio"
+                 name="taggedState"
+                 id="taggedStateTagged"
+                 value="tagged"
+                 <% if params[:taggedState] == "tagged" %>checked<% end %>
```
koetsier: could set the tagged/not_tagged/all values in a constant somewhere and reference them

app/controllers/projects_controller.rb
cbaines: Yes, but it only ended up that way as I think I'm more use to using that style for ids on the DOM nodes. I've changed this to use underscores instead, as I think this would be more consistent with the Ruby code, which is what I'm guessing is the reason behind your question.

app/models/project_content_item.rb
cbaines: As far as I understand scopes, which is not very well at all, I think the scope method is equivalent? Assuming this was a suggestion, and not a question, I've gone ahead and changed it to a scope, as that is shorter, and more consistent with the existing scope.

app/controllers/projects_controller.rb
```diff
@@ -30,9 +31,22 @@ def project
...
+    @_project_content_items_to_display = begin
+      items = project.content_items.uncompleted
+      items = items.matching_search(search_query) if search_query
```
whoojemaflip: would this be better expressed as an `if..else` or am I missing something?

app/views/projects/show.html.erb
```diff
@@ -52,6 +53,24 @@
...
     });
+
+    var contentItemTitlesToFormElements = {}
+    $(".edit_project_content_item").each(function(index, formElement) {
+      var $formElement = $(formElement);
+      var titleAnchor = $formElement.find("a")[0];
+      contentItemTitlesToFormElements[titleAnchor.innerHTML.toLowerCase()] = $formElement;
+    });
+
+    $("#project-tagging-content-item-search").keyup(function(event) {
```
whoojemaflip: might it be worth debouncing this function, so it only fires off an http request once the user has stopped typing?

cbaines: I don't believe the callback here does any HTTP stuff?

cbaines: It just filters the data already on the page, hiding any elements that don't match the search, and showing the rest.

app/controllers/projects_controller.rb
cbaines: What do you think might be better using an `if..else`?

app/views/projects/show.html.erb
whoojemaflip: totes.. sorry, misread.

app/controllers/projects_controller.rb
whoojemaflip: this was rewritten in a later commit. nm.

app/controllers/projects_controller.rb
cbaines: I've now extended the feature tests in 736b401

app/views/projects/show.html.erb
cbaines: I've done this in 6b3725c and 794ceb8.

PR #427, opened 2017-08-14
Tagathon Bulk Tagging UI prototype

whoojemaflip:
bulk-tagging

[trello](https://trello.com/c/cE05oHPa/132-tag-in-bulk)

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -8,5 +8,5 @@
 //= require_self
 
 $(document).ready(function() {
-  $(".select2:not(.tagging_project)").select2({ allowClear: true });
+  $(".select2:not(.tagging_project):not(.bulk_tagger)").select2({ allowClear: true });
 });
--- a/app/services/projects/bulk_tagger.rb
+++ b/app/services/projects/bulk_tagger.rb
@@ -1,0 +1,22 @@
+module Projects
+  class BulkTagger
+    def initialize(content_items:, taxons:)
+      @content_ids = content_items
+      @taxon_content_ids = taxons
+    end
+
+    def commit
+      @content_ids.each do |id|
+        content_item = ProjectContentItem.find(id)
+        TagContentWorker.perform_async(content_item.content_id, @taxon_content_ids)
+        content_item.mark_complete
+      end
+    end
+
+    def result
+      @content_ids.reduce([]) do |acc, content_id|
+        acc << { content_id: content_id, taxons: @taxon_content_ids }
+      end
+    end
+  end
+end
--- a/app/views/projects/show.html.erb
+++ b/app/views/projects/show.html.erb
@@ -3,7 +3,7 @@
 <div class="row tagathon-project">
   <h1><%= project.name %></h1>
 
-  <div class="col-md-3">
+  <div class="col-md-3 filter-controls">
     <h3>Filter</h3>
 
     <form method="GET">
--- a/app/views/projects/show.html.erb
+++ b/app/views/projects/show.html.erb
@@ -86,6 +86,23 @@
       });
     </script>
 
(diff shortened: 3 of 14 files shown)


app/assets/javascripts/application.js
```diff
@@ -6,8 +6,11 @@
 //= require Chart.bundle
 //= require chartkick
 //= require vendor/datatables.min
+
+// This solves one problem and creates another one.
+// TODO: 🤔
 //= require jquery_ujs
```
whoojemaflip: there is a conflict in the javascript dependancies 😢

app/assets/javascripts/application.js
```diff
@@ -6,8 +6,11 @@
...
 
 $(document).ready(function() {
-  $(".select2:not(.tagging_project)").select2({ allowClear: true });
+  $(".select2:not(.tagging_project):not(.bulk_tagger)").select2({ allowClear: true });
```
whoojemaflip: this is gross, and we should maybe look at how we use `select2` across the app.

app/services/projects/bulk_tagger.rb
```diff
@@ -0,0 +1,37 @@
+module Projects
+  class BulkTagger
+    def initialize(content_items:, taxons:)
+      @content_ids = content_items
+      @taxon_content_ids = taxons
+    end
+
+    def commit
+      @content_ids.each do |id|
```
koetsier: could this use some error-handling, say for instance `content_item_id_lookup` fails to return something sensible.

app/views/projects/show.html.erb
```diff
@@ -12,8 +12,25 @@
       .html_safe
   %>;
 
+  var taxon_lookup = <%=
```
koetsier: does it make sense to move this javascript out of this view and into one of these govuk_frontend_toolkit modules in assets/javascript?

spec/services/projects/bulk_tagger_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+      result.each do |r|
+        expect(r[:content_id]).to be_in content_items
+        expect(r[:taxons]).to eql taxons
+      end
+    end
+  end
+
+  def array_of_three_uuids
+    [SecureRandom.uuid, SecureRandom.uuid, SecureRandom.uuid]
```
koetsier: You could save loads of characters by using `Array.new(3) {SecureRandom.uuid}`

app/views/projects/show.html.erb
```diff
@@ -23,17 +40,78 @@
...
+        $(this).removeClass('error-saving');
+        $(this).effect("highlight", {color: "#d8f3d8"}, 1000);
+      }
+    ).on(
+      'ajax:error',
+      function(event, data, status, xhr) {
+        $(this).addClass('error-saving');
+      }
+    ).each(function(index, el) {
+      ref = $(el).attr('data-ref');
```
koetsier: could you write a comment/move to a separate function so that the intent of this code is a bit clearer.

app/views/projects/show.html.erb
whoojemaflip: I'm not sure that this view-specific code can move to a module, but all the jQuery below needs to be refactored into one for sure.

app/assets/javascripts/application.js
whoojemaflip: resolved in a separate PR

app/services/projects/bulk_tagger.rb
whoojemaflip: this has changed since you last reviewed it. The content_id lookup takes place at the time that a project is created, for one. Also now the actual tagging is handled by a SideKiq worker which gives us free retries in the event of an HTTP error.

PR #432, opened 2017-08-21
Improve the "New project" page

tijmenb:
Add some clues on how to upload the spreadsheet.

## Before

screen shot 2017-08-21 at 17 30 15

## After

screen shot 2017-08-21 at 17 30 10

### DIFF
--- a/app/validators/google_url_validator.rb
+++ b/app/validators/google_url_validator.rb
@@ -27,7 +27,8 @@ private
       record.errors[:url] << I18n.t('errors.missing_gid')
     end
 
-    unless parameters['output'].present? && parameters['output'].include?('tsv')
+    # rubocop:disable Style/GuardClause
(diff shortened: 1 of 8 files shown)


whoojemaflip:
The advice suggests that we accept a TSV file, but I don't think the CSV parser is actually configured for tab-separated.

tijmenb:
@whoojemaflip I've changed where we use TSV to CSV to be consistent.

app/validators/google_url_validator.rb
```diff
@@ -27,7 +27,8 @@ def validate_parameters(uri, record)
       record.errors[:url] << I18n.t('errors.missing_gid')
     end
 
-    unless parameters['output'].present? && parameters['output'].include?('tsv')
+    # rubocop:disable Style/GuardClause
```
whoojemaflip: :robot: :rofl:

PR #433, opened 2017-08-21
Refactor the search form (and fix a small issue)

tijmenb:
This refactors the search controller code for the tagathon projects. Also makes sure that when the page is loaded "All" is selected.

## Before

screen shot 2017-08-21 at 17 52 16

## After

screen shot 2017-08-21 at 17 52 10


whoojemaflip:
How do you feel about writing a simple feature for this? It would be good to see it at least executed in CI.

PR #435, opened 2017-08-22
Permissions

whoojemaflip:
Introduces the concept of Permissions into the content-tagger application.

Up until now any user with the default `signin` permission granted by `Signon` could access all the features. With the coming rollout of the tool to non-administrative roles there is a need to restrict certain aspects of the application to only those with the responsibility for making changes to the GOV.UK navigation structure. The `signin` permission will no longer grant access to any of the functionality of content-tagger.

This PR introduces two new roles: GDS Editor, and Tagathon Participant.  The roles have been created in Signon already, however not all users will have been assigned them. This means that there will initially be some disruption when this is deployed, unless we can preemptively assign the correct permissions.

[Trello](https://trello.com/c/jWJa4rEi/176-permissions-in-content-tagger)

### DIFF
--- a/spec/support/authentication_helpers.rb
+++ b/spec/support/authentication_helpers.rb
@@ -1,0 +1,48 @@
+module AuthenticationControllerHelpers
+  def login_as(user)
+    request.env['warden'] = double(
+      authenticate!: true,
+      authenticated?: true,
+      user: user
+    )
+  end
+
+  def stub_user
+    @stub_user ||= FactoryGirl.create(:user, :gds_editor)
+  end
+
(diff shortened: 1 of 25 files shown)


spec/support/authentication_helpers.rb
```diff
@@ -0,0 +1,48 @@
+module AuthenticationControllerHelpers
```
tijmenb: Does `GDS::SSO` provide some of these methods?

whoojemaflip: I don't think that GDS::SSO contains any test helpers?

PR #440, opened 2017-08-29
Include current tags in tagathon tool

whoojemaflip:
Uses the new `/v2/links/by-content-id` endpoint in `publishing-api` to augment the page with previously assigned tags. This endpoint is capable of supplying links for up to 1000 content items at a time.

[Trello](https://trello.com/c/H3BqIPWc/190-current-tags-are-visible-in-the-bulk-tagging-tool)

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -7,8 +7,9 @@ class ProjectsController < ApplicationController
 
   def show
     render :show, locals: { project: project,
-                            bulk_search: Projects::BulkSearch.new(project, params),
-                            taxons: taxons_json }
+                            bulk_search: searcher,
+                            taxons: taxons_json,
+                            content_items: content_items }
   end
 
   def new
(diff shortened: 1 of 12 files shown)


app/controllers/projects_controller.rb
```diff
@@ -24,6 +25,30 @@ def create
...
+
+  def content_items
+    searcher.project_content_items_to_display.map do |content_item|
+      content_item.taxons = taxons_for(content_item)
+      content_item
+    end
+  end
+
+  def taxons_for(content_item)
+    result = content_tags.find { |hsh| hsh['content_id'] == content_item.content_id }
```
koetsier: could potentially use `content_tags.find(->{[]}) { |hsh...` to give it a default value

whoojemaflip: I'm going to replace the publishing-api endpoint so that it returns a hash of content-id => links to make this faster and easier

PR #444, opened 2017-09-05
Hide bulk tagger

whoojemaflip:
Provides an option on the New Project screen, to enable the bulk-tagging UI.

Bulk-tagging is disabled by default for existing projects, and for new projects unless it is explicitly enabled.

[Trello](https://trello.com/c/ZzG433q5/220-hide-bulk-tagging-and-checkboxes)

### DIFF
--- a/spec/lib/project_builder_spec.rb
+++ b/spec/lib/project_builder_spec.rb
@@ -5,20 +5,40 @@ RSpec.describe ProjectBuilder do
     allow(LookupContentIdWorker).to receive(:perform_async)
   end
 
+  let(:project_name) { 'project_name' }
+  let(:taxonomy_branch_content_id) { SecureRandom.uuid }
+  let(:content_item_attributes_enum) { [] }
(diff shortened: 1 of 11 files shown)


spec/lib/project_builder_spec.rb
```diff
@@ -5,20 +5,40 @@
     allow(LookupContentIdWorker).to receive(:perform_async)
   end
 
+  let(:project_name) { 'project_name' }
+  let(:taxonomy_branch_content_id) { SecureRandom.uuid }
+  let(:content_item_attributes_enum) { [] }
+  let(:bulk_tagging_enabled) { false }
+
+  def subject(
```
tijmenb: Would it make sense to name this `create_project` instead of `subject`?

Then:

```rb
expect { subject }.to change { Project.count }.by(1)
```

Becomes self-explanatory as:

```rb
expect { create_project }.to change { Project.count }.by(1)
```

PR #448, opened 2017-09-06
Redesign the management of root taxons

cbaines:
https://trello.com/c/AHfdTD94/188-create-visualisations-to-show-how-much-is-tagged-to-where

Move the previous index page for root taxons to be an explicit edit
page, and change the index page to a show page with a table. This show
page is now used to link out to the pages for the individual root
taxons.

This will allow users to navigate to the show pages for root taxons.

root-taxons-1
root-taxons-2


cbaines:
@tijmenb I've now restructured the routes, controller and views around a plural resource (rather than a singular one). Let me know what you think?

PR #451, opened 2017-09-07
Display taxon path in typeahead

whoojemaflip:
Allows the user to typeahead search through the entire taxonomy tree when selecting tags for an item of content in the Project tagger view.

typeahead

[Trello](https://trello.com/c/GwcQBQu5/249-display-taxon-path-in-typeahead-results)


alextea:
That looks great 👌

PR #453, opened 2017-09-07
Ensure administrative access for developers

whoojemaflip:
Add some documentation about permissions, to help developers avoid being stymied by the **Permission Denied** screen.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,22 @@ This is a Ruby on Rails application.
 It interacts with the publishing-api to manage the "links hash" for content on GOV.UK.
 It also allows the creation of new taxons, updating taxons and bulk-tagging.
 
+### Permissions
+
+Users need to have either the `GDS Editor` or the `Tagathon participant` permission to access
+this application.
+
+ - **GDS Editor** users will have full unrestricted access to the GOV.UK taxonomy and navigation tools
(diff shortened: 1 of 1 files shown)


tijmenb:
I'm a bit unsure about this, because it prevents developers from using the app as a real user.

As a low-fi alternative, maybe we can show a specific error message in development that explains how to update the users' permissions? I always use `User.first.update permissions: ["signin", "GDS Editor"]`

whoojemaflip:
@tijmenb How easy is it to edit the **Permission Denied** page from a client app?

tijmenb:
Add this to the controller and/or override the template:

https://github.com/alphagov/gds-sso/blob/master/lib/gds-sso/controller_methods.rb#L8-L10

tomsabin:
I ran into this problem yesterday when I got the application up and running for the first time, thankfully @whoojemaflip pointed me to this PR otherwise I probably would have fell into a rabbit hole. 

If we didn't want to change the application just for the development environment, perhaps we could explain the user permissions and include `User.first.update permissions: ["signin", "GDS Editor"]` in the "Running the application" section of the README?

whoojemaflip:
@tijmenb @tomsabin I've removed the more far-out commits, so this is just a documentation change now.

PR #455, opened 2017-09-12
Set flags for content

whoojemaflip:
First in a the series of changes that introduce the concept of Flagging to the Tagathon application.

Flagging is implemented as orthogonal to the 'doneness' of an item of content.

image

image

image

[Trello](https://trello.com/c/FWCp5ShV/274-set-flags-for-content)

### DIFF
--- a/app/assets/stylesheets/tagathon_project.scss
+++ b/app/assets/stylesheets/tagathon_project.scss
@@ -58,3 +58,10 @@
     }
   }
 }
+
+.flag-for-review {
+  .add-left-border {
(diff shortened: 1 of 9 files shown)


alextea:
Is the cancel button on the left of the continue button elsewhere?

I think it should be to the right.

cbaines:
> Is the cancel button on the left of the continue button elsewhere?

On the bulk tagging confirmation page, it looks like the Cancel is to the right.

whoojemaflip:
Button positioning and style updated

PR #456, opened 2017-09-12
Flag status in table view

whoojemaflip:
image

[Trello](https://trello.com/c/3REg1jaX/269-display-flag-status-in-table-view)


alextea:
For the flags can we use Bootstrap's existing label styles: https://getbootstrap.com/docs/3.3/components/#labels

PR #457, opened 2017-09-13
Remove a flag

whoojemaflip:
Remove a previously added flag from project content.

image

image

image

[Trello](https://trello.com/c/hBX0nlAk/267-remove-a-flag)

### DIFF
--- a/app/views/project_content_items/_add_flags.html.erb
+++ b/app/views/project_content_items/_add_flags.html.erb
@@ -1,0 +1,30 @@
+<div class='row flag-for-review'>
+  <h1>Flag '<%= content_item.title %>' for review</h1>
+
+  <%= form_for content_item, url: update_flags_project_content_item_path(project, content_item), method: :post, html: { class: 'col-md-6', data: { module: 'radio-toggle' } } do |f| %>
+
+    <div class='radio'>
+      <label>
+        <%= f.radio_button :flag, "needs_help" %>
+        I need help tagging this
+      </label>
+    </div>
+
+    <div class='radio'>
+      <label>
+        <%= f.radio_button :flag, "missing_topic", data: { target: 'suggested-tags' } %>
+        There's no relevant topic for this
+      </label>
+    </div>
+
+    <div id='suggested-tags' class='form-group add-top-margin add-left-border'>
+      <%= f.label :suggested_tags, "Suggest a new topic" %>
+      <%= f.text_field :suggested_tags, class: 'form-control input-md-4' %>
+    </div>
+
+    <div class='add-top-margin'>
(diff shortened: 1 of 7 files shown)


app/views/project_content_items/_add_flags.html.erb
```diff
@@ -0,0 +1,32 @@
...
+  <%= form_for content_item, url: update_flags_project_content_item_path(project, content_item), method: :post, html: { class: 'col-md-6', data: { module: 'radio-toggle' } } do |f| %>
+
+    <div class='radio'>
+      <label>
+        <%= f.radio_button :flag, "needs_help" %>
+        I need help tagging this
+      </label>
+    </div>
+
+    <div class='radio-toggle'>
```
tomsabin: This class seems redundant? I can't seem to find any CSS for it and the [RadioToggle module](https://github.com/alphagov/govuk_admin_template/blob/master/app/assets/javascripts/govuk-admin-template/modules/radio_toggle.js) doesn't use this class. Was it added just to group the two child `div`s or signify that this is where the JS module is being used?

app/views/project_content_items/_add_flags.html.erb
```diff
@@ -0,0 +1,32 @@
...
+      <div class='radio'>
+        <label>
+          <%= f.radio_button :flag, "missing_topic", data: { target: 'suggested-tags' } %>
+          There's no relevant topic for this
+        </label>
+      </div>
+
+      <div id='suggested-tags' class='form-group add-top-margin add-left-border'>
+        <%= f.label :suggested_tags, "Suggest a new topic" %>
+        <%= f.text_field :suggested_tags, class: 'form-control input-md-4' %>
```
tomsabin: Perhaps this is an unlikely use case but the "Suggested topic" isn't being removed, only hidden: 

If you were to "Flag for review" a content item and suggest a new topic, resolve that flag and then flag it again, the previously suggested topic would still be prefilled in the text input field. It's probably an unlikely scenario, but it poses more a question from a data perspective, as it might have a undesirable effect on what data we have left behind, as the flag isn't being deleted.

Was that intentional or have I misunderstood something along the way?

app/views/project_content_items/_add_flags.html.erb
whoojemaflip: yep, this is entirely vestigal I think. good :eyes:

app/views/project_content_items/_add_flags.html.erb
whoojemaflip: I'm OK with postponing the decision until we've got a better picture of what we're doing with the data :)

tomsabin:
This looks good to me, but I may be missing something without the domain knowledge. Perhaps somebody else could take a look and give the final thumbs up?

PR #461, opened 2017-09-18
Display tagging history

cbaines:
Adds a couple of new pages that expose the tagging history either recently for all taxons, or for a specific taxon.

tagging_history_index

tagging_history_page

Trello: https://trello.com/c/woqZQI8n/253-display-tagging-history-in-content-tagger


cbaines:
There is still a bit of work remaining on this. I want to do a little more refactoring, including moving the Publishing API endpoint handling out in to the gds-api-adapters library.

app/lib/services.rb
```diff
@@ -10,6 +10,14 @@ def self.publishing_api
     )
   end
 
+  def self.link_changes_api
```
koetsier: could add a comment that we're moving this

app/services/tagging_history/index_page.rb
```diff
@@ -0,0 +1,35 @@
...
+        @filter_by_user_options[user.name] = user.uid
+        result[:organisation] = user.organisation_slug.try do |slug|
+          slug.capitalize.tr('-', ' ')
+        end
+      end
+
+      # The Publishing API returns the results ordered so that the
+      # latest come last. Reverse this so that the latest events are
+      # displayed first.
+      results.reverse
```
koetsier: you could also iterate using reverse_each instead of each, which saves making a copy of the array and may be slightly more efficient

app/services/tagging_history/index_page.rb
```diff
@@ -0,0 +1,35 @@
+module TaggingHistory
+  class IndexPage
+    attr_reader :link_changes, :filter_by_user_options
+
+    def initialize(params = { link_types: ['taxons'] })
+      @filter_by_user_options = {}
```
koetsier: It may be clearer to replace this with a method that goes through the `@link_changes` and builds up the list of user options instead of doing everything in the `get_link_changes` method.

spec/features/tagging_history_spec.rb
```diff
@@ -0,0 +1,97 @@
+require "rails_helper"
```
koetsier: need some tests for the 'show' method of the controller

app/controllers/tagging_history_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+               {
+                 link_types: ['taxons'],
+                 target_content_ids: [content_item.content_id],
+               }.merge(filter_params)
+             )
+           }
+  end
+
+  def filter_params
+    params.permit(users: []).to_h.symbolize_keys
```
koetsier: you could reverse the merge in line 15, so you don't have to symbolize the keys

app/controllers/tagging_history_controller.rb
```diff
@@ -0,0 +1,26 @@
+class TaggingHistoryController < ApplicationController
+  before_action :ensure_user_can_access_tagathon_tools!
+
+  def index
+    render :index, locals: { page: TaggingHistory::IndexPage.new({ link_types: ['taxons'] }.merge(filter_params)) }
```
koetsier: link_types: ['taxons'] could be merged in the IndexPage itself, so this bit becomes simpler

spec/features/tagging_history_spec.rb
cbaines: I've added some tests in d1f6f9f

cbaines: Make that 1 test.

app/services/tagging_history/index_page.rb
cbaines: I've now done this as part of refactoring this class in f93598b

app/services/tagging_history/index_page.rb
cbaines: I put this in before reversing the order that the publishing-api was returning the results in, so I've removed it now as its unnecessary.

app/controllers/tagging_history_controller.rb
cbaines: I've now done this as part of refactoring this class in f93598b

app/controllers/tagging_history_controller.rb
cbaines: I did this, but I found I still needed to symbolize the keys.

app/services/tagging_history/link_changes.rb
```diff
@@ -0,0 +1,45 @@
...
+        end
+
+        change
+      end
+    end
+
+    def filter_by_user_options
+      options = {}
+
+      changes.each do |change|
```
koetsier: could filter the changes instead and then iterate.

`changes.keep_if {|change| change.key? :user_id }.each...`

Could also use inject/each_with_object to build the options:

```
filtered_changes.each_with_object({}) do |change, options|
  options[change[:user_name]] = change[:user_uid]
end
```

cbaines: I don't particularly like the chaining with blocks, mostly because it looks really messy, I do really like the use of each_with_object (or reduce as I would normally call it...), so I've switched to that here 3f7304a

PR #465, opened 2017-09-19
Update project filters to new states

tomsabin:
This changes the tagging participant's workflow to show 'incomplete'
content items as the default view in the project page. Once a content
item has been flagged or marked as done, they will be filtered out from
the default list.

The filters have been updated to 'to do' (default), 'flagged' and 'done'.

[Trello](https://trello.com/c/tIWCM08k/266-filter-content-list-by-to-do-flagged-and-done)


app/views/projects/show.html.erb
```diff
@@ -9,19 +9,19 @@
...
-             class="form-control js-bulk-tagging-search-filter"
-             name="query"
-             value="<%= bulk_search.query %>"
+             class="form-control"
+             name="title_search"
+             value="<%= query.params[:title_search] %>"
              placeholder="Filter by title...">
 
-      <% Projects::BulkSearch::TAGGED_STATES.each do |state|  %>
+      <% ProjectFilterQuery::FILTERS.each do |state|  %>
```
whoojemaflip: could this be an instance method?

tomsabin: Yep, I've updated this in https://github.com/alphagov/content-tagger/pull/465/commits/7fd7ddd23f10c1ae1059fc4c76be10ee274a3dec#diff-2476f44a5f0802d146b181eeb56e8a12R17

PR #468, opened 2017-09-20
Show text feedback when saving tags for a content item

tomsabin:
We found that the visual colour flash when saving tags caused some confusion on whether the item was updated. Now, as well as the colour flash, we also display "Saved" or "Failed to save" in the interface.

## Save failed

[Trello](https://trello.com/c/NHABw8Sf/275-show-text-feedback-when-saving-tags-for-a-content-item)

### DIFF
--- a/app/assets/javascripts/bulk-tagger.js
+++ b/app/assets/javascripts/bulk-tagger.js
@@ -112,14 +112,27 @@
       $content_item_forms.on(
         'ajax:success',
         function() {
+          display_response_state($(this), "Saved");
           self.mark_form_as_successfully_updated($(this));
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/bulk-tagger.js
```diff
@@ -112,14 +112,27 @@
       $content_item_forms.on(
         'ajax:success',
         function() {
+          display_response_state($(this), "Saved");
```
whoojemaflip: Should this have a "Saving..." state also? How will users know that the 2nd or 3rd tag they enter has been saved?

PR #471, opened 2017-09-20
Make Bulk Tagging UI modal and async

whoojemaflip:
This PR contains enough Javascript to allow the `Done` and `Flag for review` user actions to be performed without a page refresh.

ajax

This is just a reactive framework away from being a Single Page Application at this point.

[Trello](https://trello.com/c/MO1410d6/296-done-and-flag-for-review-can-be-used-without-triggering-a-page-refresh)


app/assets/javascripts/bulk-tagger.js
```diff
@@ -76,6 +77,18 @@
           self.mark_form_as_failed_to_update($(this));
         }
       );
+
+      $element.find(self.selectors.done_form_selector).on(
+        'ajax:success',
+        function() {
+          $(this).parents('.content-item').fadeOut('fast');
```
tomsabin: It's not necessary to do this but should we also remove the element from the DOM when the element has faded out?

app/controllers/project_content_items_controller.rb
```diff
@@ -21,13 +21,20 @@ def bulk_update
   end
 
   def flags
-    render :flags, locals: { project: project, content_item: content_item }
+    locals = { project: project, content_item: content_item }
+
+    respond_to do |format|
+      format.js { render layout: false, locals: locals }
```
tomsabin: Do we still want to allow these flags to be updated without JS? With the `flags.html.erb` file removed, this'll only work with JavaScript. Perhaps we could include the flags partial (without the modal styling) so that we can provide both the JS enhancement and the HTML page.

app/views/project_content_items/update_flags.js.erb
```diff
@@ -0,0 +1,2 @@
+$('#flags-modal').modal('toggle');
+$('.content-item form[data-ref="<%= content_item.id %>"]').parent('.content-item').fadeOut('slow');
```
tomsabin: Same question as a [previous comment](), should we also remove the element from the page when it's visually hidden?

tomsabin: Also not to be too picky, but to be consistent with the animation style, maybe this can fade out fast to match what happens when you press "Done".

app/views/projects/_content_item.html.erb
```diff
@@ -31,7 +31,7 @@
       &nbsp;
     </p>
     <div class='actions actions-inline'>
-      <%= button_to 'Done', mark_done_project_content_item_path(content_item.project, content_item), class: 'btn btn-primary' %>
+      <%= button_to 'Done', mark_done_project_content_item_path(content_item.project, content_item), class: 'btn btn-primary', form_class: 'js-mark-as-done', remote: true %>
```
tomsabin: When this button is pressed even if no action is taken, the modal is hidden and the element is removed. 

Perhaps we could get around this with HTML5 form validation to require a field to be selected (and maybe text to be input if the second option is selected). Alternatively, hook onto this form submit event with the same style as the Done form.

sep-21-2017 14-36-19

spec/features/flagging_a_content_item_spec.rb
```diff
@@ -4,14 +4,14 @@
   include TaxonomyHelper
   include PublishingApiHelper
 
-  scenario "flagging a content_item as 'I need help'" do
+  scenario "flagging a content_item as 'I need help'", js: true do
```
tomsabin: I think we should retain the non-JS scenario if that's what we're meant to be supporting. This is quite a vital part of the application that should still be covered as a feature specs.

app/views/project_content_items/_flags.html.erb
```diff
@@ -0,0 +1,36 @@
...
+
+      <div id='suggested-tags' class='form-group add-top-margin add-left-border'>
+        <%= f.label :suggested_tags, "Suggest a new topic" %>
+        <%= f.text_field :suggested_tags, class: 'form-control input-md-4' %>
+      </div>
+
+      <div class='modal-footer'>
+        <%= f.submit "Continue", class: 'btn btn-success add-right-margin' %>
+        <button type="button" class="btn btn-link" data-dismiss="modal">Cancel</button>
+      </div>
```
cbaines: If you move the modal-footer out of the modal-body, it looks better. I've pushed my change up here [1].

1: https://github.com/alphagov/content-tagger/commit/8766ff791d87dc13b573d1464bd75cf5cb4ff755

spec/features/flagging_a_content_item_spec.rb
tomsabin: As we discussed yesterday, the primary tagging functionality only works with select2, so we don't have a non-JS version. This isn't a problem currently as we're providing laptops for the tagathon. Feel free to ignore this comment and the [controller response comment](https://github.com/alphagov/content-tagger/pull/471#discussion_r140239102).

app/assets/javascripts/bulk-tagger.js
whoojemaflip: Yeah I was wondering about this. There will be some undesirable interaction between the auto-hide and the bulk-tagger interface too, but I'm not going to address that here. I suspect that we can remove the multi-tag UI now and save the effort of maintaining it.

app/views/project_content_items/update_flags.js.erb
whoojemaflip: I played around with different animation speeds, and I think the slower fade here works well. There's a few things happening onscreen, and if this fades too fast then it's easy to miss what's happened while the modal closes.

app/views/projects/_content_item.html.erb
tomsabin: Seems this still _might_ be a problem for when the suggested tags input field is left blank after toggling the second option:

sep-22-2017 14-23-39

PR #472, opened 2017-09-21
Ability to delete a project

tomsabin:
Users with the `GDS Editor` permission can delete projects. To be used to clean up test projects before the mini tagathon.

[Trello](https://trello.com/c/eevVDIt9/270-ia-can-delete-a-project-in-tagging-tool)

### DIFF
--- a/app/views/projects/confirm_delete.html.erb
+++ b/app/views/projects/confirm_delete.html.erb
@@ -1,0 +1,18 @@
+<header class="heading-with-actions">
+  <h1><%= project.name %></h1>
+</header>
+
+<div class="lead">
+  You are about to <strong>delete</strong> this project - this will remove the
+  project and any participant suggested tags that may still be outstanding,
+  and cannot be undone.
+</div>
+
+<div class="confirmation-box">
+  <%= button_to "Confirm delete",
+    project_path(project),
+    method: :delete,
+    class: 'btn btn-danger add-right-margin' %>
+
+  <%= link_to "Cancel", :back, class: "cancel-link" %>
+</div>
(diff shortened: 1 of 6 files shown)


app/views/projects/confirm_delete.html.erb
```diff
@@ -0,0 +1,18 @@
+<header class="heading-with-actions">
+  <h1><%= project.name %></h1>
+</header>
+
+<div class="lead">
+  You are about to <strong>delete</strong> this project and the project content
+  items that belong to it - this will remove the project and any suggested tags
```
cbaines: I'd suggest leaving out `and the project content items that belong to it ` as I don't think project content items is a concept that we want to expose.

app/views/projects/confirm_delete.html.erb
```diff
@@ -0,0 +1,18 @@
...
+  for content items from the system, and cannot be undone.
```
cbaines: It would be good to try and clarify this more, as I think "remove ... any suggested tags for content items from the system" sounds like it removes the tags that it has been used to create, which I don't think is the case.

cbaines:
I've tested this locally, and it works for me :)

whoojemaflip:
This change will need the `Information architect` role to be created in Signon Production & Integration. Hopefully @tijmenb can comment on that.

app/lib/permission_checker.rb
```diff
@@ -1,6 +1,7 @@
 class PermissionChecker
   GDS_EDITOR_PERMISSION = "GDS Editor".freeze
   TAGATHON_PARTICIPANT_PERMISSION = "Tagathon participant".freeze
+  IA_PERMISSION = "Information architect".freeze
```
whoojemaflip: Capital `A`?

tomsabin:
Based on what @tijmenb just said about the GDS Editor role, it seems that the additional role isn't necessary, so I'm going to remove the `Information architect` permission.

app/views/projects/confirm_delete.html.erb
tomsabin: If Sona were here I'd check with her, but what do you think about the following:

> You are about to delete this project - this will remove the project and any participant suggested tags that may still be outstanding, and cannot be undone.

PR #474, opened 2017-09-21
Run gov-lint-ruby as part of default rake task

tomsabin:
There's been numerous times where I've forgotten to run the govuk lint
task and see CI fail because of it.

When running the default rake task, we should include the lint task and
have it run in a similar way to CI.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -4,3 +4,5 @@
 require File.expand_path('../config/application', __FILE__)
 
 Rails.application.load_tasks
+
+task(:default).prerequisites << "lint" unless ENV.has_key?("JENKINS_URL")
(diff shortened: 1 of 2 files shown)


tijmenb:
I think this will make it run twice on CI, perhaps skip it there by checking for `JENKINS`env variable?

tomsabin:
I've used `JENKINS_URL` as there wasn't a `JENKINS` environment variable. Out of the  [documented env vars](https://wiki.jenkins.io/display/JENKINS/Building+a+software+project#Buildingasoftwareproject-belowJenkinsSetEnvironmentVariables), that seemed like the most descriptive.

PR #476, opened 2017-09-22
Use gds-api-adapers for link changes

koetsier:
Removed the temporary api call to get the link changes from Publishing Api and use
the one provided by gds-api-adapters.

trello: https://trello.com/c/nYBkNBDc/294-switch-to-using-the-link-changes-endpoint-through-the-gds-api-adapters-gem


spec/features/tagging_history_spec.rb
```diff
@@ -110,7 +110,7 @@ def then_i_see_the_link_changes_for_the_individual_taxon
   end
 
   def stub_link_changes_request(link_changes, params = { link_types: ['taxons'] })
-    stub_request(:get, "#{PUBLISHING_API}/v2/links/changes?#{params.to_query}")
+    stub_request(:get, "#{PUBLISHING_API}/v2/links/changes/?#{params.to_query}")
```
cbaines: I think it would be good to fix this, as I probably got the URL construction wrong when I wrote it (assuming I did write it). The `/` in the gds-api-adapters is unnecessary [1], and might even cause problems.

1: https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/publishing_api_v2.rb#L412

PR #479, opened 2017-09-22
Completely remove the Bulk Tagger UI

whoojemaflip:
User testing suggested that this feature wasn't as useful as we'd hoped - it encouraged low-quality tagging of multiple items to a higher level term than desirable.

It's been hidden by default for a while now, but this just means it's harder to maintain and it's at risk of breakage through inattention.


cbaines:
> User testing suggested that this feature wasn't as useful as we'd hoped - it encouraged low-quality tagging of multiple items to a higher level term than desirable.
> 
> It's been hidden by default for a while now, but this just means it's harder to maintain and it's at risk of breakage through inattention.

I'm reasonably happy with keeping this code for the time being, so I'm going to close this PR. Thanks for doing it though, if it is decided to remove it, we can come back to this.

PR #489, opened 2017-10-05
Iframe_allowing_proxy now handles headers without content-type

koetsier:


app/lib/proxies/iframe_allowing_proxy.rb
```diff
@@ -21,7 +21,7 @@ def rewrite_response(triplet)
       status, headers, body = triplet
 
       result = []
-      if headers['content-type'].any? { |header| header.include?('text/html') }
+      if headers['content-type'] && headers['content-type'].any? { |header| header.include?('text/html') }
```
cbaines: `headers.fetch('content-type', []).any? ...` might work, and might be neater here.

PR #490, opened 2017-10-05
Add rake task to export tagging metadata to JSON

davidbasalla:
For https://trello.com/c/cdMOOFtb/14-make-mini-tagathon-data-available-for-analysis

We add a rake task that writes out a JSON file with information about tagged content items for the Transport taxonomy (as tagged in the mini-tagathon). For each content item, we provide the following information:
```
  {
    "url": "/government/publications/great-western-franchise-2013",
    "content_id": "5c7ac403-7631-11e4-a3cb-005056011aef",
    "taxons": [
      {
        "organisation_slug": "department-for-transport",
        "user_uid": "5071c800-b495-012f-73f1-12313c09b582",
        "url": "/transport/rail-procurement-documents",
        "depth": 3,
        "content_id": "bd191af7-cb8b-47d5-bcd5-82e7f95fe96d"
      }
    ],
  }
```

Paired with @tomsabin 🎉

### DIFF
--- a/lib/tagged_content_exporter.rb
+++ b/lib/tagged_content_exporter.rb
@@ -1,0 +1,63 @@
+class TaggedContentExporter
+  def self.call
+    transport_taxon_id = "a4038b29-b332-4f13-98b1-1c9709e216bc"
+    content_items = ProjectContentItem.for_taxonomy_branch(transport_taxon_id).done
+
+    content_items = content_items.map do |c|
+      puts c.content_id
+
+      content_item = {}
+      content_item["content_id"] = c.content_id
+      content_item["url"] = c.url.gsub("https://www.gov.uk", "")
+      response = Services.publishing_api.get_links(c.content_id)
+
+      if response.to_h["links"]["taxons"].blank?
+        puts "--FAIL no taxons in links hash"
+        next
+      end
+
+      content_item["taxons"] = response.to_h["links"]["taxons"].map do |taxon_id|
+        puts "--#{taxon_id}"
+
+        params = { source_content_ids: [c.content_id], link_types: ["taxons"] }
+        taxon_change = Services
+          .publishing_api
+          .get_links_changes(params)
+          .to_h["link_changes"]
+          .find { |link| link.dig("target", "content_id") == taxon_id }
+
+        if taxon_change.blank?
+          puts "--FAIL taxon_change is blank"
+          next
+        end
+
+        user = User.find_by(uid: taxon_change.dig("user_uid")) || User.new
+        taxon_hash = Services.publishing_api.get_expanded_links(taxon_id).to_h
+
+        {
+          content_id: taxon_id,
+          depth: self.taxon_depth(taxon_hash),
+          url: taxon_change.dig("target", "base_path"),
+          user_uid: taxon_change.dig("user_uid"),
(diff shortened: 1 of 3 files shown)


lib/tagged_content_exporter.rb
```diff
@@ -0,0 +1,72 @@
...
+      {
+        content_id: taxon_id
+      }.merge(metadata_for_taxon(taxon_id, content_item.content_id))
+    end
+  end
+
+  def metadata_for_taxon(taxon_id, content_id)
+    metadata = taxon_links_change(taxon_id, content_id)
+
+    user = User.find_by(uid: metadata.dig("user_uid")) || User.new
```
cbaines: The `User.new` bit here looks a bit odd to me.

Another option might be to not do that, and use `user.try(:organisation_slug)` below?

lib/tagged_content_exporter.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  def taxon_links_change(taxon_id, content_id)
+    changes = Services
+      .publishing_api
+      .get_links_changes(source_content_ids: [content_id],
+                         link_types: ["taxons"])
+      .to_h["link_changes"]
+
+    changes.find { |link| link.dig("target", "content_id") == taxon_id } || {}
```
cbaines: You might be able to get the publishing-api to do this if you pass `target_content_ids`.

lib/tagged_content_exporter.rb
```diff
@@ -0,0 +1,72 @@
...
+      Services.publishing_api.get_expanded_links(taxon_id).to_h
+    )
+  end
+
+  def calculate_taxon_depth(taxon_hash, depth = 0)
+    if taxon_hash.has_key?("links") && taxon_hash["links"].blank?
+      depth
+    else
+      links = taxon_hash["expanded_links"] || taxon_hash["links"]
+      calculate_taxon_depth(links["parent_taxons"].first, depth + 1)
```
cbaines: Nice recursion 👍

lib/tagged_content_exporter.rb
```diff
@@ -0,0 +1,72 @@
...
+    @content_items = ProjectContentItem
+      .for_taxonomy_branch(TRANSPORT_TAXON_ID)
+      .done
+  end
+
+  def content_items_with_taxons
+    @content_items.map do |content_item|
+      {
+        content_id: content_item.content_id,
+        url: content_item.url.gsub("https://www.gov.uk", ""),
```
cbaines: You could use Plek here, to make this environment independent.

lib/tagged_content_exporter.rb
tomsabin: This was added as a lazy implementation for the null pattern (e.g. instead of a `NullUser` object, perhaps). I think originally we were also taking out the `uid` from the user, so following this pattern was more beneficial before hand rather than scattering `try` around.

lib/tagged_content_exporter.rb
davidbasalla: Ah good point! We've turned over the data already, so if we end up using this code again we'll make the improvement then.

lib/tagged_content_exporter.rb
davidbasalla: Amended and rebased

PR #491, opened 2017-10-11
Include taxonomy path in tagged content item export

tomsabin:
https://trello.com/c/cdMOOFtb/14-make-mini-tagathon-data-available-for-analysis

### DIFF
--- a/lib/tagged_content_exporter.rb
+++ b/lib/tagged_content_exporter.rb
@@ -1,6 +1,4 @@
 class TaggedContentExporter
-  WEBSITE_ROOT = Plek.new.website_root
-
   def initialize(content_items)
     @content_items = content_items
   end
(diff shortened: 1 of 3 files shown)


lib/tagged_content_exporter.rb
```diff
@@ -63,4 +65,23 @@ def calculate_taxon_depth(taxon_hash, depth = 0)
       calculate_taxon_depth(links["parent_taxons"].first, depth + 1)
     end
   end
+
+  def calculate_taxon_path(taxon_hash, paths = [])
```
davidbasalla: Could we combine this somehow with the `calculate_taxon_depth` method? It looks like it's doing the same recursion so not sure what the performance hit is when we do it twice

davidbasalla:
Looks good, only one minor comment

PR #493, opened 2017-10-13
Rename BulkTagger to TaxonTagger to avoid confusion

tomsabin:
We should avoid using the BulkTagger module name when it is not
directly tied to the "bulk tagger" interface that is hidden by default.

### DIFF
--- a/app/assets/javascripts/taxon-tagger.js
+++ b/app/assets/javascripts/taxon-tagger.js
@@ -2,7 +2,7 @@
   "use strict";
 
   // Constructor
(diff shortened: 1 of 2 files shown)


tijmenb:
Does TaxonTagger sound like it's tagging taxons?

tomsabin:
@tijmenb yes I think that could be an interpretation of the name, however I'm hoping that given the context of where it's used, `TaxonTagger` would be a clearer name of its main function (to tag taxons to content items) than `BulkTagger`.

PR #494, opened 2017-10-13
Add comment field for need help option

davidbasalla:
For https://trello.com/c/CJRdMIYn/40-allow-text-comments-to-be-left-when-a-user-needs-help-tagging-this-and-display-this-comment-in-flagged-view

See screenshots below

------------------------------------------------------

## Before
screen shot 2017-10-13 at 15 40 50

screen shot 2017-10-13 at 15 42 27

## After

screen shot 2017-10-16 at 14 50 43

### Short comment

screen shot 2017-10-16 at 12 22 56

### Longer comment

screen shot 2017-10-16 at 12 22 40

### DIFF
--- a/app/views/projects/_content_item.html.erb
+++ b/app/views/projects/_content_item.html.erb
@@ -3,7 +3,15 @@
     <%= token_tag %>
 
     <% if content_item.needs_help? %>
-      <span class='label label-danger'><%= I18n.t('views.projects.flags.help-needed') %></span>
+      <div class='flagged-for-needs-help'>
+        <div>
+          <span class='label label-danger'><%= I18n.t('views.projects.flags.help-needed') %></span>
+        </div>
+
+        <% if content_item.need_help_comment.present? %>
+            <span class="needs-help-comment">Comment: <%= content_item.need_help_comment %></span>
+        <% end %>
+      </div>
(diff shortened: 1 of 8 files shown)


app/views/projects/_content_item.html.erb
```diff
@@ -16,6 +16,11 @@
     </h4>
 
     <p><%= content_item.description %></p>
+
+    <% if content_item.needs_help? && content_item.need_help_comment.present? %>
+      <p>Comment: <%= content_item.need_help_comment %></p>
```
tomsabin: Perhaps this text could shown next to the `help-needed` flag, similar to the "Suggested topic"?

davidbasalla:
Yeah, as briefly discussed offline, the main reason is probably that we don't know how long the text comment would be, and it might look strange when it goes over multiple lines. I've add some warning alert styling to the comment to make it pop a bit more, but will discuss with Luke about what her prefers...

tomsabin:
I've deploy this to integration, perhaps we also want to include the styling for the "Suggested topic" flag? See the screenshot below with items with each flag.

screen shot 2017-10-18 at 09 23 45

davidbasalla:
@tomsabin sounds good, I'll give it a go as it would be nice to standardise the styling

PR #498, opened 2017-10-20
Make 'suggested tags' field optional

davidbasalla:
For https://trello.com/c/4nQoaMMe/60-tag-suggestions-are-no-longer-mandatory-if-a-tagger-says-there-is-no-suitable-topic

Makes the 'suggested tags' field when flagging a ProjectContentItem as 'there's not relevant topic' optional. This was enforced via JavaScript so we just remove the relevant lines.

We also don't want flagged ContentItems that have no suggested taxons to
show up on the index page. This commit updates the FlaggedContentQuery
class to not return those content items. We still need to return all
content items that are flagged with "needs_help".

Since this makes the FlaggedContentQuery more complex, we add a test
that documents this behaviour.

### DIFF
--- a/app/queries/flagged_content_query.rb
+++ b/app/queries/flagged_content_query.rb
@@ -10,6 +10,20 @@ class FlaggedContentQuery
   end
 
   def items
+    if params[:flagged] == "missing_topic"
+      project_content_items_with_suggested_tags_only
+    else
+      project_content_items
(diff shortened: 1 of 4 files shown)


davidbasalla:
Cool thanks, will merge this in the interest of time as I won't get much time today with all the meetings :)

PR #501, opened 2017-10-24
Show tagging progress for each organisation on projects page

tomsabin:
[Trello](https://trello.com/c/m0kfPG5I/38-show-tagging-progress-made-by-an-organisation)

To surface overall tagging progress for an organisation, we display the
percentage of content tagged on the projects index page. This uses rummager as
the source of all content for an organisation, comparing to content items that
have a taxon applied to it.

---

The table below shows on the projects index page

> screen shot 2017-10-25 at 16 18 34

>  screen shot 2017-10-25 at 17 00 05


davidbasalla:
As discussed, the issue around `NaN%` should be resolved (where orgs have 0 total content), possibly via a product decision about whether to show these at all

spec/features/project_spec.rb
```diff
@@ -146,7 +175,7 @@ def then_i_see_the_content_item_and_its_tag_data
   end
 
   def and_i_click_my_new_projects_name_in_the_project_list
-    within 'table' do
+    within 'table:first-of-type' do
```
davidbasalla: What does the `first-of-type` and `last-of-type` do?

davidbasalla:
Looks good! 👍 

Just wondering... did you have a unit test before for "tagging_progress_by_organisations_query.rb"? I guess it's pretty well covered by the feature spec, but could add it back in for completeness' sake.

spec/features/project_spec.rb
tomsabin: There were added in when two tables were being shown and Capybara didn't know which table to selector. It was a quick solution to use the CSS selectors to find the correct table based on the order in which they show on the page, but this isn't the case anymore because we do not show the table if there is no content (for both projects and the tagging progress tables).

I'll update this selector so it's clearer.

tomsabin:
> Just wondering... did you have a unit test before for "tagging_progress_by_organisations_query.rb"? I guess it's pretty well covered by the feature spec, but could add it back in for completeness' sake.

@davidbasalla I started out with a unit test but moved it to a feature spec when there was a clarified use case for the tagging progress in the projects page. The unit test was almost exactly the same but without the presentation, so I decided to remove it.

tomsabin:
This PR is dependent on https://github.com/alphagov/rummager/pull/962, so I'm waiting until that PR has been merged before merging & deploying this PR to integration.

PR #507, opened 2017-11-07
Remove themes concept

tomsabin:
https://trello.com/c/CyEREE5H/91-sort-out-themes-in-content-tagger

We have removed the Theme concept from content-tagger as it's a
restriction that was hindering the Taxon creation process. Before, it
was mandatory to assign a Theme to a Taxon, which was primarily used to
defined path prefix (e.g. /education). This URL structure was [defined
earlier this year](https://gov-uk.atlassian.net/wiki/spaces/FS/pages/127664172/URL+structure+for+the+new+navigation+pages).

Now we have removed the Theme model and requirement, the URL restriction
is dictated by the selected parent taxon:

- For root taxons (i.e. a taxon with no parent taxon), the user is free
  to define the slug.
- For non-root taxons, the path will be prefixed by the selected parent
  taxon, e.g. if a parent taxon is selected that is from the Education
  branch, the path will be prefixed with "/education".
---

Editing a root taxon

> screen shot 2017-11-15 at 10 34 40

Editing a 3rd level taxon

> screen shot 2017-11-15 at 10 35 13

New taxon

> nov-15-2017 10-34-00


tijmenb:
I may be wrong about this, but will this set the first part of the URL to whatever is the URL of the taxon parent?

tomsabin:
@tijmenb no, there isn't any automatic prefixing for the path/URL field. The path that's displayed beneath the Parent field is only a hint. This video might help show the user interaction:

nov-07-2017 10-09-26

tijmenb:
To recap from a offline / Slack conversation.

The reason for having the structure `/#{root-branch-slug}/#{taxon.slug}":

- More tightly control the root level of GOV.UK. Historically, having things published on gov.uk/something has caused trouble. It's not that we don't ever want it, but it should be done carefully. The alternative is to nest everything within a `/topic` namespace or something, but that's ugly and doesn't tell the user much.
- The URLs become more informative by showing the context (gov.uk/education/exam-regulation-and-administration vs gov.uk/exam-regulation-and-administration)
- The prefix prevents potential conflicts of similarly named taxons. An example could be https://www.gov.uk/childcare-parenting/pregnancy-and-birth, you could easily see a separate topic at https://www.gov.uk/healthcare/pregnancy-and-birth

Some background on this wiki page: https://gov-uk.atlassian.net/wiki/spaces/FS/pages/127664172/URL+structure+for+the+new+navigation+pages

I would've thought that the solution would be to populate the dropdown from the [root taxons](https://github.com/alphagov/content-tagger/blob/master/app/models/govuk_taxonomy/branches.rb) but @tomsabin explained the chicken-egg problem that you wouldn't be able to create the taxon before promoting it.

Perhaps a solution is to add separate UI to "create" root taxons, which would allow you to create a thing at root? The edit page already has a [special case for root taxons](https://github.com/alphagov/content-tagger/blob/master/app/views/taxons/_form_fields.html.erb#L34-L41) so there's prior art.

tijmenb:
Oh yeah, that's pretty elegant. 👍

app/assets/javascripts/taxon-path-prefixer.js
```diff
@@ -0,0 +1,47 @@
...
+        $pathPrefixInputEl.val('');
+
+        $previewBasePathEl.text(pathPrefix);
+      }
+
+      updateBasePath();
+      $parentSelectEl.change(updateBasePath);
+
+      var timeout = null;
+      $pathSlugInputEl.on('keyup', function() {
```
davidbasalla: When a user populates the field without keyboard input (pasting something via right-click -> 'Paste' or selecting an auto-complete option provided by the browser) the preview wouldn't trigger. Just tested this on my dev vm.

app/assets/javascripts/taxon-path-prefixer.js
```diff
@@ -0,0 +1,47 @@
...
+        clearTimeout(timeout);
+        timeout = setTimeout(updateBasePath, 500);
```
davidbasalla: Just curious about this timeout... why did you go for 500 or a timeout in general?

davidbasalla:
As discussed offline, we should avoid relying on JavaScript to set the path_prefix for non-root Taxons because it's a critical part of the base path creation. 

For example, with JavaScript disabled, if you were to select a parent taxon (e.g. `/childcare-parenting`) and define a slug (e.g. `/finding-childcare`), the string concatenation would not occur as the path_prefix value would not defined (as it's expected to be filled from the ajax request when the parent taxon field is selected). This would lead to that taxon being created only with the base path of `/finding-childcare` instead of `/childcare-parenting/finding-childcare` which is not what was intended.

Perhaps we could move that path_prefix logic into the taxon controller instead, to remove the dependency on JavaScript.

(actually written by Tom)

davidbasalla:
Something to look out for when it comes to Product Review: It appears that in the past we were careful about creating new base paths on the root level of GOV.UK (see Tijmen's comment above), and the only way to do so in ContentTagger was by creating a new theme programmatically. With this PR we've sort of unshackled any restrictions on creating a root taxon, so it might be worth to have a think about sign-posting the significance of creating root taxons more clearly (ie some warning text in the Form maybe)...

PR #513, opened 2017-11-09
Add longer timeouts to improve Capybara stability

koetsier:
Capybara tests, mostly involving modals have been failing intermittently. This PR adds longer timeouts to some of the Capybara assertions giving them longer to appear.


koetsier:
DId not really end up improving matters

PR #514, opened 2017-11-09
Add tagged content count to taxon show page

davidbasalla:
For https://trello.com/c/17sEgPum/86-show-per-taxon-metrics-in-content-tagger

Adds a metadata section to the show page, including a count of tagged content.

First step in adding more metadata to the show taxon page. Eventually we'll display another count here for tagged content to the taxon's child taxons too.

--------------

## Screenshot

screen shot 2017-11-09 at 16 46 18

### DIFF
--- a/app/views/taxons/show.html.erb
+++ b/app/views/taxons/show.html.erb
@@ -68,6 +68,12 @@
     </div>
   <% end %>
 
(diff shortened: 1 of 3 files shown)


cbaines:
Could this perhaps be put in the title, e.g. "Pages tagged to this taxon (173)"? I think this might be more concise and still clear?

PR #516, opened 2017-11-14
[DO NOT MERGE] Set visible_to_department_editors to true on publish

davidbasalla:
For https://trello.com/c/b3TTdin2/157-prevent-publishers-seeing-draft-taxons-in-government-frontend

Add a `PublishService` class to encapsulate the logic for setting the `visible_to_department_editors` field to `true` when a taxon is published. This commit goes hand in hand with code changes to NavigationHelpers to enable showing/hiding taxons based on the `visible_to_department_editors` value.

### DIFF
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -104,7 +104,7 @@ class TaxonsController < ApplicationController
   end
 
   def publish
-    Services.publishing_api.publish(content_id)
+    Taxonomy::PublishTaxon.call(taxon)
     redirect_to taxon_path(content_id), success: "You have successfully published the taxon"
   end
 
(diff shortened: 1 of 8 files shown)


app/models/taxon.rb
```diff
@@ -93,4 +93,8 @@ def visible_to_departmental_editors=(val)
   def visible_to_departmental_editors
     @visible_to_departmental_editors || false
   end
+
+  def root_taxon?
+    parent.nil?
```
cbaines: I think this is a bit risky, as _real_ root taxons are probably referred to by the homepage...

davidbasalla: Looking at some examples, I don't think they are. But do you have any suggestions to make this less risky?

cbaines: > Looking at some examples, I don't think they are. But do you have any suggestions to make this less risky?

Well, what it really tests is does this taxon have a parent (e.g. `has_parent?`). Checking if it is a root taxon requires checking the links for the homepage (which is what the GovukBranches module/class does if I remember correctly).

davidbasalla: Ah right, so I think I'll remove that method, and just replace it by a `@taxon.parent.nil?` check. 
The code for displaying the checkbox for this attribute also does this -> https://github.com/alphagov/content-tagger/blob/master/app/services/taxonomy/edit_page.rb#L27-L29

davidbasalla:
Marking as do not merge as we might abandon this approach...

davidbasalla:
Closing in favour of https://github.com/alphagov/govuk_navigation_helpers/pull/82

PR #519, opened 2017-11-21
Coverage measurement WIP

koetsier:
Adds measurement of the number of content items tagged to a particular level of the taxonomy.

The measurements are sent to Statsd as `govuk.tagging.level_n.content_tagged` where 'n' is the level of the taxonomy, counting from 1.


.rubocop.yml
```diff
@@ -34,9 +34,7 @@ Metrics/ClassLength:
   Enabled: true
 
 Metrics/ModuleLength:
-  # TODO: bring this down to 100
-  Max: 150
-  Enabled: true
+  Enabled: false
```
tomsabin: This change seems to only be for the `spec/services/taxonomy/taxonomy_query_spec.rb` file. Perhaps we could describe the class in the spec as follows:

```diff
- module Taxonomy
-   RSpec.describe TaxonomyQuery do
+ RSpec.describe Taxonomy::TaxonomyQuery do
```

spec/services/taxonomy/taxonomy_query_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+      end
+
+      context 'there are root taxons and two levels of children' do
+        before :each do
+          allow(Services.content_store).to receive(:content_item).with('/').and_return root_taxon
+          allow(Services.content_store).to receive(:content_item).with('/taxons/root_taxon')
+                                             .and_return multi_level_child_taxons
+        end
+        it 'returns three levels' do
+          expect(TaxonomyQuery.new.taxonomy(:content_id, :base_path).size).to eq(3)
```
tomsabin: Reading this without any knowledge of the internals of `TaxonomyQuery#taxonomy`, I'd expect this method to be called `depth`. I find "size of taxonomy" to be ambiguous, could it be depth, width, number of taxons?

app/services/data_export/taxon_export.rb
```diff
@@ -1,10 +1,8 @@
 module DataExport
   class TaxonExport
-    TAXON_FIELDS = %w[content_id base_path title].freeze
```
tomsabin: I'm not sure what benefit we're getting having removed this constant. Instead, we're now passing down  `content_id` and `base_path` in almost every method in `TaxonomyQuery` which I think is causing unnessary noise. That is unless I've massively overlooked something!

spec/services/taxonomy/taxonomy_query_spec.rb
koetsier: TaxonomyQuery.new.taxonomy returns an Array of numbers - to have a 'depth' method I'd have to have it return an instance of a custom class, which seems a bit heavyweigth

.rubocop.yml
koetsier: fixed

spec/services/taxonomy/taxonomy_query_spec.rb
tomsabin: It might be something we want to have, though perhaps I'm the only one that finds the class interface quite difficult to understand what going to be returned. 

Using one of the `taxonomy_query_spec.rb` examples, we seem to be returning arrays of hashes and then relying on Ruby's array methods to return the types of queries we're after.

```rb
taxonomy = TaxonomyQuery.new.taxonomy
=> [
  [
    {"content_id"=>"root_taxon_a", "base_path"=>"/taxons/root_taxon_a"},
    {"content_id"=>"root_taxon_b", "base_path"=>"/taxons/root_taxon_b"}
  ],
  [
    {"content_id"=>"child_a_1", "base_path"=>"/taxons/child_a_1"},
    {"content_id"=>"child_a_2", "base_path"=>"/taxons/child_a_2"},
    {"content_id"=>"child_b_1", "base_path"=>"/taxons/child_b_1"},
    {"content_id"=>"child_b_2", "base_path"=>"/taxons/child_b_2"}
  ]
]

taxonomy.size
=> 2

taxonomy.first
=> [
  {"content_id"=>"root_taxon_a", "base_path"=>"/taxons/root_taxon_a"},
  {"content_id"=>"root_taxon_b", "base_path"=>"/taxons/root_taxon_b"}
]

```

Perhaps renaming this class to something like `TaxonomyDepthQuery` that has public methods `depth`, `taxons_at_level(n)` might clarify how to use the query?

app/services/taxonomy/taxonomy_query.rb
```diff
@@ -0,0 +1,39 @@
+module Taxonomy
+  class TaxonomyQuery
```
tomsabin: I think my previous comment was lost in an [outdated conversation](https://github.com/alphagov/content-tagger/pull/519#discussion_r152553193) so I'm copying the comment I made in here as I still think it's applicable.

---

> TaxonomyQuery.new.taxonomy returns an Array of numbers - to have a 'depth' method I'd have to have it return an instance of a custom class, which seems a bit heavyweigth

It might be something we want to have, though perhaps I'm the only one that finds the class interface quite difficult to understand what going to be returned. 

Using one of the `taxonomy_query_spec.rb` examples, we seem to be returning arrays of hashes and then relying on Ruby's array methods to return the types of queries we're after.

```rb
taxonomy = TaxonomyQuery.new.taxonomy
=> [
  [
    {"content_id"=>"root_taxon_a", "base_path"=>"/taxons/root_taxon_a"},
    {"content_id"=>"root_taxon_b", "base_path"=>"/taxons/root_taxon_b"}
  ],
  [
    {"content_id"=>"child_a_1", "base_path"=>"/taxons/child_a_1"},
    {"content_id"=>"child_a_2", "base_path"=>"/taxons/child_a_2"},
    {"content_id"=>"child_b_1", "base_path"=>"/taxons/child_b_1"},
    {"content_id"=>"child_b_2", "base_path"=>"/taxons/child_b_2"}
  ]
]

taxonomy.size
=> 2

taxonomy.first
=> [
  {"content_id"=>"root_taxon_a", "base_path"=>"/taxons/root_taxon_a"},
  {"content_id"=>"root_taxon_b", "base_path"=>"/taxons/root_taxon_b"}
]

```

Perhaps renaming this class to something like `TaxonomyDepthQuery` that has public methods `depth`, `taxons_at_level(n)` might clarify how to use the query?

app/lib/services.rb
```diff
@@ -28,4 +28,19 @@ def self.rummager
       Plek.new.find('rummager'),
     )
   end
+
+  def self.rummager_enum(rummager_params, window: 100, size: Float::INFINITY)
```
tijmenb: Perhaps this can be moved into gds-api-adapters? The `Services` module is for API clients itself, and it would be useful to have add this auto-paginating option available elsewhere too.

cbaines:
I'm having trouble following what is going on in the first commit. The description is "Add a method to retrieve all taxons ordered by level", but I'm not quite sure where this method is,. The change regarding the `TaxonExport` class, and the `TAXON_FIELDS` looks like a good candidate to have in another commit.

app/services/taxonomy/taxonomy_query.rb
koetsier: I renamed this method to 'taxons per level' I hope that clarifies it a bit

app/lib/services.rb
koetsier: Good Idea, I'll put up a PR

app/services/taxonomy/taxonomy_query.rb
```diff
@@ -17,6 +17,19 @@ def child_taxons(base_path)
       recursive_child_taxons(taxons, root_content_hash['content_id'])
     end
 
+    def taxons_per_level
+      taxons = root_taxons
+      sibling_hashes = taxons.map { |h| get_content_hash(h['base_path']) }
+      recursive_taxons_per_level([], sibling_hashes)
+    end
+
+    def tagged_to_taxons(content_ids, slice_size: 50)
```
cbaines: Maybe `content_tagged_to_taxons`?

app/services/taxonomy/taxonomy_query.rb
```diff
@@ -17,6 +17,19 @@ def child_taxons(base_path)
       recursive_child_taxons(taxons, root_content_hash['content_id'])
     end
 
+    def taxons_per_level
+      taxons = root_taxons
```
cbaines: I think this is redundant, perhaps replace taxons with root_taxons.

app/services/taxonomy/taxonomy_query.rb
```diff
@@ -27,6 +40,18 @@ def recursive_child_taxons(taxons, parent_content_id)
       end
     end
 
+    def recursive_taxons_per_level(partial_results, sibling_hashes)
+      return partial_results if sibling_hashes.nil? || sibling_hashes.empty?
```
cbaines: You might just need the `empty?` check?

app/services/taxonomy/taxonomy_query.rb
```diff
@@ -27,6 +40,18 @@ def recursive_child_taxons(taxons, parent_content_id)
...
+      sibling_child_hashes = sibling_hashes.flat_map { |h| h.dig('links', 'child_taxons') }.compact
+      taxons = sibling_hashes.map { |h| slice_taxon_hash(h) }
+      recursive_taxons_per_level(partial_results + [taxons], sibling_child_hashes)
+    end
+
+    def slice_taxon_hash(taxon_hash)
+      slice = @taxon_fields.map(&:to_s).append('base_path').uniq
```
cbaines: There is some duplication regarding 'base_path' here with initialize.

spec/services/taxonomy/taxonomy_query_spec.rb
```diff
@@ -39,6 +39,71 @@ def query
     end
   end
 
+  describe '#tagged_to_taxons' do
+    it 'returns an empty array' do
+      expect(TaxonomyQuery.new.tagged_to_taxons([], slice_size: 50)).to eq([])
+    end
+    it 'returns tagged taxons' do
```
cbaines: returns content tagged to the taxons?

app/services/metrics/taxonomy_count.rb
```diff
@@ -0,0 +1,13 @@
+module Metrics
+  class TaxonomyCount
```
cbaines: What is the class adding here, could we just use the module?

app/services/metrics/taxonomy_count.rb
```diff
@@ -0,0 +1,13 @@
...
+    def self.count_taggings_per_level
```
cbaines: From looking at the code, I think this is counting content, rather than tags/taggings.

app/services/metrics/taxonomy_count.rb
```diff
@@ -0,0 +1,13 @@
...
+      counts_by_level = Taxonomy::TaxonomyQuery.new.taxons_per_level.map do |level|
```
cbaines: "level" is also used below, and even though its in a different scope, the meaning is quite different. I think "taxons" could work here as a name?

lib/tasks/taxonomy_metrics.rake
```diff
@@ -0,0 +1,8 @@
+namespace :metrics do
+  namespace :taxonomy do
+    desc "Count all content tagged to each level in the taxonomy"
+    task count_taggings_per_level: :environment do
```
cbaines: s/taggings/content as per other comment.

app/services/taxonomy/taxonomy_query.rb
cbaines: @koetsier this looks like its still relevant

app/services/taxonomy/taxonomy_query.rb
cbaines: @koetsier bump

cbaines:
@koetsier I bumped a couple of changes that we talked about yesterday, but apart from that I think this is looking good :)

PR #520, opened 2017-11-21
Content export exclude duplicates

cbaines:


app/services/data_export/content_export.rb
```diff
@@ -86,6 +90,14 @@ def content_links_enum(window = 1000, size = Float::INFINITY)
 
     def get_content(base_path, base_fields: CONTENT_BASE_FIELDS, taxon_fields: CONTENT_TAXON_FIELDS, ppo_fields: CONTENT_PPO_FIELDS)
       hash = get_content_hash(base_path)
+
+      # Skip this if we don't get back the content we expect, e.g. if
+      # the Content Store has redirected the request
+      return {} if hash.dig('base_path') != base_path
```
koetsier: if these are exceptional conditions, you could raise an error instead so it gets logged for future reference.

cbaines: I don't think this is exceptional, it just means we've encountered a redirect, either an document_type of `redirect` (which shouldn't happen that much) or some unpublished content where the unpublishing results in a redirect.

PR #521, opened 2017-11-23
Show a friendly error message on publish failure

thomasleese:
When trying to publish a taxon with the Publishing API sometimes this can fail. Rather than showing the user a 500 error, it seems like a good idea to show a slightly more useful error message while also recording the problem in Sentry for us to investigate.

This was found since Content Tagger relies on Publishing API to perform validation of base paths, and we will shortly be adding checks to make sure they are unique regardless of letter case which may affect users of the application.

[Trello Card](https://trello.com/c/TP0m8elk/317-case-insensitive-base-paths-in-content-tagger)

### DIFF
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -109,6 +109,10 @@ class TaxonsController < ApplicationController
   def publish
     Services.publishing_api.publish(content_id)
     redirect_to taxon_path(content_id), success: "You have successfully published the taxon"
+  rescue GdsApi::HTTPUnprocessableEntity => e
+    GovukError.notify(e, level: "warning")
+    flash.now[:danger] = I18n.t('errors.invalid_taxon')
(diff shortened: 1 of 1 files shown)


cbaines:
> This was found since Content Tagger relies on Publishing API to perform validation of base paths, and we will shortly be adding checks to make sure they are unique regardless of letter case which may affect users of the application.

This sounds fine in general.

On the specific point of base path validation, it seems odd to me that this would happen on publish, rather than when the content is updated (the PutContent command in the Publishing API)?

thomasleese:
@cbaines I was surprised by that too, but I was able to reproduce the problem of seeing a scary 500 error in integration by creating a taxon with the same base path as another. I think the work on adding that validation in the Publishing API will work on moving the validation into put content.

Showing a (albeit only slightly) more useful error to the user still feels like a good thing to do though.

PR #524, opened 2017-11-29
Add average tagging depth measurement.

koetsier:
Trello: https://trello.com/c/qfrviEaq/145-execute-content-tagging-coverage-measurement


app/services/metrics/content_per_level_metric.rb
```diff
@@ -1,13 +1,26 @@
 module Metrics
   class ContentPerLevelMetric
```
cbaines: Maybe changing this class would be helpful, now that its about a couple of sets of metrics. Something like ContentDistributionMetrics?

PR #534, opened 2017-12-04
Metrics changes

cbaines:
A few changes to metrics, to try and move taxonomy related metrics to a single namespace, and fix some issues that were missed in previous Pull Requests (like the missing blacklist file).

This follows on from the changes in https://github.com/alphagov/content-tagger/pull/525


app/services/metrics/content_distribution_metrics.rb
```diff
@@ -3,6 +3,7 @@ class ContentDistributionMetrics
     def count_content_per_level
       counts_by_level.each_with_index do |count, level|
```
koetsier: could do the `to_enum.with_index(1) ` like in the test in the previous commit

cbaines: I probably should have done this in a separate commit, but I did make this change in 6cb6857b4a0bf29488e96919e4b6ff2b4b01487d

PR #538, opened 2017-12-04
Document type tagging

koetsier:
Adds the `taxonomy:bulk_tag_document_type[document_type, taxon_base_path]` rake task to enable tagging of all content items of a given document type to a single taxon.

https://trello.com/c/ZIOFIqoT/162-bulk-tag-content-items-that-belongs-to-a-single-taxon

### DIFF
--- a/app/services/bulk_tagging/document_type_tagger.rb
+++ b/app/services/bulk_tagging/document_type_tagger.rb
@@ -1,0 +1,37 @@
+module BulkTagging
+  class DocumentTypeTagger
+    def self.call(taxon_base_path:, document_type:)
+      new(taxon_base_path: taxon_base_path, document_type: document_type).call
+    end
+
+    def initialize(taxon_base_path:, document_type:)
+      @taxon_base_path = taxon_base_path
+      @document_type = document_type
+    end
+
+    def call
+      taxon_content_id = Services.publishing_api.lookup_content_id(base_path: @taxon_base_path)
+      raise StandardError, "Cannot find taxon with base path #{@taxon_base_path}" if taxon_content_id.nil?
+
+      Services.publishing_api.get_content_items_enum(document_type: @document_type, fields: ['content_id']).lazy.map do |document|
+        begin
+          content_id = document.fetch('content_id')
+          new_taxons = add_taxon_link(content_id, taxon_content_id)
+          { status: 'success', message: 'success', content_id: content_id, new_taxons: new_taxons }
+        rescue StandardError => ex
+          { status: 'error', message: ex.message, content_id: content_id, new_taxons: [] }
+        end
+      end
+    end
+
+  private
+
+    def add_taxon_link(content_id, taxon_content_id)
+      response_hash = Services.publishing_api.get_links(content_id).to_h
+      version = response_hash['version']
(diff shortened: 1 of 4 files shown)


app/services/bulk_tagging/document_type_tagger.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    def call
+      taxon_content_id = Services.publishing_api.lookup_content_id(base_path: @taxon_base_path)
+      raise StandardError, "Cannot find taxon with base path #{@taxon_base_path}" if taxon_content_id.nil?
+
+      Services.publishing_api.get_content_items_enum(document_type: @document_type, fields: ['content_id']).lazy.map do |document|
+        begin
+          content_id = document.fetch('content_id')
+          Services.publishing_api.patch_links(content_id, links: { taxons: [taxon_content_id] })
```
cbaines: This is going to eliminate all the other tags for this content, which might be ok for some uses I guess, but maybe not all...?

app/services/bulk_tagging/document_type_tagger.rb
```diff
@@ -0,0 +1,27 @@
...
+    def self.call(taxon_base_path:, document_type:)
+      new(taxon_base_path: taxon_base_path, document_type: document_type).call
+    end
+
+    def initialize(taxon_base_path:, document_type:)
+      @taxon_base_path = taxon_base_path
+      @document_type = document_type
+    end
+
+    def call
```
cbaines: Looks like you could save some lines by just changing the above line to `self.call(taxon_base_path:, document_type:)` and removing the `self.call` and `initialize` thing, which might be clearer.

koetsier: I agree that this initialize->call thing is a bit weird, but it is used all over the place for services in content tagger so I used it for consistency sake.

app/services/bulk_tagging/document_type_tagger.rb
cbaines: I see that this has changed now, I think that is good 👍

app/services/bulk_tagging/document_type_tagger.rb
```diff
@@ -0,0 +1,36 @@
...
+        end
+      end
+    end
+
+  private
+
+    def add_taxon_link(content_id, taxon_content_id)
+      response_hash = Services.publishing_api.get_links(content_id).to_h
+      version = response_hash['version']
+      new_links = response_hash['links'].merge('taxons' => [taxon_content_id])
```
cbaines: I think this will have the same behaviour as before, to really add the `taxon_content_id` to the existing links you will need to combine the list of content ids.

app/services/bulk_tagging/document_type_tagger.rb
```diff
@@ -0,0 +1,37 @@
...
+        end
+      end
+    end
+
+  private
+
+    def add_taxon_link(content_id, taxon_content_id)
+      response_hash = Services.publishing_api.get_links(content_id).to_h
+      version = response_hash['version']
+      new_taxons = (Array.wrap(response_hash.dig('links', 'taxons')) + [taxon_content_id]).uniq
```
cbaines: This looks good :)

PR #544, opened 2017-12-05
Accept a content_id for bulk tagging of a document type

koetsier:
THe lookup function for finding a taxon using its base path only finds
live content. We switch to using content_id to identify a taxon to bulk tag

### DIFF
--- a/app/services/bulk_tagging/document_type_tagger.rb
+++ b/app/services/bulk_tagging/document_type_tagger.rb
@@ -1,22 +1,26 @@
 module BulkTagging
   class DocumentTypeTagger
-    def self.call(taxon_base_path:, document_type:)
-      new(taxon_base_path: taxon_base_path, document_type: document_type).call
+    def self.call(taxon_content_id:, document_type:)
+      new(taxon_content_id: taxon_content_id, document_type: document_type).call
     end
 
-    def initialize(taxon_base_path:, document_type:)
-      @taxon_base_path = taxon_base_path
+    def initialize(taxon_content_id:, document_type:)
+      @taxon_content_id = taxon_content_id
       @document_type = document_type
     end
 
     def call
-      taxon_content_id = Services.publishing_api.lookup_content_id(base_path: @taxon_base_path)
-      raise StandardError, "Cannot find taxon with base path #{@taxon_base_path}" if taxon_content_id.nil?
+      # Ensure @taxon_content_id exists in the publishing api
+      begin
+        Services.publishing_api.get_content(@taxon_content_id)
+      rescue GdsApi::HTTPNotFound
+        raise
+      end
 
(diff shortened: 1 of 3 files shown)


app/services/bulk_tagging/document_type_tagger.rb
```diff
@@ -1,22 +1,20 @@
...
+    def initialize(taxon_content_id:, document_type:)
+      @taxon_content_id = taxon_content_id
       @document_type = document_type
     end
 
     def call
-      taxon_content_id = Services.publishing_api.lookup_content_id(base_path: @taxon_base_path)
-      raise StandardError, "Cannot find taxon with base path #{@taxon_base_path}" if taxon_content_id.nil?
-
+      Services.publishing_api.get_content(@taxon_content_id)
```
cbaines: This call now looks pretty redundant, though I guess it will error if the provided taxon doesn't exist?

koetsier: thats why it is there

tomsabin: I'd rather we explicitly raise the error here if the taxon doesn't exist, the previous implementation documented what error case we were covering. However in this addition, it's not obvious as to why we're asking the publishing API for some content but not using its response.

I think this could be equivalent but shows that we're expecting a `HTTPNotFound` error to potentially be raised.

```ruby
begin
  Services.publishing_api.get_content(@taxon_content_id)
rescue GdsApi::HTTPNotFound => e
  raise
end

File: lib/tasks/taxonomy/bulk_tag_document_type.rake [hunk 2]
```diff
@@ -7,11 +7,11 @@ namespace :taxonomy do
     A JSON representation of the output is sent to STDOUT and can be
     redirected to a file if needed.
   DESC
-  task :bulk_tag_document_type, %i[document_type taxon_base_path] => :environment do |_, args|
+  task :bulk_tag_document_type, %i[document_type taxon_content_id] => :environment do |_, args|
```
tomsabin: I'm not sure where this rake task is used. Do we need to change the rake task parameters in a job somewhere or is this a developer task to manually run this rake task?

app/services/bulk_tagging/document_type_tagger.rb
cbaines: You can actually link to things which don't exist, but this is probably a good feature. A comment saying why this is there might be helpful though?

lib/tasks/taxonomy/bulk_tag_document_type.rake
koetsier: This is something a developer does. We are not expecting this type of bulk tagging to happen a lot (if at all) after doing the list on the ticket, so we decided it should be a rake task instead of cluttering up the Content tagger UI

PR #551, opened 2017-12-08
Add home page taxon to the taxonomy visualisation

koetsier:


cbaines:
`s/pgae/page` in the commit message.

app/models/taxonomy/expanded_taxonomy.rb
```diff
@@ -80,11 +85,24 @@ def expand_parent_nodes(start_node:, parent:)
...
+    end
+
+    def home_page_linked_content_item
       GovukTaxonomyHelpers::LinkedContentItem.new(
-        internal_name: content_item.dig('details', 'internal_name'),
-        title: content_item.fetch('title'),
-        base_path: content_item.fetch('base_path'),
-        content_id: content_item.fetch('content_id')
+        internal_name: 'Root of the taxonomy',
+        title: 'Root Taxon',
```
cbaines: Personally I'd call this the "Homepage", but I guess that's something that Sona might have an opinion on as well?

cbaines:
Looks really good :)

Some information on the "why" would be useful in the first two commits, as well as the what. Doesn't have to be lots of information.

PR #566, opened 2017-12-13
Move the visualisations from the branch page to the taxon page

cbaines:
This ended up being a bit more unwieldy than the branch name might suggest.

This branch includes 3 changes:

 - The move of the visualisations
 - The change to the projects interface
 - A implementation change to the TaxonomySize class (renamed to TaxonsWithContentCount)

tagathon-participant-view-of-taxon-page
taxon-page-bubbles-visualisation
taxon-page-tree-visualisation


app/controllers/branches_controller.rb
```diff
@@ -1,24 +1,7 @@
...
-
-    respond_to do |format|
-      format.html
-      format.json do
-        render json: @taxonomy_size.nested_tree
-      end
-    end
-  end
-
   def edit_all
```
tomsabin: (Comment slightly outside of the scope of this original PR...)

Perhaps this is an opportunity to clarify the `edit_all` and `update_all` actions too: when we're editing the branches, it is the single resource of the homepage. We're not _editing all_ the branches, but instead editing how branches are linked to a single content item. Perhaps one for another PR, but I just wanted to bring it up for discussion.

cbaines: I'm hoping that the work that Jos is doing as part of https://trello.com/c/TQgjro97/260-improve-the-management-of-level-1-taxons-in-content-tagger will remove these actions, and then we can potentially get rid of this entire controller, which would be good, as I agree that it's not very clear.

cbaines:
> This looks good to me.

Great :)

> The only suggestion I could make is if we could change the taxon links in the "List" visualisation to to match the "Taxonomy" visualisation: linking to /taxons/:uuid instead of /tagging_history/:uuid, now that we have the tagging history link in the list of actions.

Yeah, that's a good suggestion. I'm going to try and think a bit more about these navigation and information layout things...

PR #568, opened 2017-12-15
Restrict taxon routes to taxon content type

tomsabin:
Trello: https://trello.com/c/qKMvPNau/258-restrict-the-content-which-content-tagger-can-view-and-delete

### DIFF
--- a/app/services/taxonomy/build_taxon.rb
+++ b/app/services/taxonomy/build_taxon.rb
@@ -3,6 +3,7 @@ module Taxonomy
     attr_reader :content_id
 
     class TaxonNotFoundError < StandardError; end
+    class DocumentTypeError < StandardError; end
 
     def initialize(content_id:)
       @content_id = content_id
@@ -13,6 +14,8 @@ module Taxonomy
     end
 
     def build
+      validate_taxon_response!
+
       Taxon.new(
         content_id: content_id,
         title: content_item["title"],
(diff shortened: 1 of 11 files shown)


app/services/taxonomy/build_taxon.rb
```diff
@@ -3,6 +3,7 @@ class BuildTaxon
     attr_reader :content_id
 
     class TaxonNotFoundError < StandardError; end
+    class ContentTypeError < StandardError; end
```
cbaines: `DocumentTypeError` might be more fitting here, as the field being checked is the `document_type`?

app/services/taxonomy/build_taxon.rb
```diff
@@ -44,6 +47,10 @@ def content_item
       raise(TaxonNotFoundError, e.message)
     end
 
+    def validate_taxon_response!
+      raise(ContentTypeError) if content_item["document_type"] != "taxon"
```
cbaines: Given we also want to display the homepage, and that has a document type of `homepage`, this might need to be a little more permissive?

tomsabin: Ah, yes! I totally overlooked the different homepage document type. Thanks for bringing it up

app/services/taxonomy/build_taxon.rb
tomsabin: Updated in https://github.com/alphagov/content-tagger/pull/568/commits/b3a255810bd5e4b18ed85d1ade3b2a0b64059b30

app/services/taxonomy/build_taxon.rb
tomsabin: Added in https://github.com/alphagov/content-tagger/pull/568/commits/b3a255810bd5e4b18ed85d1ade3b2a0b64059b30

app/services/taxonomy/build_taxon.rb
cbaines: Great :)

app/services/taxonomy/build_taxon.rb
cbaines: Looks good 👍

PR #578, opened 2017-12-21
Use new root taxons

koetsier:
Root taxons are no longer represented as a list of 'root_taxons' link under the home page content_item but as a reverse link with 'level_one_taxons' link on the root (homepage) content item and 'root_taxon' link on the child taxons.

This PR enables content tagger to use the new links


app/views/taxons/_form_fields.html.erb
```diff
@@ -8,7 +8,7 @@
       value: page.taxon.path_slug.presence || page.taxon.path_prefix
     } %>
 
-  <div class="base-path-hint"'>
```
cbaines: Spotting things like this is great, but given this is unrelated, being in a separate commit would be even better.

app/services/taxonomy/parent_update.rb
```diff
@@ -0,0 +1,20 @@
+module Taxonomy
+  class ParentUpdate
+    def set_parent(content_id, parent_taxon_id: nil, associated_taxon_ids: [])
```
cbaines: Looking at the overall change, this class/method is only used from the UpdateTaxon service. Neither of these services are very big, which makes me think that having this logic in the UpdateTaxon service might be simpler, and a little less difuse. What do you think?

app/services/taxonomy/taxonomy_root_retrieval.rb
```diff
@@ -0,0 +1,8 @@
+module Taxonomy
+  class TaxonomyRootRetrieval
```
cbaines: I'm not sure what I think about the name here, while I guess this retrieves the homepage, which is the "Root" of the taxonomy, what it actually returns is the level one taxons. So maybe "LevelOneTaxons" would be a more fitting service name?

app/services/taxonomy/parent_update.rb
koetsier: Services in Ruby I believe are mostly single public method objects that do one thing (a lot of objects in content-tagger use a single 'call' method). It is not very big, but it does abstract away the way the way the parent is set.

app/services/taxonomy/taxonomy_root_retrieval.rb
koetsier: Good point. I'll rename this to LevelOneTaxonsRetrieval - I think 'root' and 'levelone' is a bit inconsistent throughout the app now

app/views/taxons/_form_fields.html.erb
koetsier: 👍

PR #582, opened 2018-01-02
Resolve linting issues

tomsabin:
Trello: https://trello.com/c/E7YWU5Pl/287-resolve-linting-issues-in-the-content-tagger-codebase


cbaines:
If this does indeed get everything, I think the next thing to do is remove `--diff` here [1], as then we'll not get in to this situation again.

1: https://github.com/alphagov/content-tagger/blob/master/lib/tasks/lint.rake#L3

tomsabin:
The last linting issue resolution is at https://github.com/alphagov/content-tagger/pull/583. Once that is merged, we can change the lint rake task to run over the full codebase.

PR #584, opened 2018-01-02
Update lint task to run against all files

tomsabin:
Now that we have resolved all of the linting issues, we can change our
local lint rake task to run against all files.

However, our CI build will still run in the diff mode as documented
here [1].

[1] https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy#L455-L473

### DIFF
--- a/lib/tasks/lint.rake
+++ b/lib/tasks/lint.rake
@@ -1,4 +1,4 @@
-desc "Run govuk-lint on files changed since origin/master"
+desc "Run govuk-lint on all files"
 task "lint" do
-  system "govuk-lint-ruby --diff"
+  system "govuk-lint-ruby"
 end


cbaines:
> However, our CI build will still run in the diff mode as documented
> here [1].

This is a little irritating, but I don't think it's worth working out how to get Jenkins to not use --diff, unless you especially want to.

cbaines:
> This is a little irritating, but I don't think it's worth working out how to get Jenkins to not use --diff, unless you especially want to.

At least not at the moment...

tomsabin:
I don't think we'll be needing to remove that option right now.

PR #591, opened 2018-01-03
Fix tree display of the root taxon.

koetsier:
The tree visualisation of the taxonomy will now only show taxons
to be attached to the 'Root of the Taxonomy' if this is actually
the case.

Trello: https://trello.com/c/RN7uU0EM/299-dont-show-the-root-of-taxonomy-in-the-taxonomy-visualisation-if-this-isnt-the-case

### DIFF
--- a/spec/models/taxonomy/expanded_taxonomy_spec.rb
+++ b/spec/models/taxonomy/expanded_taxonomy_spec.rb
@@ -24,7 +24,13 @@ RSpec.describe Taxonomy::ExpandedTaxonomy do
   end
 
   # parent taxons
-  let(:food) { fake_taxon("Food") }
+  let(:food) do
+    fake_taxon("Food").merge(
+      "links" => {
+        "root_taxon" => [GovukTaxonomy::ROOT_CONTENT_ID]
+      }
+  )
+  end
   let(:fruits) do
     fake_taxon("Fruits").merge(
       "links" => {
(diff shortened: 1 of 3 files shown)


spec/models/taxonomy/expanded_taxonomy_spec.rb
```diff
@@ -40,8 +46,44 @@ def home_page
...
+        },
+      )
+    end
+
+    describe 'Rootless taxonomies' do
+      it 'has one parent, which has no root' do
+        taxonomy = Taxonomy::ExpandedTaxonomy.new(@rootless_child['content_id']).build
+        expect(taxonomy.parent_expansion.children.map(&:internal_name)).to match_array(['i-Rootless Parent'])
+      end
+      it 'has no parents' do
```
cbaines: These tests are quite hard to read, as while the tests are in the same describe block, the object under test is different (so `it` is a different thing in the first test to the second test). Perhaps two describe blocks could be used?

spec/models/taxonomy/expanded_taxonomy_spec.rb
```diff
@@ -40,8 +46,44 @@ def home_page
...
+        expanded_links: {
+          parent_taxons: [@rootless_parent],
+        },
+      )
+    end
+
+    describe 'Rootless taxonomies' do
+      it 'has one parent, which has no root' do
+        taxonomy = Taxonomy::ExpandedTaxonomy.new(@rootless_child['content_id']).build
+        expect(taxonomy.parent_expansion.children.map(&:internal_name)).to match_array(['i-Rootless Parent'])
```
cbaines: It took me a while to figure out what `'i-Rootless Parent'` is, it might be clearer to say `@rootless_parent.dig('details', 'internal_name')` (or whatever the correct code is), and perhaps easier to compare the `content_id`?

spec/models/taxonomy/expanded_taxonomy_spec.rb
```diff
@@ -77,7 +119,7 @@ def home_page
 
   describe 'Ask for the Homepage' do
     before :each do
-      @taxonomy = ExpandedTaxonomy.new(GovukTaxonomy::ROOT_CONTENT_ID).build
+      @taxonomy = Taxonomy::ExpandedTaxonomy.new(GovukTaxonomy::ROOT_CONTENT_ID).build
```
cbaines: I'm not sure why this is changing?

spec/models/taxonomy/expanded_taxonomy_spec.rb
koetsier: changed it to use two blocks with some more description

spec/models/taxonomy/expanded_taxonomy_spec.rb
koetsier: Changed to use @rootless_child['content_id'] to compare

spec/models/taxonomy/expanded_taxonomy_spec.rb
koetsier: I believe these changes are necessary because the code used to be enclosed in a module statement and now uses RSpec.describe Taxonomy::ExpandedTaxonomy to satisfy the Linters maximum module length.

I'll put these changes in a separate commit

cbaines: Awesome, thanks 👍

spec/models/taxonomy/expanded_taxonomy_spec.rb
cbaines: Good stuff :)

spec/models/taxonomy/expanded_taxonomy_spec.rb
cbaines: 👍

PR #599, opened 2018-01-05
Fix bulk publishing

cbaines:
The interface for the taxonomy_partial changed, but this view wasn't
updated.


cbaines:
I'm working on writing a test to cover this feature, but I'm really slow at writing tests with Rails, so I may be some time...

cbaines:
> I'm working on writing a test to cover this feature, but I'm really slow at writing tests with Rails, so I may be some time...

I've now added a second commit with a test.

PR #603, opened 2018-01-08
Change behaviour for users with no permissions

cbaines:
Signon has a concept of access (which means the signin permission),
and other permissions (which are "GDS Editor" and "Tagathon
participant" in the case of Content Tagger).

Previously, just giving a user access to Content Tagger in Signon,
would not have given them any access to the app. This commit changes
that behaviour, and gives them access to see the taxonomy, but not
make any changes through Content Tagger.

The aim here is to neaten up the user experience. Now that it's
possible for a user of Content Tagger to have access to see the
taxonomy, but not to change it, this makes a good default experience
for users without other permissions, but who do have access to the
app.

For a user who just has access to Content Tagger, but no additional
permissions, the effect will be as follows.

# Before
before

# After
after

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,7 +4,6 @@ class ApplicationController < ActionController::Base
   include GDS::SSO::ControllerMethods
   before_action :authenticate_user!
   before_action :set_authenticated_user_header
-  before_action :ensure_user_can_use_application!
 
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
@@ -15,6 +14,8 @@ class ApplicationController < ActionController::Base
       redirect_to lookup_taggings_path
     elsif user_can_access_tagathon_tools?
       redirect_to projects_path
+    else
+      redirect_to taxons_path
     end
   end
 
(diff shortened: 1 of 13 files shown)


tomsabin:
This is good, the only comment I have to make is the availability of actions that lead to pages with "You are not authorised to ..." and some incorrect copy. I've highlighted these on two screenshots:

screen shot 2018-01-08 at 16 20 34
screen shot 2018-01-08 at 16 21 44

- "Edit taxonomy" - this doesn't correctly reflect what the user is seeing (in the breadcrumb?)
- "Download published taxonomy as CSV" - the user with no permissions sees an unauthorised page if they click this
- "..., edit the taxon or move its content" is also not correct for users with no permissions
- "View tagging history" - leads to unauthorised page
- "Edit tagging" - also leads to unauthorised page

I'll leave it up to you whether you want to address these at a later date as they shouldn't block this PR from being merged.

cbaines:
> This is good, the only comment I have to make is the availability of actions that lead to pages with "You are not authorised to ..." and some incorrect copy. I've highlighted these on two screenshots:

Thanks for taking a look Tom :) I've added some more commits that should fix the issues you highlighted.

I've also taken an updated screenshot:

after2

PR #605, opened 2018-01-09
Add Dockerfile for content-tagger app

adrianclay:
Is required for running publisher under publishing-e2e-tests.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+FROM ruby:2.4.2
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential nodejs && apt-get clean
+
+ENV DATABASE_URL postgresql://postgres@postgres/content-tagger
+ENV GOVUK_APP_NAME content-tagger
+ENV PORT 3116
+ENV RAILS_ENV development
+ENV REDIS_HOST redis
+
+ENV APP_HOME /app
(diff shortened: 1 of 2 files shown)


cbaines:
> Is required for running publisher under publishing-e2e-tests.

You probably want to say Content Tagger, rather than publisher.

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM ruby:2.4.2
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential nodejs && apt-get clean
+
+ENV DATABASE_URL postgresql://postgres@postgres/content_tagger_development
```
kevindew: shall we drop development from this? (since we try keep that portable)

db/seeds.rb
```diff
@@ -0,0 +1,13 @@
...
+end
+
+gds_organisation_id = "af07d5a5-df63-4ddc-9383-6a666845ebe9"
+
+User.create!(
+  uid: SecureRandom.uuid,
+  name: "Test user",
+  permissions: ["signin", "GDS Editor", "Tagathon participant"],
+  organisation_content_id: gds_organisation_id,
+)
```
cbaines: You're missing a newline here.

adrianclay: Doh, new IDE needed reconfiguring.

PR #609, opened 2018-01-11
Add statsd metrics for number of taxons per level

koetsier:
Trello: https://trello.com/c/fNMNMaiz/277-visualise-number-of-taxons-at-each-level-in-grafana

### DIFF
--- a/app/services/metrics/taxons_per_level_metrics.rb
+++ b/app/services/metrics/taxons_per_level_metrics.rb
@@ -1,0 +1,17 @@
+require_relative '../metrics'
+
+module Metrics
+  class TaxonsPerLevelMetrics
+    def count_taxons_per_level
+      Taxonomy::TaxonomyQuery.new.taxons_per_level.to_enum.with_index(1).each do |taxons, level|
+        gauge("level_#{level}.number_of_taxons", taxons.length)
+      end
+    end
+
+  private
+
+    def gauge(stat, value)
+      Metrics.statsd.gauge(stat, value)
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


app/services/metrics/taxonomy_metrics.rb
```diff
@@ -0,0 +1,17 @@
+require_relative '../metrics'
+
+module Metrics
+  class TaxonomyMetrics
```
cbaines: All the metrics in this namespace are currently about the taxonomy, so maybe this class could be called something more specific, TaxonsPerLevelMetrics perhaps?

spec/services/metrics/taxonomy_metrics_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+require 'gds_api/test_helpers/content_store'
+
+include ::GdsApi::TestHelpers::ContentStore
+
+module Metrics
+  RSpec.describe TaxonomyMetrics do
+    describe '#count_taxons_per_level' do
+      before do
+        content_store_has_item('/', root_taxon.to_json, draft: true)
+        content_store_has_item('/taxons/root_taxon', multi_level_child_taxons.to_json, draft: true)
```
cbaines: Might be good to use the newer "level one" terminology, especially given you went to the effort of making that change elsewhere.

app/services/metrics/taxonomy_metrics.rb
koetsier: fixed

spec/services/metrics/taxonomy_metrics_spec.rb
koetsier: fixed

PR #613, opened 2018-01-12
Rename FactoryGirl->FactoryBot

koetsier:
FactoryGirl is now called FactoryBot

see also: https://github.com/thoughtbot/factory_bot/issues/1053

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -80,10 +80,10 @@ GEM
       unf (>= 0.0.5, < 1.0.0)
     erubi (1.7.0)
     execjs (2.7.0)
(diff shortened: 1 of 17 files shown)


Gemfile.lock
```diff
@@ -80,10 +80,10 @@ GEM
       unf (>= 0.0.5, < 1.0.0)
     erubi (1.7.0)
     execjs (2.7.0)
-    factory_girl (4.9.0)
+    factory_bot (4.8.2)
```
cbaines: It's a bit odd that we would be downgrading as part of this change?

PR #624, opened 2018-01-24
Remove data export methods and supporting classes

koetsier:
Trello: https://trello.com/c/xmHpofxM/35-rewrite-the-data-extraction-part-of-the-pipeline-in-python


cbaines:
I've rebased this, hopefully that will cause the Publishing E2E tests to run...

tomsabin:
The codebase has moved on since this was first opened, and we're also now using the TaxonomyQuery#child_taxons outside of the classes touched in this branch. Perhaps we can remove the data export methods as they're now in the govuk-taxonomy-supervised-machine-learning repo.

Closing this PR for now, we should revisit this again with an updated version of the codebase.

PR #625, opened 2018-01-24
Add link to GOV.UK Homepage taxon from Edit Taxonomy page

tomsabin:
The only way to navigate to this taxon page before was to choose a
random taxon in the list (hoping to choose one that does exist within
the Topic Taxonomy) and then following the link in the tree
visualisation.

Now we've added a link to this taxon page so that it is more easily
accessible.

screen shot 2018-01-24 at 14 59 47

### DIFF
--- a/app/views/taxons/index.html.erb
+++ b/app/views/taxons/index.html.erb
@@ -1,11 +1,14 @@
 <%= display_header title: t("views.taxons.#{params[:action]}.title"), breadcrumbs: ['Taxons'] do %>
+  <%= link_to taxon_path(GovukTaxonomy::ROOT_CONTENT_ID), class: 'btn btn-md btn-default' do %>
+    View homepage taxon
+  <% end %>
+
   <% if user_can_administer_taxonomy? %>
     <%= link_to new_taxon_path, class: 'btn btn-default' do %>
       <i class="glyphicon glyphicon-plus"></i>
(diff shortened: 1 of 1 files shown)


app/views/taxons/index.html.erb
```diff
@@ -1,11 +1,14 @@
 <%= display_header title: t("views.taxons.#{params[:action]}.title"), breadcrumbs: ['Taxons'] do %>
+  <%= link_to taxon_path(GovukTaxonomy::ROOT_CONTENT_ID), class: 'btn btn-md btn-default' do %>
+    View homepage taxon
```
cbaines: My thinking is that the homepage is the homepage, and we treat it as a taxon. This could be mistaken for there being a homepage, and then a homepage taxon.

Anyway, not something to worry about too much, but hopefully over time the naming will sort itself out.

tomsabin: Interesting point, I didn't think of that. Perhaps we could have used the taxon's internal/external name instead (GOV.UK homepage) as well.

However, we probably want to be consistent with what this taxon is called, because now we have:

- homepage (as introduced in this PR)
- GOV.UK homepage (taxon internal & external name)
- Root of the taxonomy (the taxon description at the top of the visualisations).

I also thinking that we could be continuing with the "Topic Taxonomy" name dubbed in Q3, but that seems to have faded away..

PR #634, opened 2018-02-01
Change titles of world location taxons

h-lame:
For: https://trello.com/c/pNb4sEL3/56-2-change-titles-of-worldwide-help-and-services-navigation-pages

The titles we have are of the form `<location name> and the UK` but these
are the same titles as those of the world location news pages for those
locations.  It's hard to tell the difference so we want to rename the
taxons to `UK help and services in <location name>`.  This one-off rake
task does that, and also changes the internal name too so that what you
see in content-tagger matches up with what you see in the live site.

The list of country slugs in the rake task was obtained with the following
code in the whitehall rails console:

    WorldLocation.order(:slug).where(world_location_type_id: 1).map(&:slug)

There are a handful of entries that may not be live taxons, but the rake
task will report these for us to check.


h-lame:
Your instincts are correct, but unfortunately I don't think it can be because it's doing it all via the Publishing API, not a db.

PR #635, opened 2018-02-02
Taxon version history

tomsabin:
Trello: https://trello.com/c/eiPbWr0O/87-record-changes-made-to-a-taxon-in-content-tagger

To record the changes to a Taxon, we rely on calling out to the
publishing API before the update request is made. Using the "current"
version from the API and the "updated" version from the validated form
parameters, a change diff is generated and stored in the Versions table.

This approach was chosen as it allows us to record changes over time,
regardless of its publishing state. For taxons which are published, we
would be able to harness the Publishing API edition versioning, and step
back in time with the user_facing_version query parameter.

---

New taxon

> screen shot 2018-02-08 at 15 34 25

Changes to existing taxon

> screen shot 2018-02-08 at 15 31 25

### DIFF
--- a/app/views/taxons/version_history.html.erb
+++ b/app/views/taxons/version_history.html.erb
@@ -1,0 +1,28 @@
+<%= display_header title: page.title,
+      breadcrumbs: [:taxons, link_to(page.taxon.title, taxon_path(page.taxon.content_id)), 'Version history'] %>
+
+<% page.version_history.each do |version| %>
+  <div class="well well-sm">
+    <p>
+      <%= version.created_at.to_s(:govuk_date) %>
+      <span class="text-muted">(#<%= version.number %>)</span>
+    </p>
+
+    <% if version.note.present? %>
+      <p class="lead">
+        <%= version.note %>
+      </p>
+    <% end %>
+
+    <dl class="dl-horizontal">
+      <% version.changes.each do |attribute, (before, after)| %>
+        <dt>
+          <%= attribute %>
+        </dt>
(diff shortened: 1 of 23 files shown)


koetsier:
Looks good to me. Can you maybe add another screenshot so that the whole interaction is a bit clearer?

app/views/taxons/version_history.html.erb
```diff
@@ -0,0 +1,28 @@
...
+      </p>
+    <% end %>
+
+    <dl class="dl-horizontal">
+      <% version.changes.each do |attribute, (before, after)| %>
+        <dt>
+          <%= attribute %>
+        </dt>
+        <dd>
+          <code><%= before.inspect %></code> &rarr; <code><%= after.inspect %></code>
```
koetsier: Would this show 'nil' -> 'something' if the taxon is new? If so, maybe this can be more descriptive?

tomsabin: Yes, that's correct. I've added a screenshot to the PR description which shows this.

It was something I thought of while I was working on the versioning, but in the end I decided to leave it as is, because I don't believe it's a deal-breaker. I think we should get it out there and to be used before we start rethinking about how to display the data.

tomsabin:
Proposed changes after an offline conversation with @cbaines 

- Change text: "Version history for ..." to "History for" on history page
- Rename field: "Version history note" to "Internal change note"
- Display the fieldname with the internal change note when viewing the taxon history

PR #639, opened 2018-02-07
Specify the rails release in the migrations

leenagupte:
Rails 5 changed the way migrations are created. You'll have to specify 
the Rails release starting from Rails 5, otherwise you get this error:

`Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:`

See example migration in the [ruby docs](http://www.rubydoc.info/github/rails/rails/ActiveRecord/Migration)


db/migrate/20151115132026_create_users.rb
```diff
@@ -1,4 +1,4 @@
-class CreateUsers < ActiveRecord::Migration
+class CreateUsers < ActiveRecord::Migration[5.1]
```
tomsabin: Does Rails allow us to use 4.2 for this version, as that was the Rails version at this point in time? I believe that's the case for all of these migrations.

leenagupte: I believe it does. I picked Rails 5.1 because that's the version in the Gem file.

tomsabin: My understanding of these Rails-versions numbers is that it represents the version of Rails at the timestamp of that migration. Here's how I found the Rails version at the earliest (db/migrate/20151115132026_create_users.rb) and latest (20161019130634_remove_source_description_from_tag_migration) migration:

For this migration (the earliest), it was added in https://github.com/alphagov/content-tagger/commit/5a8ad96a93e38f6a17b75f9b6b7858fbff98c8fd, and the Rails version at this point in time is [4.2.5](https://github.com/alphagov/content-tagger/blob/5a8ad96a93e38f6a17b75f9b6b7858fbff98c8fd/Gemfile.lock#L117)

The latest migration (without the Rails-versioning) is 20161019130634_remove_source_description_from_tag_migration. which was added in https://github.com/alphagov/content-tagger/commit/52257953419e84e478781364e2c410bff124da4a, and the Rails version at this point in time is [4.2.7.1](https://github.com/alphagov/content-tagger/blob/52257953419e84e478781364e2c410bff124da4a/Gemfile.lock#L208)

tomsabin: This is what I dug out from the [Rails 5.0 release notes](http://guides.rubyonrails.org/5_0_release_notes.html):

> Version the API presented to migration classes, so we can change parameter defaults without breaking existing migrations, or forcing them to be rewritten through a deprecation cycle. ([Pull Request](https://github.com/rails/rails/pull/21538))

leenagupte: @tomsabin Thanks! I've updated and pushed.

PR #641, opened 2018-02-08
Download taxon branch

fmlharrison:
This PR is for the work I did changing Rake task to download a taxon tree as json to a file.

https://trello.com/c/mkTTLKkT/50-add-functionality-to-download-taxonomy-branch
https://trello.com/c/yhzHGgly/51-rake-task-to-download-taxon-branch-from-integration

### DIFF
--- a/app/services/taxonomy/taxon_tree_export.rb
+++ b/app/services/taxonomy/taxon_tree_export.rb
@@ -1,9 +1,18 @@
+require 'json'
+
 module Taxonomy
   class TaxonTreeExport
     attr_reader :taxon_content_id
 
     def initialize(content_id)
       @taxon_content_id = content_id
+    end
+
+    def build
+      taxonomy = expanded_taxon
+      taxonomy.build
+      flattened_taxonomy = flatten_taxonomy(taxonomy.child_expansion)
+      generate_json(flattened_taxonomy)
     end
 
     def expanded_taxon
@@ -52,6 +61,10 @@ module Taxonomy
     end
 
   private
+
+    def generate_json(taxonomy)
+      JSON.generate(taxonomy)
+    end
 
     def content_struct(content_id)
       OpenStruct.new(Services.publishing_api.get_content(content_id).to_h)
--- a/app/services/taxonomy/taxon_tree_export.rb
+++ b/app/services/taxonomy/taxon_tree_export.rb
@@ -11,6 +11,46 @@ module Taxonomy
       Taxonomy::ExpandedTaxonomy.new(top_taxon.content_id)
     end
 
+    def flatten_taxonomy(taxon)
+      flattened_taxonomy = [convert_to_content_item(taxon)]
+
+      if taxon.children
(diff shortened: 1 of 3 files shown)


app/services/taxonomy/taxon_tree_export.rb
```diff
@@ -0,0 +1,73 @@
...
+    attr_reader :taxon_content_id
+
+    def initialize(content_id)
+      @taxon_content_id = content_id
+    end
+
+    def build
+      taxonomy = expanded_taxon
+      taxonomy.build
+      flattened_taxonomy = flatten_taxonomy(taxonomy.child_expansion)
```
cbaines: Unfortunately, Content Tagger is a bit of a messy codebase. While the `ExpandedTaxonomy` class does kind of have the information you're interested, it looks like you might be better off here using the `TaxonomyQuery.new.child_taxons` method. That does the flattening of the tree structure of the expanded links, and I think the data might be in a more useful format as well.

leenagupte: @cbaines I think the problem is the amount of data that `TaxonomyQuery.new.child_taxons` returns. At the moment we're only interested in the direct children of a taxon, not all of the descendants.

The rake task `taxononmy:export:to_json` gives us the data we want in a format we can use, but rather than printing the output, we want to download it to a file. We could have updated the existing task, but decided to create a new one and share the code.

fmlharrison: @cbaines Thanks for the tip but as @leenagupte said, the data that we currently get after running the rake task is what we need.  I did try using the ```TaxonomyQuery.child_taxon``` but the data wasn't quite what we needed.

fmlharrison: @cbaines and @leenagupte Are you guys happy to approve this PR?

cbaines: I don't mind these changes, but it would be good to get some clarity on what has been done, and what the intent is.

> @cbaines I think the problem is the amount of data that TaxonomyQuery.new.child_taxons returns. At the moment we're only interested in the direct children of a taxon, not all of the descendants.

Ok, but this is not what the previous rake task did, and as far as I can tell, the behaviour with respect to the taxons which are outputted hasn't changed. With the changes in this branch, as far as I can tell, both of the rake tasks will still output the full list of descendants.

fmlharrison: @cbaines You are right. The currently functionality of the previous rake is unchanged. We just moved out the code out into a module so we could re-use it in our rake task.  
There was a mistake in the previous comment. Getting the full descendants of a taxon is fine (which is the current functionality of the ```TaxonTreeExport```) and the data we get back is in a form very similar to what will be returned from the content store if we called it directly from our prototype (which we are doing for the published taxons, education and childcare) which what we are looking for.

cbaines: Ok, thanks for the clarification.

PR #643, opened 2018-02-13
Order the project items the same way as the spreadsheet

cbaines:
Order by the id field, as this will reflect the order in which the
items were imported from the spreadsheet.

### DIFF
--- a/app/queries/project_filter_query.rb
+++ b/app/queries/project_filter_query.rb
@@ -27,7 +27,9 @@ class ProjectFilterQuery
 
     case current_filter
     when FILTER_TODO
-      items = items.todo
+      items = items
+                .todo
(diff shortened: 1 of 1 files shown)


app/queries/project_filter_query.rb
```diff
@@ -19,7 +19,7 @@ def items
     items = @project
       .content_items
       .with_valid_ids
-      .order(updated_at: :desc)
+      .order(id: :asc)
```
tomsabin: I think we wanted to change the order of the to-do view only, so we could change the order in the case statement below instead.

cbaines: That sounds fine, I've updated the commit.

PR #653, opened 2018-02-20
Modify taxon phases

tomsabin:
- Allow a taxon phase (alpha, beta, live) to be changed from the edit form.
- New action to bulk change the phase for a taxon and all its decedents

Trello: https://trello.com/c/YWruwhl9/137-change-content-tagger-to-include-phases-for-taxons

### DIFF
--- a/app/services/taxonomy/bulk_publish_taxon.rb
+++ b/app/services/taxonomy/bulk_publish_taxon.rb
@@ -1,5 +1,5 @@
 module Taxonomy
-  class BulkUpdateTaxon
+  class BulkPublishTaxon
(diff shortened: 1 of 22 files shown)


app/services/taxonomy/bulk_publish_taxon.rb
```diff
@@ -1,5 +1,5 @@
 module Taxonomy
-  class BulkUpdateTaxon
+  class BulkPublishTaxon
```
cbaines: I'm guessing the why here is that you want to add another different service with this name?

PR #656, opened 2018-02-23
Add task to check taxon base path structure

tomsabin:
The task iterates down the taxonomy and checks against each taxon that
it is following the defined taxon URL structure [1].

[1] https://gov-uk.atlassian.net/wiki/spaces/FS/pages/127664172/URL+structure+for+the+new+navigation+pages

### DIFF
--- a/app/services/taxon_base_path_structure_check.rb
+++ b/app/services/taxon_base_path_structure_check.rb
@@ -1,0 +1,71 @@
+class TaxonBasePathStructureCheck
+  LEVEL_ONE_URL_REGEX = %r{^\/([A-z0-9\-]+)$}
+  PATH_COMPONENTS_REGEX = %r{^\/(?<prefix>[A-z0-9\-]+)(\/(?<slug>[A-z0-9\-]+))?$}
+
+  attr_reader :path_validation_output, :invalid_taxons
+
+  def initialize(level_one_taxons:)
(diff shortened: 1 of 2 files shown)


app/services/taxon_base_path_structure_check.rb
```diff
@@ -0,0 +1,71 @@
+class TaxonBasePathStructureCheck
+  LEVEL_ONE_URL_REGEX = %r{^\/([A-z0-9\-]+)$}
+  PATH_COMPONENTS_REGEX = %r{^\/(?<prefix>[A-z0-9\-]+)(\/(?<slug>[A-z0-9\-]+))?$}
+
+  attr_reader :path_validation_output, :invalid_taxons
+
+  def initialize(parent_taxons:)
+    @parent_taxons = parent_taxons
```
cbaines: It might be clearer to change `parent_taxons` to `level_one_taxons` here, as that's what's expected?

tomsabin: Yep, good point. I think I was half way through renaming :)

tomsabin:
I've rake this task on integration (https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/152/console), though it took a second attempt to successfully complete as the first failed due to a `GdsApi::TimedOutException` error.

PR #658, opened 2018-02-26
Add required fields for PUT content payload

tomsabin:
After testing this on integration, I missed a few required fields for the Publishing API request. Sentry error can be seen here: https://sentry.io/govuk/app-content-tagger/issues/361348692/

### DIFF
--- a/app/workers/update_taxon_worker.rb
+++ b/app/workers/update_taxon_worker.rb
@@ -1,12 +1,14 @@
 class UpdateTaxonWorker
   include Sidekiq::Worker
 
-  def perform(content_id, payload)
-    taxon = Services.publishing_api.get_content(content_id)
+  def perform(content_id, attributes)
+    taxon = Taxonomy::BuildTaxon.call(content_id: content_id)
+    taxon.phase = attributes.with_indifferent_access.fetch(:phase)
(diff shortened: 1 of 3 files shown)


app/workers/update_taxon_worker.rb
```diff
@@ -1,12 +1,14 @@
 class UpdateTaxonWorker
   include Sidekiq::Worker
 
-  def perform(content_id, payload)
-    taxon = Services.publishing_api.get_content(content_id)
+  def perform(content_id, attributes)
+    taxon = Taxonomy::BuildTaxon.call(content_id: content_id)
+    taxon.phase = attributes.with_indifferent_access.fetch(:phase)
```
cbaines: It doesn't seem particularly clear that this worker takes a hash of attributes, but then only uses the phase. Could the attributes hash be merged with the taxon?

tomsabin: Updated in https://github.com/alphagov/content-tagger/pull/658/commits/a7600ad79c2ca9ca4a4d2fc4cc0910db2d151a91

cbaines: :+1:

PR #659, opened 2018-02-26
Update taxon show page

tomsabin:
Trello: https://trello.com/c/YWruwhl9/137-change-content-tagger-to-include-phases-for-taxons

screen shot 2018-02-26 at 15 38 51

### DIFF
--- a/app/views/taxons/show.html.erb
+++ b/app/views/taxons/show.html.erb
@@ -17,17 +17,17 @@
   <tbody>
     <tr>
       <td>
-        <span class="label state-label--<%= page.publication_state_name %>">
+        <span class="label add-right-margin state-label--<%= page.publication_state_name %>">
           <%= page.publication_state_name %>
         </span>
       </td>
       <td>
-        <span class="label phase-label--<%= page.taxon.phase %>">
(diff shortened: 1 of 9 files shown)


tomsabin:
If it makes the reviewing any easier, Sona and I worked on this a few days ago, and Chris P. is also happy with the changes.

app/views/taxons/show.html.erb
```diff
@@ -1,108 +1,176 @@
...
+      <td>
+        <span class="label phase-label--<%= page.taxon.phase %>">
+          <%= page.taxon.phase %>
+        </span>
+      </td>
+      <td>
+        <span class="label label-default">
+          <%= page.tagged.count %>
+        </span>
+      </td>
```
cbaines: The layout of these spans seems to involve the whitespace on webkit browsers (e.g. Chromium), but doesn't seem to affect Firefox, meaning that the spans appear to touch.

tomsabin: Updated in https://github.com/alphagov/content-tagger/pull/659/commits/26b58d4de4247eff6f84fb19d9f7b26df4a0e906

PR #660, opened 2018-02-27
Find common ancestors

koetsier:
Untags content items with common ancestors:

Trello: https://trello.com/c/jwol8AXZ/141-remove-superfluous-ancestor-tags

Adds a rake task: `content:content_tagged_to_ancestor` that removes all
redundant common-ancestor taggings

The algorithm gets taxon 'paths' from the taxonomy tree using the expanded 
links method of the publishing api. A 'path' is a sequence of content_ids from 
the 'root taxon' to the taxon the content item is tagged to.

The algorithm recursively searches for paths that are contained in others. For 
example we may have paths:

Root->A->B->C
Root->A->B
Root->A->D

Here, the content item is tagged to taxon 'C' as well as its direct ancestor 'B'.  
Because being tagged to taxon 'C' implies it is also tagged to 'B' (and 'A' and 'Root')
the tagging to taxon 'B' is considered redundant and can be removed. The 
 `CommonAncestorFinder` method will return taxon 'B' as its result.

### DIFF
--- a/app/services/tagging/common_ancestor_finder.rb
+++ b/app/services/tagging/common_ancestor_finder.rb
@@ -1,0 +1,55 @@
+module Tagging
+  class CommonAncestorFinder
+    def self.call
+      new.find_all
+    end
+
+    def find_all
+      content_enum = Services.rummager.search_enum({ reject_content_store_document_type: Tagging.blacklisted_document_types,
+                                                     fields: %w[content_id title] }, page_size: 100)
+      filtered_content_enum = content_enum.lazy.select { |c| c.has_key?('content_id') }
+      all_results = filtered_content_enum.map do |content_hash|
+        content_id = content_hash['content_id']
+        title = content_hash['title']
+        expanded_links_hash = Services.publishing_api.get_expanded_links(content_id).to_h
+        {
+          content_id: content_id,
+          title: title,
+          common_ancestors: find_common_ancestors(taxon_paths(expanded_links_hash))
+        }
+      end
+      all_results.select { |result| result[:common_ancestors].present? }
+    end
+
+  private
+
+    def taxon_paths(content_hash)
+      taxon_path = lambda do |taxon_hash|
+        return [] if taxon_hash.nil?
+        taxon_path.call(taxon_hash.dig('links', 'parent_taxons', 0) ||
+                        taxon_hash.dig('links', 'root_taxon', 0)).append(taxon_hash['content_id'])
+      end
+
+      taxon_hashes = content_hash.dig('expanded_links', 'taxons') || []
+      taxon_hashes.map do |taxon_hash|
+        taxon_path.call(taxon_hash)
+      end
+    end
+
+    # Finds the common ancestors from a list of paths of one or more trees
+    # The paths are ordered from the root to leaf
+    def find_common_ancestors(paths, common_ancestors = [])
+      return common_ancestors if paths.empty?
+
+      # Remove paths with a unique root - these cannot have common ancestors
+      paths_grouped_by_root = paths.group_by(&:first)
+      paths_with_common_root = paths_grouped_by_root.select { |_, v| v.length > 1 }.values.flatten(1)
+
+      # Remaining paths with length one have a common ancestor, i.e. the common root
+      paths_with_length_one = paths_with_common_root.select { |v| v.length == 1 }
+      paths_with_length_greater_than_one = paths_with_common_root.select { |n| n.length > 1 }
+
+      find_common_ancestors(paths_with_length_greater_than_one.map { |n| n[1..-1] }, common_ancestors + paths_with_length_one.flatten)
+    end
+  end
+end
--- a/app/services/tagging/untagger.rb
+++ b/app/services/tagging/untagger.rb
@@ -1,0 +1,12 @@
+module Tagging
+  class Untagger
+    def self.call(content_id, taxon_content_ids)
+      new.untag(content_id, taxon_content_ids)
+    end
+
+    def untag(content_id, taxon_content_ids)
+      existing_taxons_ids = Services.publishing_api.get_links(content_id).dig('links', 'taxons')
+      Services.publishing_api.patch_links(content_id, links: { taxons: (existing_taxons_ids - taxon_content_ids) })
+    end
+  end
+end
--- a/lib/tasks/content_tagged_to_ancestors.rake
+++ b/lib/tasks/content_tagged_to_ancestors.rake
@@ -1,0 +1,11 @@
+namespace :content do
+  desc "Find all content that is tagged to a taxon and its direct ancestor and optionally removes redundant tagging (option: untag='untag')"
+  task :tagged_to_ancestor, [:untag] => :environment do |_, args|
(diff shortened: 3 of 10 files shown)


spec/support/taxon_helper.rb
```diff
@@ -0,0 +1,31 @@
...
+
+        return {} if head.nil?
+        links_hash =
+          if tail.empty?
+            {}
+          else
+            { "links" =>
+               {
+                 tail.length == 1 ? "root_taxon" : "parent_taxons" => [path_converter.call(tail)]
+               } }
```
cbaines: This formatting is rather unique, laying it out like the hash at the bottom of the file would be easier on the eyes in my opinion?

app/services/tagging/untagger.rb
```diff
@@ -0,0 +1,12 @@
+module Tagging
+  class Untagger
```
cbaines: Maybe CommonAncestorRemover (or CommonAncestorUntagger) would be a more informative name?

spec/services/tagging/untagger_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+RSpec.describe Tagging::Untagger do
+  it 'untags a taxon' do
+    content_id = "51ac4247-fd92-470a-a207-6b852a97f2db"
+    publishing_api_has_links(
+      "content_id" => content_id,
+      "links" => {
+        "taxons" => ['aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa', 'bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb']
+      },
+        )
```
cbaines: You've got some odd indentation here.

lib/tasks/content_tagged_to_ancestors.rake
```diff
@@ -0,0 +1,11 @@
+namespace :content do
+  desc "Find all content that is tagged to a taxon and its direct ancestor"
```
cbaines: This task looks like it does more than finding, this might need updating?

app/services/tagging/untagger.rb
koetsier: It does general untagging - not necessarily to do with common ancestors

spec/services/tagging/untagger_spec.rb
koetsier: fixed

lib/tasks/content_tagged_to_ancestors.rake
koetsier: fixed

app/services/tagging/untagger.rb
cbaines: Ok, that sounds fine.

lib/tasks/content_tagged_to_ancestors.rake
```diff
@@ -0,0 +1,11 @@
+namespace :content do
+  desc "Find all content that is tagged to a taxon and its direct ancestor and removes the redundant tagging"
+  task content_tagged_to_ancestor: :environment do
+    results = Tagging::CommonAncestorFinder.call
+    results.each do |result|
+      puts "Document #{result[:title]}, content_id: #{result[:content_id]} has common ancestors tagged to: "
+      puts result[:common_ancestors].inspect
+      Tagging::Untagger.call(content_hash['content_id'], result[:common_ancestors])
```
cbaines: `content_hash` looks to be undefined.

koetsier: good spotting! Fixed that

app/services/tagging/common_ancestor_finder.rb
```diff
@@ -0,0 +1,54 @@
+module Tagging
+  class CommonAncestorFinder
+    def self.call
+      new.find_all
+    end
+
+    def find_all
+      content_enum = Services.rummager.search_enum({ fields: %w[content_id title] }, page_size: 100)
```
cbaines: We could use the list of in scope document types here, as that would avoid considering the tags to the ~70,000 HMRC Manual Sections, plus other stuff?

app/services/tagging/common_ancestor_finder.rb
```diff
@@ -0,0 +1,54 @@
...
+      filtered_content_enum = content_enum.lazy.select { |c| c.has_key?('content_id') }
+      all_results = filtered_content_enum.map do |content_hash|
+        content_id = content_hash['content_id']
+        title = content_hash['title']
+        expanded_links_hash = Services.publishing_api.get_expanded_links(content_id).to_h
```
cbaines: At least for now, while most of the taxons are in draft, we'll need to use the `with_drafts` option to the expanded links endpoint to actually catch most of the tagging.

koetsier: the publishing api's default is to include drafts:
`  def get_expanded_links(content_id, with_drafts: true, generate: false)`

app/services/tagging/common_ancestor_finder.rb
koetsier: Fixed - had to copy the BLACKLIST array, because the export_data* files are deprecated and will be removed in future

app/services/tagging/common_ancestor_finder.rb
cbaines: Ah, ok, that's good :)

app/services/tagging/common_ancestor_finder.rb
cbaines: How about using this blacklist https://github.com/alphagov/content-tagger/blob/master/config/document_types_excluded_from_the_topic_taxonomy.yml ? That is still relevant for the metrics, so won't be removed any time soon. It's also a better list, as it excludes less document types.

koetsier: ok, used those instead

lib/tasks/content_tagged_to_ancestors.rake
```diff
@@ -0,0 +1,11 @@
+namespace :content do
+  desc "Find all content that is tagged to a taxon and its direct ancestor and removes the redundant tagging"
+  task content_tagged_to_ancestor: :environment do
+    results = Tagging::CommonAncestorFinder.call
+    results.each do |result|
+      puts "Document #{result[:title]}, content_id: #{result[:content_id]} has common ancestors tagged to: "
+      puts result[:common_ancestors].inspect
+      Tagging::Untagger.call(result[:content_id], result[:common_ancestors])
```
cbaines: It might be useful to have a way of running this, without it changing everything, just so we can manually check some of the things it would do before it does them. What do you think?

lib/tasks/content_tagged_to_ancestors.rake
```diff
@@ -0,0 +1,11 @@
+namespace :content do
+  desc "Find all content that is tagged to a taxon and its direct ancestor and removes the redundant tagging"
+  task content_tagged_to_ancestor: :environment do
```
cbaines: This task is in the `content` namespace, so you could probably lose that from the front of the task name, so it would be `content:tagged_to_ancestor` rather than `content:content_tagged_to_ancestor`.

PR #663, opened 2018-02-28
Update rake task to automatically fix invalid base paths

tomsabin:
Trello: https://trello.com/c/TcubFZOu/20-fix-taxons-with-incorrect-base-paths

There are ~4400 taxons with invalid taxon paths due to the work to bring taxons from topics, policy areas, etc. When importing, the base paths were created to match the tree structure (e.g.. `/imported-topic/topic/medicines-medical-devices-blood/medical-devices-regulation-safety`). However, this should match the URL structure of `/highest-level-taxon-name/taxon-name`, rather than `/level_1/level_2/level_3/...`.

There are two cases that we can automatically fix, one is the above, whereby we take the level one prefix (e.g. `/health-and-social-care` in the example above) and parameterise the title (e.g.`/medical-devices-regulation-and-safety`) to build a valid base path. It's likely that we'll run into errors from the Publishing API from the generated base paths that have already been taken, but that's okay, the aim of this task is to reduce the number of taxons with invalid base paths down from ~4400.

The second case is simpler, where the level one taxon base path has been updated, but the children's prefix has not. An example of this currently is the Employment taxon:

```
✅ /employment
❌    ├── /work/pensions-and-ageing-society
❌      ├── /work/older-people
❌      ├── /work/public-service-pensions
❌      ├── /pensions/automatic-enrolment-in-workplace-pensions
❌      ├── /pensions/state-pension-simplification
```


lib/tasks/taxonomy/validate_taxons_base_paths.rake
```diff
@@ -14,9 +14,20 @@ namespace :taxonomy do
     if base_path_checker.invalid_taxons.any?
       puts "-" * 36
 
-      puts "The following taxons do not follow the taxon URL structure:"
+      puts "The following taxons did not match the taxon URL structure. Attempting to fix this..."
```
cbaines: I think it would be useful to make this optional, and opt in, e.g. `rake taxonomy:validate_taxons_base_paths[and_fix]` just so we can still run this without it doing anything.

tomsabin:
Closing this for now until I produce a better implementation that will result in more unique base paths. After running this in integration, there was still ~1100 taxons that were failing validation.

PR #666, opened 2018-03-01
Record changes when bulk updating

tomsabin:


app/models/taxon.rb
```diff
@@ -50,7 +52,11 @@ def base_path=(base_path)
   end
 
   def base_path
-    @base_path ||= '/' + [path_prefix, path_slug].compact.join('/')
+    if [path_prefix, path_slug].any?
+      '/' + [path_prefix, path_slug].compact.join('/')
+    else
+      @base_path
+    end
```
tomsabin: This also fixes a UI problem that we've had for some time, e.g. for the following taxon that has a base path of `/imported-topic/topic/...`: https://content-tagger.integration.publishing.service.gov.uk/taxons/fd074d43-dca1-4bce-ab32-334e7a50d017

Development (this PR):
screen shot 2018-03-01 at 12 46 26

Integration (latest master):
screen shot 2018-03-01 at 12 46 37

app/workers/update_taxon_worker.rb
```diff
@@ -2,13 +2,16 @@ class UpdateTaxonWorker
...
 
   def perform(content_id, attributes)
-    taxon = Taxonomy::BuildTaxon.call(content_id: content_id)
-    taxon.assign_attributes(attributes)
+    previous_taxon = Taxonomy::BuildTaxon.call(content_id: content_id)
+    updated_taxon = previous_taxon.clone
+    updated_taxon.assign_attributes(attributes)
 
-    payload = Taxonomy::BuildTaxonPayload.call(taxon: taxon)
+    Taxonomy::SaveTaxonVersion.call(updated_taxon, 'Bulk update', previous_taxon: previous_taxon)
```
cbaines: It might be good to make the "Bulk update" parameter, a optional argument to this worker, so that it can be set to a more specific message where the worker is used. Just a suggestion, no need to implement this straight away.

PR #667, opened 2018-03-01
Automatically fix invalid base paths

tomsabin:
Instead of using the taxon title (https://github.com/alphagov/content-tagger/pull/663), which produced some taxons with little description (e.g. `/environment/english`), we plan to use the existing base path to produce the most accurately descriptive base path, with the side effect that they could be quite long paths (e.g. for the taxon earlier, it would be produced as `/environment/keeping-farmed-animals-cattle-deaths-english`).

A few other examples of what the update would produce:

```
/imported-browse/sale-goods-services-data
  └─ /business/sale-goods-services-data

/imported-browse/browse/driving/parking-public-transport-environment/environment-and-emissions
  └─ environment/driving-parking-public-transport-environment-environment-and-emissions

/imported-topic/land-management
  └─ /environment/land-management

/imported-topic/topic/environmental-management/environmental-permits/application-forms
  └─ /environment/environmental-management-environmental-permits-application-forms

/imported-policies/export-controls
  └─ /business/export-controls
```


tomsabin:
I've ran this integration (with the auto-fix param), and the task has successfully updated the ~4400 taxons (except level one taxons): https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/189/consoleFull

(currently) Verifying with another run (without the auto-fix param): https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/190/consoleFull

PR #674, opened 2018-03-06
Add a "Add a child taxon" button to the taxon page

cbaines:
This makes it easier to add a child for a specific taxon, without
having to find the parent by hand on the new taxon page.

To make this possible, support using the taxon_params when rendering
the new page, and don't require :taxon to be present, as this would
break the new page if a parent isn't prefilled.

https://trello.com/c/OcwX17Xe/135-add-child-taxon-to-taxon-page

add-child-taxon

### DIFF
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -20,7 +20,7 @@ class TaxonsController < ApplicationController
   end
 
   def new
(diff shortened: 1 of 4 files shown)


koetsier:
Adding a screenshot would be helpful!

cbaines:
> Adding a screenshot would be helpful!

No problem, I've added one to the description.

PR #675, opened 2018-03-06
Handle 404 in common_ancestor_finder

koetsier:


app/services/tagging/common_ancestor_finder.rb
```diff
@@ -11,12 +11,16 @@ def find_all
...
+        begin
+          expanded_links_hash = Services.publishing_api.get_expanded_links(content_id).to_h
+          {
+            content_id: content_id,
+            title: title,
+            common_ancestors: find_common_ancestors(taxon_paths(expanded_links_hash))
+          }
+        rescue GdsApi::HTTPNotFound
+          puts("Cannot find content with id: #{content_id}")
+        end
```
cbaines: What will this block evaluate to in the case of an exception? I'm not sure the select on line 21 will work without the result being a dictionary.

koetsier: oops...

PR #678, opened 2018-03-07
Replace the Save&Publish button on 'edit published taxon' with 'save draft' button

koetsier:
Related e2e PR: https://github.com/alphagov/publishing-e2e-tests/pull/223

Replace the Save&Publish button on the edit published taxon with a 'Save Draft' button that creates a new draft taxon which can optionally be published in the show taxon screen.

Bulk updating of taxons now also saves them as new draft editions, so they should be bulk-published manually after.

image

Trello: https://trello.com/c/OLoVUuaI/180-introduce-separate-save-and-publish-buttons-to-content-tagger

### DIFF
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -75,10 +75,6 @@ class TaxonsController < ApplicationController
     if taxon.valid?
       Taxonomy::UpdateTaxon.call(taxon: taxon, version_note: params[:internal_change_note])
 
-      if params[:publish_taxon_on_save] == "true"
-        Services.publishing_api.publish(taxon.content_id)
-      end
-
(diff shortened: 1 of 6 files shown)


cbaines:
I think this is heading in the right direction, but we'll need to be a bit more thorough with removing everywhere where Content Tagger publishes taxons if they have been published in the past.

The bulk updating of the phases is an example of this. If we make it possible to draft changes without publishing, it will be not a great experience for the bulk updating of the phases to publish drafted changes to taxons (which have been published in the past).

On minor point on the UI, is that it might be better to have the publish button on the taxon page, rather than the edit page, as that would allow you to publish without having to do in to the edit page. What do you think?

cbaines:
The changes look fine, some additional information in the commits about the "why" would be good.

PR #679, opened 2018-03-07
Add legacy taxons link to form

tomsabin:
Add field to Taxon edit field to manually add link to legacy taxons (e.g. mainstream browse, policy, policy areas, topics & subtopics).

Trello: https://trello.com/c/2n0C0KdR/151-map-legacy-taxons-to-merged-taxons

Field:
screen shot 2018-03-15 at 10 23 31


koetsier:
A screenshot would be useful!

tomsabin:
@koetsier I've added a screenshot of the field to the description

app/models/taxon.rb
```diff
@@ -57,6 +57,10 @@ def base_path=(base_path)
     @path_slug = path_components['slug']
   end
 
+  def legacy_taxons=(legacy_taxons)
+    @legacy_taxons = legacy_taxons.select(&:present?)
```
koetsier: is this necessary? not-present legacy taxons seem to be removed before calling this?, i.e. links_update:7

app/services/taxonomy/links_update.rb
```diff
@@ -0,0 +1,32 @@
+module Taxonomy
+  class LinksUpdate
```
cbaines: The "why" would be nice to have in this commit message, even though I can guess that this rename is a sensible preparation step for adding more responsibilities to this class, I think it's useful to make that explicit in the commit message.

PR #682, opened 2018-03-09
Prefill the base path prefix on page load

cbaines:
prefilled-base-path

### DIFF
--- a/app/assets/javascripts/parent-taxon-prefix-preview.js
+++ b/app/assets/javascripts/parent-taxon-prefix-preview.js
@@ -5,6 +5,7 @@
     this.start = function(el) {
       var $parentSelectEl = $(el).find('select.js-parent-taxon');
       var $pathPrefixEl = $(el).find('.js-path-prefix-hint');
+      var $taxonBasePathEl = $(el).find('#taxon_base_path');
 
       function getParentPathPrefix(callback) {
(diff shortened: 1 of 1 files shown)


cbaines:
> There's still a bit of view logic in the getParentPathPrefix function. Could we use the callback to do that work as well?

Good spot, as discussed, I've changed the first commit to fix this.

app/assets/javascripts/parent-taxon-prefix-preview.js
```diff
@@ -5,31 +5,43 @@
...
           } else {
             $pathPrefixEl.addClass('hidden');
             $pathPrefixEl.text('');
           }
         });
       }
+
+      getParentPathPrefix(function (path_prefix) {
+        if (path_prefix) {
+          $taxonBasePathEl.val('/'+ path_prefix + '/');
```
tomsabin: This value change seems to have an undesired effect on integration: for a valid base path, the base path field is being modified from `/education/school-places` to `/education/` on https://content-tagger.integration.publishing.service.gov.uk/taxons/4dafe1f8-593c-47da-a3b5-577d96feeee4/edit

cbaines: Yep, I was a little too enthusiastic, I've got a fix up here https://github.com/alphagov/content-tagger/pull/687

PR #694, opened 2018-03-16
Rename 'Root of the Taxonomy' to 'GOV.UK'

tomsabin:
Trello: https://trello.com/c/yX4GkBIg/149-be-able-to-copy-the-full-basepath-of-a-level-3-4-taxon-from-a-view-taxon-page-in-content-tagger#comment-5aabb9138c3b7f0146e5e747


cbaines:
The title of the homepage is "GOV.UK homepage". If that is ok, it might be neat and consistent to use the title from the content item instead?

tomsabin:
I've updated the text from 'GOV.UK' to 'GOV.UK homepage'

PR #698, opened 2018-03-19
Add rake task to populate topic taxons with legacy taxons.

benthorner:


lib/tasks/load_legacy_mappings.rake
```diff
@@ -0,0 +1,29 @@
...
+      raise "Lookup failed for #{topic_taxon}" unless topic_taxon_id
+      topic_taxon_links = api.get_links(topic_taxon_id)["links"]
+
+      legacy_taxon_ids = api.lookup_content_ids(base_paths: legacy_taxons)
+      missing_legacy_taxons = legacy_taxons - legacy_taxon_ids.keys
+      raise "Lookup failed for #{missing_legacy_taxons}" if missing_legacy_taxons.any?
+      topic_taxon_links["legacy_taxons"] = legacy_taxon_ids.values
+
+      puts "Adding #{legacy_taxons.count} legacy taxons for #{topic_taxon}"
+      api.patch_links(topic_taxon_id, links: topic_taxon_links, bulk_publishing: true)
```
cbaines: The Publishing API supports "optimistic locking" for these kind of requests. I don't think this is something we'd want to handle explicitly here, but I was wondering if we could add some block based thing that would handle fetching the links and changing them, while handling the `previous_version` thing to ensure that there are not any conflicts. Something like...
```
api.update_links(topic_taxon_id, bulk_publishing: true) do |links|
  links["legacy_taxons"] = legacy_taxon_ids.values
end
```
What do you think? I don't see this as something that is necessary to solve now, but it would be good to completely avoid the possibility of overriding any changes made concurrently in Content Tagger when running this script.

cbaines: I forgot the link to the docs https://docs.publishing.service.gov.uk/apis/publishing-api/api.html#optimistic-locking-previous_version

benthorner: I think supporting concurrent transactions is a good habit, so why not...done.

lib/tasks/load_legacy_mappings.rake
```diff
@@ -0,0 +1,29 @@
...
+      topic_taxon_id = api.lookup_content_id(base_path: topic_taxon)
+      raise "Lookup failed for #{topic_taxon}" unless topic_taxon_id
+
+      legacy_taxon_ids = api.lookup_content_ids(base_paths: legacy_taxons)
+      missing_legacy_taxons = legacy_taxons - legacy_taxon_ids.keys
+      raise "Lookup failed for #{missing_legacy_taxons}" if missing_legacy_taxons.any?
+
+      puts "Adding #{legacy_taxons.count} legacy taxons for #{topic_taxon}"
+      api.patch_links(topic_taxon_id,
+                      links: { legacy_taxons: legacy_taxon_ids.values },
```
cbaines: Looks good :+1:

PR #712, opened 2018-03-29
Tag content from a remote CSV file

koetsier:
Tags content from a remote CSV file, i.e. google sheets.

To access a remote CSV file, go to google sheets and select file->publish to web and copy the link

The rake task is: `rake tag_from_csv[remote_url]`

The task expects the CSV file to have headers and contain the fields 'content_id' and 'taxon_id'

Trello: https://trello.com/c/R3XBQbdz/200-tag-untagged-data

### DIFF
--- a/app/services/tagging/tagger.rb
+++ b/app/services/tagging/tagger.rb
@@ -1,0 +1,14 @@
+module Tagging
+  class Tagger
+    def self.add_tags(content_id, taxon_ids)
+      link_content = Services.publishing_api.get_links(content_id)
+      version = link_content['version']
+      taxons = (link_content.dig('links', 'taxons') || []) + taxon_ids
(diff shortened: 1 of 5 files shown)


app/services/tagging/tagger.rb
```diff
@@ -0,0 +1,14 @@
+module Tagging
+  class Tagger
+    def self.add_tags(content_id, taxon_ids)
+      link_content = Services.publishing_api.get_links(content_id)
+      version = link_content['version']
+      taxons = (link_content.dig('links', 'taxons') || []) + taxon_ids
```
cbaines: Maybe add a `.uniq` in here somewhere? That I think would make this idempotent... although saying that, I guess the Publishing API might do this already.

PR #732, opened 2018-04-19
[WIP] Due to alteration of the connected GovukTaxonomyHelpers method

1pretz1:
Change in the GovukTaxononomyHelpers gem means the 'from_publishing_api'
method has been merged with 'from_content_id', which means the
'build_child_expansion' function has been modified accordingly.
Functionality is still the same.


leenagupte:
@1pretz1  Can this change go out before https://github.com/alphagov/govuk_taxonomy_helpers/pull/16? If not, I'd add [WIP] to the title.

After the [govuk_taxonomy_helpers gem](https://github.com/alphagov/govuk_taxonomy_helpers) has been updated, the gem version number will need to be updated.

app/models/taxonomy/expanded_taxonomy.rb
```diff
@@ -118,9 +118,9 @@ def build_parent_expansion
     end
 
     def build_child_expansion
-      @child_expansion = GovukTaxonomyHelpers::LinkedContentItem.from_publishing_api(
-        content_item: root_content_item,
-        expanded_links: root_expanded_links
+      @child_expansion = GovukTaxonomyHelpers::LinkedContentItem.from_content_id(
+        content_id: @content_id,
+        publishing_api: Services.publishing_api,
```
leenagupte: It feels weird that you have to pass in a publishing_api object. 

It might be worth refactoring `from_content_id ` at some point so that you don't have to do that.

PR #755, opened 2018-06-04
Unpublishing a taxon tags all its content to its parent.

koetsier:
When unpublishing a taxon, a checkbox can be checked that automatically tags all its content to
its parent. 

Trello: https://trello.com/c/YrANYCMo/126-improve-the-unpublish-feature-in-content-tagger-to-also-handle-content

image

### DIFF
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -88,12 +88,14 @@ class TaxonsController < ApplicationController
   def destroy
     if content_id == GovukTaxonomy::ROOT_CONTENT_ID
       redirect_to taxon_path(content_id), danger: t("controllers.taxons.destroy_homepage")
-    elsif params[:taxon][:redirect_to].empty?
+    elsif params[:taxonomy_delete_page][:redirect_to].empty?
       flash.now[:danger] = t("controllers.taxons.destroy_no_redirect")
       render :confirm_delete, locals: { page: Taxonomy::DeletePage.new(taxon) }
     else
-      base_path = Services.publishing_api.get_content(params[:taxon][:redirect_to])['base_path']
-      Services.publishing_api.unpublish(params[:id], type: "redirect", alternative_path: base_path)
+      Taxonomy::TaxonUnpublisher.call(taxon_content_id: params[:id],
+                                      redirect_to_content_id: params[:taxonomy_delete_page][:redirect_to],
+                                      user: current_user,
+                                      retag: params[:taxonomy_delete_page][:do_tag] == '1')
       redirect_to taxon_path(taxon.content_id), success: t("controllers.taxons.destroy_success")
     end
   end
--- a/app/services/taxonomy/delete_page.rb
+++ b/app/services/taxonomy/delete_page.rb
@@ -1,11 +1,14 @@
 module Taxonomy
   class DeletePage
-    delegate :content_id, to: :taxon
+    include ActiveModel::Model
 
+    delegate :content_id, :redirect_to, to: :taxon
+    attr_reader :do_tag
     attr_reader :taxon
 
     def initialize(taxon)
       @taxon = taxon
+      @do_tag = true
     end
 
     def taxonomy_tree
--- a/app/services/taxonomy/taxon_unpublisher.rb
+++ b/app/services/taxonomy/taxon_unpublisher.rb
@@ -1,0 +1,41 @@
+module Taxonomy
+  class TaxonUnpublisher
+    def self.call(taxon_content_id:, redirect_to_content_id:, user:, retag: true)
+      new.unpublish(taxon_content_id: taxon_content_id,
+                    redirect_to_content_id: redirect_to_content_id,
+                    user: user, retag: retag)
+    end
+
+    def unpublish(taxon_content_id:, redirect_to_content_id:, user:, retag:)
+      tag_to_parent(taxon_content_id, user) if retag
+      unpublish_taxon(taxon_content_id, redirect_to_content_id)
+    end
+
+  private
+
+    def tag_to_parent(taxon_content_id, user)
+      parent_taxon = Taxonomy::TaxonomyQuery.new.parent(taxon_content_id)
+      return if parent_taxon.nil?
+      tag_migration = BulkTagging::BuildTagMigration.call(
+        source_content_item: ContentItem.find!(taxon_content_id),
+        taxon_content_ids: [parent_taxon['content_id']],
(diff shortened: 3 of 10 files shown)


app/services/taxonomy/taxonomy_query.rb
```diff
@@ -29,6 +29,12 @@ def content_tagged_to_taxons(content_ids, slice_size: 50)
       content_id_hashes.map { |h| h['content_id'] }.uniq
     end
 
+    def parent(content_id)
```
leenagupte: I feel that this functionality should already exist somewhere? A [taxon](https://github.com/alphagov/content-tagger/blob/master/app/models/govuk_taxonomy/taxon.rb) has a `parent_node` attribute, isn't that the same thing?

app/controllers/taxons_controller.rb
```diff
@@ -92,8 +92,7 @@ def destroy
       flash.now[:danger] = t("controllers.taxons.destroy_no_redirect")
       render :confirm_delete, locals: { page: Taxonomy::DeletePage.new(taxon) }
     else
-      base_path = Services.publishing_api.get_content(params[:taxon][:redirect_to])['base_path']
-      Services.publishing_api.unpublish(params[:id], type: "redirect", alternative_path: base_path)
+      UnpublishTaxonWorker.perform_async(params[:id], params[:taxon][:redirect_to])
```
cbaines: Why make this asynchronous? It makes it a lot harder to provide feedback to the user if this fails.

koetsier: Because it loops through a possibly large number of content items and tags them to the parent - this may involve a large number of calls to the publishing api. If I make this synchronous, the user may have to wait a long time.

spec/services/taxonomy/taxonomy_query_spec.rb
```diff
@@ -3,11 +3,25 @@
...
   end
+
+  describe '#parent' do
+    it 'returns nil because the taxon is a level one taxon' do
+      publishing_api_has_expanded_links('content_id' => 'f3bbdec2-0e62-4520-a7fd-6ffd5d36e03a', 'expanded_links' => {})
+      expect(query.parent('f3bbdec2-0e62-4520-a7fd-6ffd5d36e03a')).to be_nil
+    end
+    it 'returns a parent' do
+      publishing_api_has_expanded_links(expanded_links)
+      expect(query.parent('f3bbdec2-0e62-4520-a7fd-6ffd5d36e03a')).to eq('base_path' => "/path/base",
```
leenagupte: You're using the content_id of the root taxon again in this test.  I'd use another content_id to illustrate the point of the test.

app/controllers/taxons_controller.rb
cbaines: How about mirroring the manual approach, and using TagMigration stuff? This has progress reporting support, and works asynchronously.

spec/services/taxonomy/taxonomy_query_spec.rb
koetsier: fixed

app/workers/unpublish_taxon_worker.rb
```diff
@@ -0,0 +1,24 @@
+class UnpublishTaxonWorker
+  include Sidekiq::Worker
+
+  def perform(taxon_content_id, redirect_to_content_id)
```
leenagupte: There are two things happening here. Redirecting the taxon (to the parent taxon?) and re-tagging content to the parent taxon. These two operations seem like they are independent of each other.

To make the code easier to read, I would move the code into two separate functions that are called by `perform`.
Something like `retag_content_to_parent_taxon(taxon_content_id)` and `redirect_taxon(redirect_to_content_id)`.

app/controllers/taxons_controller.rb
leenagupte: @cbaines I'd be interested to see how this works.

koetsier: fixed

app/workers/unpublish_taxon_worker.rb
koetsier: fixed

app/services/taxonomy/delete_page.rb
```diff
@@ -1,11 +1,14 @@
...
-    delegate :content_id, to: :taxon
+    include ActiveModel::Model
 
+    delegate :content_id, :redirect_to, to: :taxon
+    attr_reader :do_tag
     attr_reader :taxon
 
     def initialize(taxon)
       @taxon = taxon
+      @do_tag = '1'
```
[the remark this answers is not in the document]
koetsier: fixed

app/services/taxonomy/delete_page.rb
```diff
@@ -1,11 +1,14 @@
 module Taxonomy
   class DeletePage
-    delegate :content_id, to: :taxon
+    include ActiveModel::Model
```
cbaines: I don't quite follow this, this doesn't seem like a model like thing to me?

koetsier: It is used as a view model. The do_tag checkbox is rendered using simple_form, which takes a (view model and so it uses this one instead of Taxon (which it used previously). All relevant attributes are delegated to the Taxon model

app/services/taxonomy/taxon_unpublisher.rb
```diff
@@ -0,0 +1,41 @@
...
+  private
+
+    def tag_to_parent(taxon_content_id, user)
+      parent_taxon = Taxonomy::TaxonomyQuery.new.parent(taxon_content_id)
+      return if parent_taxon.nil?
+      tag_migration = BulkTagging::BuildTagMigration.call(
+        source_content_item: ContentItem.find!(taxon_content_id),
+        taxon_content_ids: [parent_taxon['content_id']],
+        content_base_paths: tagged_content_base_paths(taxon_content_id)
+      )
```
cbaines: :+1:

app/services/taxonomy/taxon_unpublisher.rb
```diff
@@ -0,0 +1,41 @@
+module Taxonomy
+  class TaxonUnpublisher
+    def self.call(taxon_content_id:, redirect_to_content_id:, user:, retag: true)
+      new.unpublish(taxon_content_id: taxon_content_id,
```
leenagupte: Will the "tag to parent" checkbox be selected by default in the form? If it's not, it's feels odd that the default behaviour is true.

koetsier: yes

PR #756, opened 2018-06-05
Remove tbc in descriptions

peteglondon:
We want to prevent descriptions like '...' and 'tbc'.

In order to do this - this PR does the following:

* Remove validation of a taxon's description
* Add a rake task taxonomy:clear_descriptions which updates these descriptions ('...' and 'tbc') in the publishing api, setting them to null.

Trello: [Remove the ... / tbc in descriptions](https://trello.com/c/MjF3xk3a/113-remove-the-tbc-in-descriptions-2)

### DIFF
--- a/lib/taxon_description_updater.rb
+++ b/lib/taxon_description_updater.rb
@@ -30,6 +30,19 @@ private
     )
     publishing_api.put_content(content_id, payload)
     publishing_api.publish(content_id) unless draft_version_exists
+    log_what_happened(taxon, draft_version_exists)
+  end
+
+  def log_what_happened(taxon, draft_version_exists)
+    if draft_version_exists
+      puts "Draft taxon: #{taxon_details(taxon)} : description cleared but not published"
+    else
+      puts "Published taxon: #{taxon_details(taxon)} : description cleared and published"
+    end
+  end
+
+  def taxon_details(taxon)
+    "title: #{taxon['title']}, content_id: #{taxon['content_id']}"
   end
 
   def draft_version_exists?(taxon)
--- a/lib/taxon_description_updater.rb
+++ b/lib/taxon_description_updater.rb
@@ -1,0 +1,38 @@
+class TaxonDescriptionUpdater
+  def initialize(descriptions_to_remove)
+    @descriptions_to_remove = descriptions_to_remove
+  end
+
+  def call
+    @descriptions_to_remove.each(&method(:replace_description))
+  end
+
+private
+
+  EXCLUDE_ATTRIBUTES = %w[content_store user_facing_version publication_state lock_version updated_at state_history].freeze
+
+  def replace_description(description_to_remove)
+    content_items = publishing_api.get_content_items(
+      per_page: 5000,
+      q: description_to_remove,
+      search_in: ['description']
+    )
+    content_items['results'].select { |item| item['description'] == description_to_remove }.each do |taxon|
+      update_description(taxon)
+    end
+  end
+
+  def update_description(taxon)
+    content_id = taxon['content_id']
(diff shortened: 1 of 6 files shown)


lib/taxon_description_updater.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+  end
+
+  def update_description(taxon)
+    content_id = taxon['content_id']
+    payload = taxon.except(*EXCLUDE_ATTRIBUTES).merge(
+      'description' => nil,
+      'update_type' => 'minor'
+    )
+    publishing_api.put_content(content_id, payload)
```
cbaines: I'm a bit concerned about the possibility that there may be some taxons which have a draft edition, which doesn't match the live edition in some way (e.g. title, phase, ...) and that running this script would publish those changes.

Would it be possible to put some checking in place, to make sure that this only automatically changes taxons where no draft previously existed? Or that it checks a few of the important user editable fields match between the draft and published editions prior to publishing?

cbaines:
There looks to be a slight issue with the handling of unpublished taxons.

```
GdsApi::HTTPUnprocessableEntity: URL: http://publishing-api.dev.gov.uk:30080/v2/content/e1bc42df-d994-4e90-ba65-009b4d1fa12a
Response body:
[{"schema":{"scheme":null,"user":null,"password":null,"host":null,"port":null,"path":"e028160a-97bb-50ce-8003-1d97195ec583","query":null,"fragment":null},"fragment":"#/","message":"The property '#/' contains additional properties [\"unpublishing\"] outside of the schema when none are allowed in schema e028160a-97bb-50ce-8003-1d97195ec583","failed_attribute":"AdditionalProperties"}]

Request body:
{"base_path"=>"/transport/individual-vehice-approval", "description"=>nil, "details"=>{"internal_name"=>"Individual vehicle approval", "notes_for_editors"=>"", "visible_to_departmental_editors"=>false}, "document_type"=>"taxon", "first_published_at"=>"2018-03-08T16:24:14Z", "last_edited_at"=>"2018-03-08T14:32:48Z", "phase"=>"alpha", "public_updated_at"=>"2018-03-08T14:32:48Z", "publishing_app"=>"content-tagger", "redirects"=>[], "rendering_app"=>"collections", "routes"=>[{"path"=>"/transport/individual-vehice-approval", "type"=>"exact"}], "schema_name"=>"taxon", "title"=>"Individual vehicle approval", "update_type"=>"minor", "content_id"=>"e1bc42df-d994-4e90-ba65-009b4d1fa12a", "unpublishing"=>{"type"=>"redirect", "explanation"=>nil, "alternative_path"=>"/transport/vehicle-approval-and-standards", "redirects"=>[{"path"=>"/transport/individual-vehice-approval", "type"=>"exact", "destination"=>"/transport/vehicle-approval-and-standards"}], "unpublished_at"=>nil}, "locale"=>"en", "links"=>{}}
```

We probably don't want to attempt to change unpublished taxons.

lib/taxon_description_updater.rb
```diff
@@ -0,0 +1,55 @@
...
+    publishing_api.publish(content_id) unless draft_version_exists
+    log_what_happened(taxon, draft_version_exists)
+  end
+
+  def log_what_happened(taxon, draft_version_exists)
+    if draft_version_exists
+      p "Draft taxon: #{taxon_details(taxon)} : description cleared but not published"
+    else
+      p "Published taxon: #{taxon_details(taxon)} : description cleared and published"
```
cbaines: `p` uses `.inspect`, `puts` wouldn't print surrounding quotes in this case, but `p` is also fine.

lib/taxon_description_updater.rb
peteglondon: Now we only publish the description change if there is not an existing draft.

PR #757, opened 2018-06-06
Allow redirects to be used in tag migrations

koetsier:
When unpublishing a Taxon, we expect all content to be copied to its parent.
This allows redirected content to be copied.

### DIFF
--- a/app/models/bulk_tagging/tag_mapping.rb
+++ b/app/models/bulk_tagging/tag_mapping.rb
@@ -24,7 +24,7 @@ module BulkTagging
 
     def content_id
       @content_id ||=
-        Services.publishing_api.lookup_content_id(base_path: content_base_path)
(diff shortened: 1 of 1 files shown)


app/models/bulk_tagging/tag_mapping.rb
```diff
@@ -24,7 +24,7 @@ class TagMapping < ActiveRecord::Base
 
     def content_id
       @content_id ||=
-        Services.publishing_api.lookup_content_id(base_path: content_base_path)
+        Services.publishing_api.lookup_content_id(base_path: content_base_path, exclude_unpublishing_types: %w[vanish gone], exclude_document_types: ['gone'])
```
cbaines: I'd consider just making both empty arrays, as it might be good to just not care about the states. But I'm fine with this also.

PR #762, opened 2018-06-11
Fix a couple of issues with fetching email subscriber lists

barrucadu:
1. A missing list just means there are no subscribers, not that there has been an error.
2. The API adapters are silly and use string keys.

---

[Trello card](https://trello.com/c/3fxFsB93/201-add-subscriber-list-count-stats-to-taxons-in-content-tagger)

### DIFF
--- a/app/services/taxonomy/show_page.rb
+++ b/app/services/taxonomy/show_page.rb
@@ -70,15 +70,13 @@ module Taxonomy
 
     def email_subscribers
       @email_subscribers ||= begin
-        email_lists = {}
-
         begin
-          email_lists = Services.email_alert_api.find_subscriber_list(links: { taxon_tree: [taxon.content_id] })
+          email_lists = Services.email_alert_api.find_subscriber_list("links" => { taxon_tree: [taxon.content_id] })
+          email_lists.dig("subscriber_list", "active_subscriptions_count")
         rescue GdsApi::BaseError, SocketError => e
           GovukError.notify(e)
+          "?"
         end
-
-        email_lists.dig("subscriber_list", "active_subscriptions_count") || "?"
       end
     end
   end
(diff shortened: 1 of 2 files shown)


app/services/taxonomy/show_page.rb
```diff
@@ -73,7 +73,9 @@ def email_subscribers
         email_lists = {}
 
         begin
-          email_lists = Services.email_alert_api.find_subscriber_list(links: { taxon_tree: [taxon.content_id] })
+          email_lists = Services.email_alert_api.find_subscriber_list("links" => { taxon_tree: [taxon.content_id] })
+        rescue GdsApi::HTTPNotFound
+          email_lists = { "subscriber_list" => { "active_subscriptions_count" => 0 } }
```
thomasleese: I would be tempted to restructure this method into something like this:

```rb
begin
  email_lists = Services.email_alert_api.find_subscriber_list("links" => { taxon_tree: [taxon.content_id] })
  email_lists.dig("subscriber_list", "active_subscriptions_count")
rescue GdsApi::HTTPNotFound
  "0"
rescue GdsApi::BaseError, SocketError => e
  "?"
end
```

app/services/taxonomy/show_page.rb
```diff
@@ -70,15 +70,15 @@ def taxon_deletable?
...
-
         begin
-          email_lists = Services.email_alert_api.find_subscriber_list(links: { taxon_tree: [taxon.content_id] })
+          email_lists = Services.email_alert_api.find_subscriber_list("links" => { taxon_tree: [taxon.content_id] })
+          email_lists.dig("subscriber_list", "active_subscriptions_count")
+        rescue GdsApi::HTTPNotFound
+          "0"
         rescue GdsApi::BaseError, SocketError => e
           GovukError.notify(e)
+          "?"
```
Davidslv: what does it mean?

barrucadu: It means it can't talk to the email-alert-api.  I opted for "?" rather than a sentence, because there's not much space:

screen shot 2018-05-31 at 16 37 52

PR #770, opened 2018-06-26
Add published with new draft state

leenagupte:
Trello: https://trello.com/c/8PzlTDxX

Run `publishing_e2e_tests` against https://github.com/alphagov/publishing-e2e-tests/pull/244

## Motivation

Now that the draft edition can be independently maintained, the properties of the draft and published edition can be different. Currently, if there is a published and draft edition, only the details for the draft taxon are displayed, so it is hard to tell if a published version already exists.

### Before
screenshot-content-tagger integration publishing service gov uk-2018 06 26-12-09-19

screenshot-content-tagger integration publishing service gov uk-2018 06 26-12-08-12

#### New state "published with new draft"
* Shows links to "preview" and "published" topic pages
screenshot-content-tagger dev gov uk-2018 06 26-11-03-19

#### Existing state "published"
* Only shows link to published topic page
screenshot-content-tagger dev gov uk-2018 06 26-10-57-03

#### Existing state "draft"
* Only shows link to preview topic page
screenshot-content-tagger dev gov uk-2018 06 26-10-56-33

### DIFF
--- a/app/models/taxon.rb
+++ b/app/models/taxon.rb
@@ -6,6 +6,7 @@ class Taxon
     :description,
     :parent_content_id,
     :publication_state,
+    :state_history,
     :phase,
     :document_type,
     :redirect_to,
--- a/app/models/taxon.rb
+++ b/app/models/taxon.rb
@@ -38,6 +38,14 @@ class Taxon
     publication_state == "unpublished" && !redirect_to.nil?
   end
 
+  def ordered_publication_state_history
+    state_history.sort_by(&:first).map(&:second)
+  end
+
+  def lastest_two_publication_states
+    ordered_publication_state_history.last(2)
+  end
+
   def level_one_taxon?
     parent_content_id.present? &&
       parent_content_id == GovukTaxonomy::ROOT_CONTENT_ID
--- a/app/models/taxon.rb
+++ b/app/models/taxon.rb
@@ -38,6 +38,11 @@ class Taxon
     publication_state == "unpublished" && !redirect_to.nil?
   end
 
+  def draft_and_published_editions_exist?
+    previous_state, latest_state = lastest_two_publication_states
+    previous_state && latest_state == "draft"
+  end
+
   def ordered_publication_state_history
     state_history.sort_by(&:first).map(&:second)
   end
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -67,6 +67,10 @@ $red: #b10e1e;
   background-color: $gray-light;
 }
 
+.state-label--published_with_new_draft  {
+  background-color: #2B8CC4;
+}
+
 .field-with-error {
   border: 2px solid $state-danger-text;
 }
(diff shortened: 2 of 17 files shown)


spec/factories/draft_taxon.rb
```diff
@@ -0,0 +1,12 @@
+FactoryBot.define do
```
koetsier: This is very similar to the Taxon factory - could this not better be a trait of the taxon factory instead?

spec/factories/previously_published_draft_taxon.rb
```diff
@@ -0,0 +1,18 @@
+FactoryBot.define do
+  factory :previously_published_draft_taxon, class: Taxon do
```
koetsier: can also be a trait of taxon

app/models/taxon.rb
```diff
@@ -37,6 +38,20 @@ def redirected?
...
   end
 
+  def draft_previously_published?
+    previous_state, latest_state = last_two_publication_states
+    previous_state && latest_state == "draft"
+  end
+
+  def ordered_publication_state_history
+    ordered_history = state_history.sort_by { |k, _v| k }
+    ordered_history.map!(&:second)
```
cbaines: Would this work instead?
```ruby
state_history.sort_by(&:first).map(&:second)
```

This would remove the use of `map!`, and the `ordered_history` variable.

app/models/taxon.rb
```diff
@@ -37,6 +38,20 @@ def redirected?
...
+  end
+
+  def last_two_publication_states
+    ordered_publication_state_history.last(2)
+  end
```
cbaines: The only other place this is called is in the tests? I'd suggest just saying the following in `draft_previously_published?` (`.last(2)` doesn't make any difference apparently).

```ruby
previous_state, latest_state = ordered_publication_state_history
```

app/models/taxon.rb
```diff
@@ -37,6 +38,20 @@ def redirected?
     publication_state == "unpublished" && !redirect_to.nil?
   end
 
+  def draft_previously_published?
+    previous_state, latest_state = last_two_publication_states
+    previous_state && latest_state == "draft"
+  end
```
cbaines: I think `draft_and_published_editions_exist?` might be a clearer name for this.

app/models/taxon.rb
leenagupte: I'lI give it a try.

app/models/taxon.rb
leenagupte: That would assign the first two states to `previous_state` and `latest_state`.

I could reverse the order in `ordered_publication_state_history` thought

app/models/taxon.rb
leenagupte: Sure.

app/models/taxon.rb
leenagupte: I'll put it back in.

leenagupte:
@koetsier @cbaines I've addressed your comments and fixed up the code. Cheers!

PR #782, opened 2018-07-16
Make the user UID available on the tagging history page

cbaines:
Currently, the user details only show if that user has signed in to
Content Tagger. This should be fixable in the future by fetching the
data from Signon, but for now, just put the UID in the HTML as a
comment.


cbaines:
Thanks for taking a look @tijmenb :)

PR #784, opened 2018-07-16
Make bulk tagging lower priority than content publishing

boffbowsh:
When we bulk tag things, we can set a `bulk_publishing` flag on the call to the Publishing API's Patch Links method. This makes Publishing API treat this action as lower priority and put it on the correct internal queue (`downstream_low`). We should always do this for bulk retagging due to the potential volumes involved. If a large bulk retagging was to be given a high priority, it would delay real time publishing, including travel advice and drug safety alerts.


cbaines:
Merging so that this can be deployed soon.

PR #800, opened 2018-08-24
Add task to remove descriptions from certain branches

1pretz1:
This will clear the descriptions of the taxons children from the taxons
specified. This is due to the descriptions not being accurate for the
corresponding taxons. 

I will have to manually clear the descriptions for the level one taxons as this code only deals with the children.

Trello;
https://trello.com/c/tYwP644l/156-remove-all-descriptions-for-all-newly-live-branches-of-the-taxonomy

### DIFF
--- a/lib/tasks/taxonomy/clear_descriptions.rake
+++ b/lib/tasks/taxonomy/clear_descriptions.rake
@@ -1,9 +1,18 @@
 require 'taxon_description_updater'
+require 'description_remover'
+
 namespace :taxonomy do
   desc <<-DESC
     Clears all descriptions containing '...' or 'tbc'.
   DESC
   task clear_descriptions: [:environment] do
     TaxonDescriptionUpdater.new(%w[... tbc TBC ...tbc]).call
+  end
+
+  desc <<-DESC
+    Clears description from taxons children recursively.
+  DESC
+  task :clear_descriptions_for_branches, [:base_path] => [:environment] do |_, args|
+    DescriptionRemover.call(args[:base_path])
   end
 end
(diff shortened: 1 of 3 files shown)


lib/tasks/taxonomy/clear_descriptions.rake
```diff
@@ -1,9 +1,30 @@
...
+      %w[
+        /work
+        /housing-local-and-community
+        /health-and-social-care
+        /corporate-information
+        /going-and-being-abroad
+        /life-circumstances
+        /regional-and-local-government
+        /society-and-culture
+        /welfare
```
cbaines: Rather than hardcoding these branches, I think it would be better to make this task take a branch as an argument. Hopefully there will be descriptions written for these branches eventually, removing them is only a short term change.

spec/lib/description_remover_spec.rb
```diff
@@ -0,0 +1,102 @@
+require 'rails_helper'
+require 'description_remover'
+require 'gds_api/test_helpers/content_store'
+
+include ::GdsApi::TestHelpers::ContentStore
+include ::GdsApi::TestHelpers::PublishingApiV2
+
+RSpec.describe DescriptionRemover do
+  context "valid taxons are updated" do
```
cbaines: I'm not sure what "valid" means here, do you mean "appropriate"?

lib/tasks/taxonomy/clear_descriptions.rake
1pretz1: Sure

spec/lib/description_remover_spec.rb
1pretz1: 👍

PR #816, opened 2018-09-13
Improve tagged_content interface

1pretz1:
Improvements made:

* Moved the javascript for the filter table from govuk-admin-template
into content tagger as this gem is now depreciated.

* The filter table is now usable when there is high amount of content
tagged to a taxon.

* There is a refreshing icon which tells the user their filtering is in
process.

Trello:
https://trello.com/c/i3UoSjH9/118-l-make-move-content-feature-in-content-tagger-easier-and-safer-for-publishers-to-use

### DIFF
--- a/app/assets/javascripts/filter-table.js
+++ b/app/assets/javascripts/filter-table.js
@@ -1,0 +1,63 @@
+(function(Modules) {
+  "use strict";
+
+  Modules.FilterableTable = function() {
+    var that = this;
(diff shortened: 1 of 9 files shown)


1pretz1:
Yep, that's what i was thinking! Also, removing the `govuk_admin_template` gem from content-tagger, but I have yet to investigate whether it is used in more places.

PR #833, opened 2018-11-01
extract all mainstream content with taxons and writes it to a csv file

koetsier:
Exports the title and base path of all mainstream content with
all the taxons that the content is tagged to.

Trello: https://trello.com/c/aRATrD91/121-export-mainstream-taxons


lib/tasks/export_mainstream_taxons.rake
```diff
@@ -0,0 +1,47 @@
+namespace :govuk do
+  # bundle exec rake govuk:export_content_by_organisations[uk-border-agency,border-force]
```
vanitabarrett: This comment needs updating?

PR #879, opened 2019-03-15
Tag content with facet values

steventux:
https://trello.com/c/ws1zcCKX/286-update-facet-values-for-a-content-item

Screenshot_from_2019-03-14_16-38-40

This PR appends an additional _Facets_ tagging select2 input which allows facet values to be updated for a content item.

At present there's only one facet group defined in the Publishing API hence the hard-wiring to this specific group until we've iterated the UI to tag within the context of multiple facet groups.
This feature means we can opt to update specific content item `facet_values` links in content tagger similar to topics and related links. The aim being to ultimately replace the current metadata tagging process in Rummager.

### DIFF
--- a/app/models/facets/remote_facet_groups_service.rb
+++ b/app/models/facets/remote_facet_groups_service.rb
@@ -1,5 +1,13 @@
 module Facets
   class RemoteFacetGroupsService
+    # TODO: There's currently only one facet group defined, and we don't yet
+    # have a sophisticated enough UI to 'choose' different facet groups so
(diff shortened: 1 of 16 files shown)


steventux:
I think this PR is missing the case where a previously untagged content item is tagged, though this would add `facet_values` links it would not add the appropriate facet group link.

steventux:
Closing, this approach works but we need to factor in pinning and using different facet groups so we're going to work on facet value tagging in a bespoke tagging view.

PR #884, opened 2019-03-20
Tag a page with facet values

steventux:
https://trello.com/c/ws1zcCKX/286-update-facet-values-for-a-content-item

Screenshot from 2019-03-21 17-15-45

- Adds a tagging interface specifically for tagging content items with facet values.
- Adds the ability to pin the content item to the top of filtered finder results.

#### TODO (in other stories):

- We make the assumption that the 1 finder using item pinning is to be updated, this is commented clearly but further work needed when pinning can be done across finders.


steventux:
@peteglondon I've pushed a couple of new commits since testing this out on integration if you wouldn't mind reviewing

PR #897, opened 2019-04-11
Add a link to facet groups

AgaDufrat:
This adds a link to the "facet_groups" page to the content-tagger menu limited to users with 'GDS Editor' permissions.

- User with 'GDS Editor' permission:
with 'GDS Editor' permissions

- User without 'GDS Editor' permission:
without 'GDS Editor' permissions

[Trello card](https://trello.com/c/QqN7QCOC/34-permissions-for-facets-page-in-content-tagger)

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -40,6 +40,12 @@
       <%= link_to t('navigation.projects'), projects_path %>
     </li>
   <% end %>
+
+  <% if user_can_administer_facet_groups? %>
+    <li class='<%= active_navigation_item.in?(%w[facet_groups]) ? 'active' : nil %>'>
+      <%= link_to t('navigation.facet_groups'), facet_groups_path %>
+    </li>
+  <% end %>
 <% end %>
 
 <%= render template: 'layouts/govuk_admin_template' %>
--- a/spec/lib/permission_checker_spec.rb
+++ b/spec/lib/permission_checker_spec.rb
@@ -8,6 +8,7 @@ RSpec.describe PermissionChecker do
     it { is_expected.not_to have_permission(:user_can_administer_taxonomy?) }
     it { is_expected.not_to have_permission(:user_can_manage_taxonomy?) }
     it { is_expected.not_to have_permission(:user_can_access_tagathon_tools?) }
+    it { is_expected.not_to have_permission(:user_can_administer_facet_groups?) }
   end
 
   context "when the user has the Gds Editor permission" do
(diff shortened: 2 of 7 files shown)


app/views/layouts/application.html.erb
```diff
@@ -40,6 +40,12 @@
       <%= link_to t('navigation.projects'), projects_path %>
     </li>
   <% end %>
+
+  <% if user_can_administer_taxonomy? %>
```
leenagupte: If we ever decide to open up this role to other users, we'll need to create a separate check for it, but at the moment it's fine to put it with the taxonomy permissions.

AgaDufrat: I would be happy to create an explicit method (eg user_can_administer_facet_groups?). 
I saw it as a trade of between DRY and Clean (meaningful names and functions) code. Very interested in your views. Would it be considered repeating?

AgaDufrat:
Thank you for the feedback! I agree that it could have been squashed into 1 commit. I decided to split them because I need to collect evidence that I "test drive code" as a part of my apprenticeship ;)

app/views/layouts/application.html.erb
leenagupte: If we were going to really DRY-up the code, we'd just make `gds_editor?` public and use that instead, but that doesn't really tell you what that user can do.

I would have created a new method, just to make it clear that administering the taxonomy and administering facet groups are two different things, even though at the moment, you need the same permissions for both. But that's because my own preference is for code that's easy to read so you don't have to do too much digging to find out what something means.

AgaDufrat: Thank you! Really insightful. I will created a new method.

spec/lib/permission_checker_spec.rb
```diff
@@ -8,6 +8,7 @@
     it { is_expected.not_to have_permission(:user_can_administer_taxonomy?) }
     it { is_expected.not_to have_permission(:user_can_manage_taxonomy?) }
     it { is_expected.not_to have_permission(:user_can_access_tagathon_tools?) }
+    it { is_expected.not_to have_permission(:user_can_administer_facet_groups?) }
```
leenagupte: Nitpick: Can you add this expectation to the "Managing Editor" and "Tagathon Participant" tests as well?

app/views/layouts/application.html.erb
leenagupte: It looks good! 💯  There was only one nitpicky comment.

spec/lib/permission_checker_spec.rb
AgaDufrat: ah, yes!

PR #906, opened 2019-04-18
Fix short name for employ eu citizens

steventux:
https://github.com/alphagov/govuk-app-deployment-secrets/blob/master/shared_config/find-eu-exit-guidance-business.yml#L194 

Screenshot from 2019-04-18 16-19-11

The facet `short_name` is attribute used as a display value for one particular facet on the business readiness finder. This custom value provides a better grouping header for results. The facet value on the relevant content item needs to be consistent with what's currently in the finder facets definition.

This change will need deploying and the facet group importing and publishing as per the rake tasks documented here : https://docs.publishing.service.gov.uk/manual/publishing-a-facet-group.html

### DIFF
--- a/lib/data/find-eu-exit-guidance-business.yml
+++ b/lib/data/find-eu-exit-guidance-business.yml
@@ -229,6 +229,7 @@
         :value: transporting
   - :content_id: b2716a1e-dc24-4566-b8ed-f3ee23a994ec
     :name: "Who you employ"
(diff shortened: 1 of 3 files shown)


leenagupte:
@steventux Does the facet schema need to be updated too?

steventux:
@leenagupte no https://github.com/alphagov/govuk-content-schemas/blob/master/formats/facet.jsonnet#L52

PR #914, opened 2019-05-07
Present taggable facets in separate fields

steventux:
https://trello.com/c/SgSAqgLR/79-split-facet-tagging-form-fields-into-separate-facets

Improves facet tagging workflow by splitting facet values out into separate facet select2 fields.

Screenshot from 2019-05-07 10-56-55

### DIFF
--- a/spec/features/tag_facets_to_a_page_spec.rb
+++ b/spec/features/tag_facets_to_a_page_spec.rb
@@ -270,7 +270,7 @@ RSpec.describe "Tagging content with facets", type: :feature do
   end
 
   def and_i_see_the_facet_values_form_prefilled_with(option)
-    facet_values_options = all('#facets_tagging_update_form_facet_values option').map(&:text)
+    facet_values_options = all('#facets_tagging_update_form_example_facet option').map(&:text)
     expect(facet_values_options).to include(option)
   end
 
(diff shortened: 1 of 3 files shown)


spec/features/tag_facets_to_a_page_spec.rb
```diff
@@ -270,7 +270,7 @@ def and_i_edit_facets_for_the_page(path)
   end
 
   def and_i_see_the_facet_values_form_prefilled_with(option)
-    facet_values_options = all('#facets_tagging_update_form_facet_values option').map(&:text)
+    facet_values_options = all('#facets_tagging_update_form_example_facet option').map(&:text)
```
leenagupte: Nitpick: should this be `facets_tagging_update_form_example_facet_group`?

steventux: It's not a facet group. It's the text from the options on a set of facet values.

leenagupte: Ok, found it: https://github.com/alphagov/content-tagger/blob/ed1797613ffd42980b12cd26e3820e1489e76253/spec/support/publishing_api_helper.rb#L244

There are a lot of things named `example_facet....` so it took a while to find.

PR #922, opened 2019-05-17
Remove link to finder when content is untagged from facet group

leenagupte:
Trello: https://trello.com/c/u5KXZwqD
Related to: #916 

If a content item is "untagged" from the business finder, it should no longer show the finder in the breadcrumb.

An empty list is still being passed publishing-api to make sure that the entry for `finder` in the links is overwritten.

### DIFF
--- a/app/services/facets/tagging_update_publisher.rb
+++ b/app/services/facets/tagging_update_publisher.rb
@@ -68,23 +68,20 @@ module Facets
     def generate_links_payload
       facet_groups_content_ids = fetch_content_ids(:facet_groups)
       facet_values_content_ids = fetch_content_ids(:facet_values)
+      finder_content_ids = [FinderService::LINKED_FINDER_CONTENT_ID]
 
       if facet_values_content_ids.any?
         facet_groups_content_ids.push(facet_group_content_id)
       else
         facet_groups_content_ids.delete(facet_group_content_id)
+        finder_content_ids = []
       end
 
-      links = {
+      {
         facet_groups: facet_groups_content_ids.uniq,
         facet_values: facet_values_content_ids,
+        finder: finder_content_ids,
       }
-
-      if facet_values_content_ids.any?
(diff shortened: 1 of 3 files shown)


app/services/facets/tagging_update_publisher.rb
```diff
@@ -68,23 +68,20 @@ def updated_pinned_items
     def generate_links_payload
       facet_groups_content_ids = fetch_content_ids(:facet_groups)
       facet_values_content_ids = fetch_content_ids(:facet_values)
+      finder_content_ids = [FinderService::LINKED_FINDER_CONTENT_ID]
 
       if facet_values_content_ids.any?
         facet_groups_content_ids.push(facet_group_content_id)
       else
         facet_groups_content_ids.delete(facet_group_content_id)
+        finder_content_ids = []
```
ChrisBAshton: Maybe add a comment to the code so we don't have to check the commit to find out why it's here?

```suggestion
        finder_content_ids = [] # overwrite any published entries for `finder` in the links
```

leenagupte: @ChrisBAshton The tests have been updated to add the empty list, so they will break if this line is removed.
I'd rather the make code easier to read than add a comment. Would it be easier to understand if `finder_content_ids` was split into a separate method?  
i.e.

```
def finder_content_ids
  return [FinderService::LINKED_FINDER_CONTENT_ID] if facet_groups_content_ids.any?
  []
end
```

ChrisBAshton: > The tests have been updated to add the empty list, so they will break if this line is removed.

D'oh 🤦‍♂ so they are!

The code itself is perfectly readable, it's just not hugely clear what the consequences of removing this would be unless you know how it fits into the publishing-api, if that makes sense. Maybe that's more of an onboarding thing, or something worth documenting in https://docs.publishing.service.gov.uk/manual/business-readiness-publish-changes.html.

Happy to approve 👍

PR #933, opened 2019-06-05
Add a related link to content tagged to the EU Exit Business Readiness Finder

brucebolt:
There are two aspects:
- New rake task to add related link for business finder linked content.  This adds a related link (back to the finder itself) to all content items that are tagged to the EU Exit Business Readiness Finder.  This is intended to be run once to add the related link to all existing content, then can be removed.  Task run as `bundle exec rake eu_exit_business_finder:add_related_link_to_tagged_content`.
- Adding a related link when tagging to the business finder.  When content is tagged to the business readiness finder, a related link will be created in the content item.  This related link will always be the first related item for that content item, so will show first in the list (in the event of there being other related links for that content).  The related link is removed when the content item is no longer tagged to any of the facets within the finder.  All other related links are retained.

Example of document with related link added:
Screen Shot 2019-06-05 at 14 00 04

Trello card: https://trello.com/c/GqRWFNoB/144-add-a-related-link-to-all-content-tagged-to-the-business-finder-to-take-users-to-it

### DIFF
--- a/app/services/facets/tagging_update_publisher.rb
+++ b/app/services/facets/tagging_update_publisher.rb
@@ -69,19 +69,30 @@ module Facets
       facet_groups_content_ids = fetch_content_ids(:facet_groups)
       facet_values_content_ids = fetch_content_ids(:facet_values)
       finder_content_ids = [FinderService::LINKED_FINDER_CONTENT_ID]
+      ordered_related_items_ids = ordered_related_links
 
       if facet_values_content_ids.any?
         facet_groups_content_ids.push(facet_group_content_id)
+        ordered_related_items_ids.unshift(FinderService::LINKED_FINDER_CONTENT_ID)
       else
         facet_groups_content_ids.delete(facet_group_content_id)
         finder_content_ids = []
+        ordered_related_items_ids.delete(FinderService::LINKED_FINDER_CONTENT_ID)
       end
 
       {
         facet_groups: facet_groups_content_ids.uniq,
         facet_values: facet_values_content_ids,
         finder: finder_content_ids,
+        ordered_related_items: ordered_related_items_ids,
       }
+    end
+
+    def ordered_related_links
+      Services.publishing_api.get_links(content_item.content_id)
+        .to_hash
+        .fetch("links")
+        .fetch("ordered_related_items", [])
     end
 
     # FIXME: This is a temporary tag set which can be removed once
--- a/lib/tasks/eu_exit_business_finder.rake
+++ b/lib/tasks/eu_exit_business_finder.rake
@@ -1,0 +1,21 @@
+namespace :eu_exit_business_finder do
+  desc <<-DESC
+    Adds a link to the EU Exit Business Readiness Finder to all content items which are tagged to this finder
+  DESC
+  task add_related_link_to_tagged_content: [:environment] do
+    facet_group_content_id = "52435175-82ed-4a04-adef-74c0199d0f46"
+    finder_content_id = "42ce66de-04f3-4192-bf31-8394538e0734"
+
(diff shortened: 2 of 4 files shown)


lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,21 @@
+namespace :eu_exit_business_finder do
+  desc <<-DESC
+    Adds a link to the EU Exit Business Readiness Finder to all content items which are tagged to this finder
+  DESC
+  task add_related_link_to_tagged_content: [:environment] do
+    facet_group_content_id = "52435175-82ed-4a04-adef-74c0199d0f46"
+    finder_content_id = "42ce66de-04f3-4192-bf31-8394538e0734"
```
leenagupte: I don't expect the finder content item is going to change, but it you wanted to, you could use `Facets::FinderService::LINKED_FINDER_CONTENT_ID` here instead. I don't think the facet group content id is stored anywhere apart from the config file though

leenagupte: Have you considered adding a test for this task?

spec/services/facets/tagging_update_publisher_spec.rb
```diff
@@ -102,4 +108,16 @@
       end
     end
   end
+
+  def stub_get_links_request(content_id: "MY-CONTENT-ID", items: ["RELATED-LINK"])
```
leenagupte: This is nice! Might be nice to add this to the test helpers in gds-api-adapters

app/services/facets/tagging_update_publisher.rb
```diff
@@ -69,21 +69,32 @@ def generate_links_payload
...
       finder_content_ids = [FinderService::LINKED_FINDER_CONTENT_ID]
+      ordered_related_items_ids = ordered_related_links
 
       if facet_values_content_ids.any?
         facet_groups_content_ids.push(facet_group_content_id)
+        ordered_related_items_ids.unshift(FinderService::LINKED_FINDER_CONTENT_ID)
       else
         facet_groups_content_ids.delete(facet_group_content_id)
         finder_content_ids = []
+        ordered_related_items_ids.delete(FinderService::LINKED_FINDER_CONTENT_ID)
```
oscarwyatt: I'm a huge fan of the simplicity here, huge
I _think_ it's clear enough to a casual viewer that we want to remove the finder content id if the facet_value is missing, I was going to suggest putting something in the commit message but you already have! What a pro 🏅

lib/tasks/eu_exit_business_finder.rake
brucebolt: I hardcoded the content IDs since the task will probably only be run once, then removed.  So it didn't see to give any benefit by using the `LINKED_FINDER_CONTENT_ID` value here.

I'll look into adding a test 😬.

spec/services/facets/tagging_update_publisher_spec.rb
brucebolt: Turns out there's already a `stub_publishing_api_has_links` helper which stubs the same endpoint.  I'll update this PR to use that method.

lib/tasks/eu_exit_business_finder.rake
brucebolt: I've had second thoughts around adding a test for the rake task.  I'm stubbing all the requests (before and after the related link gets added), so it's not actually testing anything, other than the correct publishing-api endpoints are being requested.

This code could be moved into self contained methods, then each tested with a unit test, but this feels against the ethos of optimising code for deletion since this task will be removed after it has been run in production.

PR #941, opened 2019-06-21
Make updates to business sectors and business activities for EU Exit Business Readiness Finder

brucebolt:
A number of changes to business sectors and business activities have been proposed:
https://docs.google.com/spreadsheets/d/1DDsWQvq3XoTwi3R3YS_z72ixON3mUl6aWjjsI2V4NHg/edit
https://docs.google.com/document/d/1xdEkrkN4OZFLsN6xWTBLWPgmXaDTMxp4PxSLqqX0wQ4/edit
https://docs.google.com/document/d/1NWADDxTfn3aaTGPYD8hPJ-vZfPg0fpPDcDdo2gur3FQ/edit

This PR encapsulates all of these changes; including the merging/splitting/deleting of sectors, updating of the business activity question and retagging of all affected content.

Deploy https://github.com/alphagov/finder-frontend/pull/1211 at the same time.

Once deployed, the following rake tasks must be run to update the `facet_groups` content item and also retag existing content:
```
## content-tagger
rake facets:import_facet_group["lib/data/find-eu-exit-guidance-business.yml"]
rake facets:publish_facet_group["lib/data/find-eu-exit-guidance-business.yml"]
rake eu_exit_business_finder:retag_documents_to_facet_values
## search-api
rake publishing_api:publish_facet_group_eu_exit_business_finder
```

Trello card: https://trello.com/c/VTKL8TXk

### DIFF
--- a/lib/tasks/eu_exit_business_finder.rake
+++ b/lib/tasks/eu_exit_business_finder.rake
@@ -1,0 +1,152 @@
+namespace :eu_exit_business_finder do
+  task retag_documents_to_facet_values: [:environment] do
+    EuExitBusinessRakeMethods.replace_facet_values
+    EuExitBusinessRakeMethods.retag_content_facet_values
+    EuExitBusinessRakeMethods.delete_facet_values_from_content
+  end
+end
+
+module EuExitBusinessRakeMethods
+  def self.facet_values_to_replace
+    # For each item in the array:
+    # All documents tagged with `old_facet_value` will be re-tagged
+    # with `new_facet_value`
+    [
+      {
+        "old_facet_value": "9f3476e1-8ff0-455d-a14e-003236b2797c",
+        "new_facet_value": "53f9ce4c-7cbb-447f-bdf1-a9b022896d3a"
+      },
+
+      {
+        "old_facet_value": "b4e507df-f067-4749-9468-3de120775216",
+        "new_facet_value": "24fd50fa-6619-46ca-96cd-8ce90fa076ce"
+      },
+
+      {
+        "old_facet_value": "1e3e8abd-135d-4844-afa8-5c51df3d3c57",
+        "new_facet_value": "34a6edd0-46ea-4a76-ae80-8c96709d4f59"
+      },
+
+      {
+        "old_facet_value": "c1d8057c-76bf-431c-9ac8-6281a4b7b9ca",
+        "new_facet_value": "34a6edd0-46ea-4a76-ae80-8c96709d4f59"
+      },
+
+      {
+        "old_facet_value": "356f46a0-17d3-4ba4-8952-8c02244f9045",
+        "new_facet_value": "18c3892e-8a0e-4884-906e-5938380eceee"
+      }
+    ]
+  end
+
+  def self.content_to_update_facet_values
+    # For each base path:
+    # The document will be untagged from `old_facet_value` and
+    # will be re-tagged with `new_facet_value`
+    [
+      {
+        "base_paths": [
+          "/guidance/export-animal-bones-protein-and-other-by-products-special-rules",
+          "/guidance/export-food-and-agricultural-products-special-rules",
+          "/government/publications/farm-payments-if-theres-no-brexit-deal/farm-payments-if-theres-no-brexit-deal",
+          "/guidance/how-to-comply-with-pesticide-regulations-after-brexit",
+          "/government/publications/industrial-emissions-standards-best-available-techniques-if-theres-no-brexit-deal",
+          "/guidance/protecting-food-and-drink-names-if-theres-no-brexit-deal",
+          "/government/publications/receiving-rural-development-funding-if-theres-no-brexit-deal/receiving-rural-development-funding-if-theres-no-brexit-deal",
+          "/guidance/the-farming-sector-and-preparing-for-eu-exit",
+          "/government/publications/upholding-environmental-standards-if-theres-no-brexit-deal"
+        ],
+        "old_facet_value": "94b3cfe2-af89-4744-b8d7-7fc79edcbc85",
+        "new_facet_value": "9d54c591-f5ca-4d0c-a484-12d5591987cb"
+      },
+      {
+        "base_paths": [
+          "/government/publications/approval-and-operation-of-plant-health-inspection-facilities-at-place-of-first-arrival",
+          "/guidance/importing-and-exporting-plants-and-plant-products-if-theres-no-withdrawal-deal",
+          "/guidance/trading-timber-imports-and-exports-if-theres-no-brexit-deal"
+        ],
+        "old_facet_value": "94b3cfe2-af89-4744-b8d7-7fc79edcbc85",
+        "new_facet_value": "afd45eef-743a-417d-9245-3eab8322116d"
+      }
+    ]
+  end
+
+  def self.facet_values_to_untag
+    # For each item in the array:
+    # All content tagged to that facet value will be untagged
+    [
+      "7536c0c4-fb41-43f4-a2c4-08f4fa9f5427",
+      "5faa1741-fc55-4110-b342-de92f6324118",
+      "14cf2a68-3297-44d3-ba01-a4426845b1b8",
+      "040649fc-4e2c-4028-b846-77fe3eebd1f7",
+      "94b3cfe2-af89-4744-b8d7-7fc79edcbc85"
+    ]
+  end
+
+  def self.replace_facet_values
+    content_changes = []
+    facet_values_to_replace.each do |facet_values_merge|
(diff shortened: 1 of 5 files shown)


lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,134 @@
...
+        "old_facet_value": facet_values_merge[:old_facet_value],
+        "new_facet_value": facet_values_merge[:new_facet_value]
+      }
+    end
+
+    bulk_retag_documents(content_changes)
+  end
+
+  desc <<-DESC
+    Removes a specified facet_value from a list of documents, then replaces this with a new facet_value
```
ChrisBAshton: Description suggests this takes a parameter, but it looks like it doesn't...?

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,134 @@
...
+        links = Services.publishing_api.get_links(content_id).to_hash.fetch('links', {})
+        content_facet_values = links.fetch('facet_values', [])
+        content_facet_values.delete(facet_value)
+        links['facet_values'] = content_facet_values
+        Services.publishing_api.patch_links(content_id, links: links)
+      end
+    end
+  end
+
+  def get_content_from_facet_value(facet_value)
```
ChrisBAshton:
```suggestion
  def get_content_ids_from_facet_value(facet_value)
```

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,134 @@
+namespace :eu_exit_business_finder do
+  desc <<-DESC
+    Updates all documents that are tagged with a facet_value with a replacement value
```
ChrisBAshton:
```suggestion
    Re-tags all documents tagged with a deprecated facet_value with its replacement facet_value
```

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,134 @@
...
+    content_links = Services.rummager.search_enum(
+      filter_facet_values: facet_value
+    ).pluck("link")
+    content_links.each do |content_link|
+      content_ids << Services.rummager.get_content(content_link).to_hash['raw_source']['content_id']
+    end
+    content_ids
+  end
+
+  def bulk_retag_documents(content_changes)
```
ChrisBAshton: Nice abstraction 👍

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,134 @@
...
+        "content_ids": [
+          "02a7bf25-bcc9-4e3c-aaf8-b30eda527589",
+          "ddcac9ba-a852-4fe6-9d1a-73c64e6fe591",
+          "b086e251-aae4-4811-8f9b-08086ad355e2",
+          "6cf648b4-fe85-4eb9-9468-59cd16309475",
+          "a115dfb1-43c7-4a07-9300-87c50187c7c1",
+          "6a8907cc-8200-4df7-877c-2770370c2b0b",
+          "474e8df7-0229-4e96-bf5e-e430f3a19bf2",
+          "4cc71bfd-4a2b-4bb0-b3b2-de9a9909ab53",
+          "9de0b81c-af80-4cf7-827a-e196d6d3811f"
```
ChrisBAshton: Could this task be avoided by just adding these content_ids to the `facet_values_merges` array in the first task?

lib/tasks/eu_exit_business_finder.rake
brucebolt: I've reworked this rake task, so the comment is no longer present

lib/tasks/eu_exit_business_finder.rake
brucebolt: I've reworked this too

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,155 @@
...
+        links = Services.publishing_api.get_links(content_id).to_hash.fetch('links', {})
+        content_facet_values = links.fetch('facet_values', [])
+        content_facet_values.delete(facet_value)
+        links['facet_values'] = content_facet_values
+        Services.publishing_api.patch_links(content_id, links: links)
+      end
+    end
+  end
+
+  def get_content_ids_from_facet_value(facet_value)
```
leenagupte: It looks like this method is not being used.

leenagupte: I think it's supposed to be called `get_content_from_facet_value`

leenagupte: This is something a test would have picked up 😃.

If the intention is to keep this rake task around for future generations, a test might be a good idea.

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,155 @@
...
+    ]
+  end
+
+  def facet_values_to_untag
+    # For each item in the array:
+    # All content tagged to that facet value will be untagged
+    [
+      "7536c0c4-fb41-43f4-a2c4-08f4fa9f5427",
+      "5faa1741-fc55-4110-b342-de92f6324118",
+      "c1d8057c-76bf-431c-9ac8-6281a4b7b9ca",
```
ChrisBAshton: This is the only facet value that has a corresponding entry in `facet_values_to_replace` - that feels like a mistake?

brucebolt: Yes, it shouldn't be there since there will be no content tagged to it by this point.  Good spot 👍

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,156 @@
...
+  end
+
+  def self.retag_content_facet_values
+    content_changes = content_to_update_facet_values
+
+    content_changes.each do |content_change|
+      content_change[:content_ids] = Services.publishing_api.lookup_content_ids(base_paths: content_change[:base_paths]).values
+    end
+
+    bulk_retag_documents(content_changes)
```
ChrisBAshton: Is there any way we can avoid calling `bulk_retag_documents` twice, by setting up all the content changes in advance?

spec/lib/tasks/eu_exit_business_finder_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+            "old-facet-value",
+            "another-facet-value"
+          ]
+      }
+    }
+  end
+
+  let(:updated_content_item_links) do
+    {
+      "facet_values" =>
```
ChrisBAshton: Mixture of styles - this one is using `=>`, the one on line 63 is using `:`. Think the linter prefers the colon.

brucebolt: This is deliberate, since `patch_links` requires the `facet_values` key as a string (which `=>` gives you) rather than a symbol (which `:` gives).

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,156 @@
...
+      "7536c0c4-fb41-43f4-a2c4-08f4fa9f5427",
+      "5faa1741-fc55-4110-b342-de92f6324118",
+      "14cf2a68-3297-44d3-ba01-a4426845b1b8",
+      "040649fc-4e2c-4028-b846-77fe3eebd1f7",
+      "94b3cfe2-af89-4744-b8d7-7fc79edcbc85"
+    ]
+  end
+
+  def self.replace_facet_values
+    facet_values_merges = facet_values_to_replace
```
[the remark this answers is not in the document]
brucebolt: That's a good point.  I'll refactor this.

lib/tasks/eu_exit_business_finder.rake
brucebolt: I was thinking that by splitting this out into two parts (one to retag specific content and another to retag all content with a facet value), then it would make future updates easier where only one type of update were needed.

spec/lib/tasks/eu_exit_business_finder_spec.rb
leenagupte: The keys are a string on line 63 as well. 🤔

brucebolt: I've updated the one on line 63 👍

leenagupte: That was just one example! I'm surprised the linter hasn't picked up on this.

leenagupte: AFAIK the linter like `:` for symbols and `=>` for strings.

PR #960, opened 2019-08-02
Add Whitehall document types that support related links

karlbaker02:
This commit adds existing Whitehall document types to Content Tagger that support related links. Originally, this was a list of apps that weren't able to render the sidebar (back in 2016) which was then converted to the associated document types (in 2017).

We now display the sidebar for many more Whitehall document types; by allowing the related links component to be renderable when editing a content item within Content Tagger, publishers will be able to update related links. Related links overrides can be added to certain Whitehall documents at the moment, but only where the document is not tagged to a taxon, which affects the minority of documents and doesn't reliably change the related links shown within the component even if set.

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -77,20 +77,78 @@ private
 
   def related_links_are_renderable?
     %w[
+      aaib_report
       answer
+      asylum_support_decision
+      authored_article
+      business_finance_support_scheme
       calculator
       calendar
+      case_study
+      closed_consultation
+      cma_case
+      consultation
+      consultation_outcome
       contact
+      corporate_report
+      correspondence
+      countryside_stewardship_grant
+      decision
+      detailed_guidance
+      detailed_guide
+      dfid_research_output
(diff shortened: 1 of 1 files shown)


kevindew:
This sounds reasonable, though I notice this includes a lot of types that are published by specialist publisher rather Whitehall. Is that intentional?

Also I notice quite a few items are repeated twice in the list. I’m assuming that’s a mistake? 

Finally, any concerns that some of these might have a design that doesn’t have a sidebar?

karlbaker02:
Apologies, I'm not sure how those duplicates got in there. I will remove them.

I've done a visual check on a large number of these document types and have seen a sidebar, regardless of whether there has been any content in the contextual sidebar. Regarding documents published by specialist publisher directly, these documents again seem to have a sidebar and therefore potentially could have related links: see https://www.gov.uk/export-health-certificates/export-to-chile-of-poultry-oils-and-fats-intended-for-animal-consumption-8503ehc-v1 as an example.

PR #963, opened 2019-08-08
Add suggested_ordered_related_items as an editable tag type

karlbaker02:
This PR adds `suggested_ordered_related_items` as an editable tag type, allowing publishers to re-order or remove any suggested related links. New links are not able to be added to `suggested_ordered_related_items` directly, as this property is updated by an automated process and as such links may be overwritten.

As a sidenote, I'm looking to remove `ordered_related_items_overrides` in a separate PR as this looks to be legacy.


karlbaker02:
Looking into the test failure.

PR #1000, opened 2019-10-17
Publish 'cy' Brexit taxon

emmabeynon:
For https://trello.com/c/3DUTcMYK/104-build-welsh-language-landing-page

We've received a request to build a Welsh translation of the '/brexit' page.  In order to achieve this we need to publish '/brexit' in the `cy` locale.  '/brexit' is a taxon and currently we don't publish taxons in locales other than `en`, so this is a somewhat hacky approach to publishing the Brexit taxon with a `cy` locale.  In addition to publishing we've ensured that any other Publishing API-related actions taken on the Brexit taxon are also reflected in its `cy` representation.  

We're not aware of any other needs to publish taxons in other locales and given the urgency of the work this approach is very much custom for the Brexit taxon and hasn't been developed in a particularly reusable way.  We will need to look at the longer-term implications of this from a product perspective and it will be logged as GOV.UK tech debt.

### DIFF
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -1,6 +1,7 @@
 class TaxonsController < ApplicationController
+  include BrexitTaxon
+
   VISUALISATIONS = %w[list bubbles taxonomy_tree].freeze
-  BREXIT_TAXON_CONTENT_ID = "d6c2de5d-ef90-45d1-82d4-5f2438369eea".freeze
 
   before_action(
     :ensure_user_can_administer_taxonomy!,
@@ -142,7 +143,7 @@ class TaxonsController < ApplicationController
 
   def publish
     Services.publishing_api.publish(content_id)
-    if content_id == BREXIT_TAXON_CONTENT_ID
+    if brexit_taxon?(content_id)
       Services.publishing_api.publish(content_id, nil, locale: "cy")
     end
 
@@ -174,7 +175,7 @@ class TaxonsController < ApplicationController
   def discard_draft
     Services.publishing_api.discard_draft(content_id)
 
-    if content_id == BREXIT_TAXON_CONTENT_ID
+    if brexit_taxon?(content_id)
       Services.publishing_api.discard_draft(content_id, locale: "cy")
     end
 
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -173,6 +173,11 @@ class TaxonsController < ApplicationController
 
   def discard_draft
     Services.publishing_api.discard_draft(content_id)
+
+    if content_id == BREXIT_TAXON_CONTENT_ID
+      Services.publishing_api.discard_draft(content_id, locale: "cy")
+    end
+
     redirect_to taxons_path, success: t("controllers.taxons.discard_draft_success")
   end
 
--- a/app/controllers/taxons_controller.rb
+++ b/app/controllers/taxons_controller.rb
@@ -1,5 +1,6 @@
 class TaxonsController < ApplicationController
   VISUALISATIONS = %w[list bubbles taxonomy_tree].freeze
+  BREXIT_TAXON_CONTENT_ID = "d6c2de5d-ef90-45d1-82d4-5f2438369eea".freeze
 
   before_action(
     :ensure_user_can_administer_taxonomy!,
@@ -141,6 +142,10 @@ class TaxonsController < ApplicationController
 
   def publish
     Services.publishing_api.publish(content_id)
+    if content_id == BREXIT_TAXON_CONTENT_ID
+      Services.publishing_api.publish(content_id, nil, locale: "cy")
+    end
+
     redirect_to taxon_path(content_id), success: "You have successfully published the taxon"
   rescue GdsApi::HTTPUnprocessableEntity => e
     # Perform a lookup on the base path to determine whether there
--- a/app/services/taxonomy/build_taxon_payload.rb
+++ b/app/services/taxonomy/build_taxon_payload.rb
@@ -4,28 +4,28 @@ module Taxonomy
       @taxon = taxon
     end
 
-    def self.call(taxon:)
-      new(taxon).build
+    def self.call(taxon:, locale: "en")
+      new(taxon).build(locale)
     end
 
-    def build
+    def build(locale)
       {
-        base_path: base_path,
+        base_path: path(locale),
         document_type: "taxon",
         schema_name: "taxon",
         title: title,
-        description: description,
+        description: locale == "en" ? description : nil,
         publishing_app: "content-tagger",
         rendering_app: "collections",
         public_updated_at: Time.now.iso8601,
-        locale: "en",
+        locale: locale,
         details: {
           internal_name: internal_name,
           notes_for_editors: notes_for_editors,
           visible_to_departmental_editors: visible_to_departmental_editors,
         },
         routes: [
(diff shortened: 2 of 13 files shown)


app/services/taxonomy/taxon_unpublisher.rb
```diff
@@ -31,7 +33,10 @@ def tag_to_parent(taxon_content_id, user)
 
     def unpublish_taxon(taxon_content_id, redirect_to_content_id)
       redirect_to_taxon = Services.publishing_api.get_content(redirect_to_content_id)
-      Services.publishing_api.unpublish(taxon_content_id, type: "redirect", alternative_path: redirect_to_taxon["base_path"])
+      publishing_api_unpublish(taxon_content_id, redirect_to_taxon["base_path"])
```
benthorner: I wonder if it would be simpler to just copy/paste the previous line and surround with an 'if' block, like we do [here](https://github.com/alphagov/content-tagger/pull/1000/files#diff-63a8e971bcba960dd75982227cab40e0R178). Although normally we'd aim to reduce duplication, in this case I think it's more important to keep the changes minimal and additive, so they're easy to identify.

app/services/taxonomy/update_taxon.rb
```diff
@@ -23,7 +25,7 @@ def publish(validate: true)
       # so that we can compare the differences between the two versions.
       Taxonomy::SaveTaxonVersion.call(taxon, @version_note)
 
-      Services.publishing_api.put_content(content_id, payload)
+      publishing_api_put_content_request(content_id)
```
benthorner: I wonder if it would be simpler to just copy/paste the previous line and surround with an 'if' block, like we do [here](https://github.com/alphagov/content-tagger/pull/1000/files#diff-63a8e971bcba960dd75982227cab40e0R178). Although normally we'd aim to reduce duplication, in this case I think it's more important to keep the changes minimal and additive, so they're easy to identify.

spec/workers/publish_taxon_worker_spec.rb
```diff
@@ -0,0 +1,15 @@
+require "rails_helper"
+
+RSpec.describe PublishTaxonWorker, "#perform" do
```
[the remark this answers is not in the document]
emmabeynon: Ah sure, will do 👍

app/services/taxonomy/update_taxon.rb
emmabeynon: Sadly using an if statement gave me a linting error 😞 `Style/GuardClause: Use a guard clause (return unless brexit_taxon?(taxon_content_id)) instead of wrapping the code inside a conditional expression. (https://rubystyle.guide#no-nested-conditionals)`

spec/controllers/taxons_controller_spec.rb
```diff
@@ -120,6 +120,17 @@
 
       expect(flash.now[:danger]).to match(/<a href="(.+)">taxon<\/a> with this slug already exists/)
     end
+
+    it "sends additional publish request to Publishing API for the Brexit taxon with 'cy' locale" do
+      brexit_taxon_content_id = "d6c2de5d-ef90-45d1-82d4-5f2438369eea"
```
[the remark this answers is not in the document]
emmabeynon: Makes sense, I've done the same in the other specs too.

app/services/taxonomy/taxon_unpublisher.rb
emmabeynon: Same linting error as below 😞

app/workers/update_taxon_worker.rb
```diff
@@ -8,7 +9,15 @@ def perform(content_id, attributes)
 
     Taxonomy::SaveTaxonVersion.call(updated_taxon, "Bulk update", previous_taxon: previous_taxon)
 
-    payload = Taxonomy::BuildTaxonPayload.call(taxon: updated_taxon)
-    Services.publishing_api.put_content(content_id, payload)
+    Services.publishing_api.put_content(content_id, payload(updated_taxon))
```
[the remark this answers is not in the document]
emmabeynon: Done 👍

emmabeynon:
Thanks for the review @benthorner, your comments should be addressed now.

app/services/taxonomy/taxon_unpublisher.rb
benthorner: I'd actually be happy if we disabled the cop for this purpose, but it's not a biggie.

app/controllers/taxons_controller.rb
```diff
@@ -141,6 +143,10 @@ def bulk_update
 
   def publish
     Services.publishing_api.publish(content_id)
```
koetsier: Would it be useful to extract the type of logic in line 145-148 to a separate module/class? It is repeated in quite a few places in the code and might make it difficult to change/delete at a later date.

emmabeynon: The publish API call for the Brexit taxon happens twice so I don't think it's too repetitive at this stage.  If there was another instance then I'd be tempted to look at extracting it out.  Also, if we lifted 145-148 out in future we would need to remember to put 145 back in anyway.  Hopefully the use of the `brexit_taxon?` method is clear enough for a future developer to know that anything it's wrapped around should be removed.

emmabeynon:
@benthorner The English description comment should actually be addressed now 😸I've opted to check if the taxon has a non-'en' locale rather than being specifically the Brexit taxon as this will cover any future non-'en' taxons that may arise while Content Tagger doesn't officially support them.

app/services/taxonomy/build_taxon_payload.rb
```diff
@@ -4,32 +4,35 @@ def initialize(taxon)
...
-      new(taxon).build
+    def self.call(taxon:, locale: "en")
+      new(taxon).build(locale)
     end
 
-    def build
-      {
-        base_path: base_path,
+    def build(locale)
+      path = locale == "en" ? base_path : base_path + ".#{locale}"
```
benthorner: Minor thing I didn't spot before: it might be clearer to make this a memoised private method.

spec/controllers/taxons_controller_spec.rb
```diff
@@ -120,6 +123,16 @@
...
       expect(flash.now[:danger]).to match(/<a href="(.+)">taxon<\/a> with this slug already exists/)
     end
+
+    it "sends additional publish request to Publishing API for the Brexit taxon with 'cy' locale" do
+      build(:taxon, publication_state: "unpublished", content_id: brexit_taxon_content_id)
+      stub_any_publishing_api_publish
+
+      post :publish, params: { taxon_id: brexit_taxon_content_id }
+
+      assert_publishing_api_publish(brexit_taxon_content_id, update_type: nil)
```
benthorner: One more minor thing: this line would benefit from `locale: "en"`, as I think otherwise it's superseded by the next line.

emmabeynon: Ah so the request doesn't have `locale: "en"` in the body as it's optional and defaults to `en` in the Publishing API anyway (see https://github.com/alphagov/publishing-api/blob/master/doc/api.md#post-v2contentcontent_idpublish).

app/services/taxonomy/build_taxon_payload.rb
emmabeynon: Done 👍

spec/controllers/taxons_controller_spec.rb
benthorner: Thanks :-).

PR #1007, opened 2019-11-06
Update to sprockets 4

cbaines:


cbaines:
> This has failing checks. flushed

I've now added some extra commits to fix the apparent SCSS issues that are now problematic with Sprockets 4.

PR #1037, opened 2020-01-29
Add rake task to unpublish/redirect brexit checker

koetsier:
We need to un-publish and redirect the current Brexit checker. This requires all query parameters to be retained and so a rake task is needed.

This PR can be reverted as soon as the content has been unpublished

trello: https://trello.com/c/LllJV5JJ/409-reslug-the-brexit-checker
see also: https://github.com/alphagov/special-route-publisher/pull/69

### DIFF
--- a/lib/tasks/brexit_checker_unpublisher.rake
+++ b/lib/tasks/brexit_checker_unpublisher.rake
@@ -1,0 +1,40 @@
+namespace :brexit_checker do
+  desc "Unpublish the brexit checker and redirect to new slugs"
+  task unpublish: [:environment] do
+    BrexitCheckerUnpublisher.call
+  end
+end
+
+class BrexitCheckerUnpublisher
+  def self.list
+    [{
+       content_id: "1102fd2b-7b29-43f5-889b-3e781e09971f",
+       new_base_path: "/transition-check/questions",
+       old_base_path: "/get-ready-brexit-check/questions",
+     },
+     {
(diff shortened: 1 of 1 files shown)


lib/tasks/brexit_checker_unpublisher.rake
```diff
@@ -0,0 +1,42 @@
...
+       new_base_path: "/transition-check/email-signup",
+       old_base_path: "/get-ready-brexit-check/email-signup",
+     }]
+  end
+
+  def self.call
+    list.each do |to_unpublish|
+      Services.publishing_api.unpublish(to_unpublish[:content_id],
+                                        type: "redirect",
+                                        explanation: "renamed to transition-check",
```
huwd: Will an exploitation be accepted for this type of unpublished? The docs seem to reckon it's only relevant for, withdrawal or gone? plus is only if it gets shown on a page?
```
 explanation (conditionally required)

    Required for a type of "withdrawal", Optional for a type of "gone".
    Message that will be displayed publicly on the page that has been unpublished.
```

huwd: https://docs.publishing.service.gov.uk/apis/publishing-api/api.html#post-v2contentcontent_idunpublish

koetsier: I'll remove it

PR #1050, opened 2020-02-13
Remove find-eu-exit-guidance-business.yml

huwd:
Trello: https://trello.com/c/MG1iAY1p/509-remove-business-readiness-finder-and-supporting-features-from-content-tagger

We should remove, not only the business finder but also the entire group-facets feature.

### DIFF
--- a/app/lib/services.rb
+++ b/app/lib/services.rb
@@ -4,7 +4,7 @@ require "gds_api/content_store"
 
 module Services
   def self.publishing_api
-    @publishing_api ||= GdsApi::PublishingApiV2.new(
+    @publishing_api ||= GdsApi::PublishingApi.new(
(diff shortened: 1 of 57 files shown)


huwd:
... at this point I'm wondering if the whole `module Facets` can go?

huwd:
@koetsier not quite sure what to do about this:
https://github.com/alphagov/content-tagger/blob/master/docs/arch/adr-001-links-based-facets.md

It was a really useful guide to how the feature came about. On the one hand we've removed the feature, but on the other I love having these archeology of past thought docs and can imagine future value.

I figure my options are:
- delete it, git has it forever
- amend it, add a post-script and change the title to say when these features were removed and why
- add a sibling, create a new ADR referencing that one that says why we ended up removing the feature.

PR #1180, opened 2021-03-18
Tidy README and reference GOV.UK Docker; remove startup.sh

kevindew:
Trello: https://trello.com/c/Psf7EzPB/204-move-away-from-startupsh

This updates this README to be more consistent with other GOV.UK apps and removes startup.sh as per https://github.com/alphagov/govuk-developer-docs/pull/2897. More details in the commits.


benthorner (approved):
Nice! I agree the screenshots thing has dubious value for the space it takes up, although I've generally left them in other repos where I've modified the README.

One extra thing we could add here would be a reference to the `docs/` directory, since this repo happens to have one already.

kevindew:
> Nice! I agree the screenshots thing has dubious value for the space it takes up, although I've generally left them in other repos where I've modified the README.
> 
> One extra thing we could add here would be a reference to the `docs/` directory, since this repo happens to have one already.

Thanks, I had a look in /docs and there's just one lone adr there so it doesn't really seem worth linking to bits in there.

PR #1186, opened 2021-03-26
Add rake task to update travel advice related links

brucebolt:
This updates the related links on all travel advice pages to include the "Travel abroad: step by step" as the second item in the related links.

These updates would normally be made through the Content Tagger UI, but there are 226 pages to update, so this would be a time consuming process.

This rake task has been [tested in the integration environment](https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/229638/console) and it can be removed after it has been run in production.

Trello card: https://trello.com/c/xu5zaFcw


thomasleese (approved):
Looks good!

PR #1204, opened 2021-05-17
Add override url to the taxon class

hannako:
## What

Make it possible to add an override_url field to taxons in content tagger.

- New input field on the Edit taxon page of content tagger.

### A user with Unreleased feature signon permissions:

Screenshot 2021-05-21 at 13 41 16

### A user without Unreleased feature signon permissions:

Screenshot 2021-05-21 at 13 44 38

### Field in [content item](https://www.integration.publishing.service.gov.uk/api/content/world/brexit-norway):

Screenshot 2021-05-21 at 13 46 31

Related https://github.com/alphagov/govuk-content-schemas/pull/1059

[Trello](https://trello.com/c/V2nmHHA6/1516-add-overrideurl-to-taxons)

### DIFF
--- a/spec/models/taxon_spec.rb
+++ b/spec/models/taxon_spec.rb
@@ -73,7 +73,7 @@ RSpec.describe Taxon do
       expect(valid_taxon).to be_valid
       expect(invalid_taxon).to_not be_valid
       expect(invalid_taxon.errors[:url_override]).to include(
-        "must be in the format '/prefix/slug'",
+        "must be in the format '/prefix/slug' or '/slug'",
       )
     end
   end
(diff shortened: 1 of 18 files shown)


spec/models/taxon_spec.rb
```diff
@@ -46,6 +46,38 @@
     end
   end
 
+  context "without url_override set" do
+    it "returns an empty string to comply with the schema definition" do
+      taxon = described_class.new
+
+      expect(taxon.url_override).to eq("")
```
sihugh, line 53: If we had this as `nil` would we be able to not set it at all in the content item? Ah, but it's being saved and retrieved from publishing-api, so that's not possible in this app, right?

I note that in https://github.com/alphagov/govuk_publishing_components/pull/2093, it would work the same whether the field is nil or an empty string because we're using `url_override.present?`, so it doesn't particularly matter on the frontend

sihugh (approved):

PR #1220, opened 2021-06-16
Use RestClient to get bulk upload CSVs

sihugh:
RestClient follows redirects by default. It looks like the behaviour for  published CSV docs has changed recently in that the published URL redirects to another short-lived location. Unfortunately, this breaks the workflow in this app because Net/HTTP doesn't follow redirects automatically.

This means we end up with lots of [errors in sentry](https://sentry.io/organizations/govuk/issues/2457278761/?project=202218&query=is%3Aunresolved) instead of tagged documents.

I've flipped some of the tests to use webmock instead of artificially producing results from Net/Http or whatever. It means we're less tied to an implementation and more accurately reflects how the system is behaving.

https://trello.com/c/0TPeCsrG/1525-create-two-brexit-child-taxons-for-businesses-and-citizens-and-tag-content-displaying-the-superbreadcrumb-to-relevant-child-taxo

### DIFF
--- a/app/lib/remote_csv.rb
+++ b/app/lib/remote_csv.rb
@@ -29,6 +29,6 @@ private
   end
 
   def response
-    Net::HTTP.get_response(URI(@csv_url))
+    RestClient.get @csv_url
   end
 end
--- a/app/services/bulk_tagging/fetch_remote_data.rb
+++ b/app/services/bulk_tagging/fetch_remote_data.rb
@@ -15,13 +15,12 @@ module BulkTagging
     end
 
     def call
-      unless valid_response?
-        GovukError.notify(RuntimeError.new(response.body))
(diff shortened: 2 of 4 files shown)


hannako (approved):
Looks good! We've lost a test that the client receives a request. [This](https://github.com/bblimke/webmock#setting-expectations-in-rspec-on-the-stub) sort of thing. But there's still a test that it returns the correct body so 🤷‍♀️

sihugh:
> Looks good! We've lost a test that the client receives a request. [This](https://github.com/bblimke/webmock#setting-expectations-in-rspec-on-the-stub) sort of thing. But there's still a test that it returns the correct body so 🤷‍♀️

Yeah, I deleted it because I didn't think it added anything. Sorry, should have called it out. Proves you read it though! 🙇

PR #1269, opened 2022-02-01
Upgrade Ruby to 2.7.5

ChrisBAshton:
Upgrades Ruby to 2.7.5. See commits for more details.
            This PR should be tested before merge, as per https://docs.publishing.service.gov.uk/manual/ruby.html#update-ruby-version-in-the-relevant-repos.


spec/validators/content_id_validator_spec.rb
```diff
@@ -1,17 +1,22 @@
 require "rails_helper"
 
 RSpec.describe ContentIdValidator do
+  let(:record_with_no_content_id) { build_stubbed(:tag_mapping) }
+  let(:valid_record) { build_stubbed(:tag_mapping) }
```
davidgisbey, line 5: Are people okay with coupling this a bit more tightly with TagMapping? Having a double with a pre-defined key in the errors felt a bit off.

ollietreend: @davidgisbey It looks like the other validators in `spec/validators/` also use doubles rather than stubbing specific models.

For example, isn't this similar to what the LinkTypeValidator spec does?

https://github.com/alphagov/content-tagger/blob/97603dca016acff9b903bd6f9fceaae1479058ca/spec/validators/link_type_validator_spec.rb#L4-L9

I might be completely misunderstanding, though 😅

Happy to talk it through if you want to. Or feel free to ignore if I'm being silly.

davidgisbey: As discussed i've made the changes to all the validators.

davidgisbey: Let me know if you're happy for this @ollietreend 👍

ollietreend (approved):

PR #1272, opened 2022-02-07
Upgrade to Rails 7.0.2

davidgisbey:
This bumps Rails to use the latest version.

I followed this guide https://guides.rubyonrails.org/upgrading_ruby_on_rails.html.

When i ran `rails app:update` there were several defaults that are different to ours. I've chosen to respect previous configuration. Does that seem right?

Trello card: 

https://trello.com/c/VUqyCoQ3/170-look-into-what-updating-all-apps-from-rails-61x-to-70x-would-involve


config/environments/development.rb
```diff
@@ -14,6 +14,11 @@
   # Show full error reports.
   config.consider_all_requests_local = true
 
+  config.active_record.verbose_query_logs = true
+
+  # Suppress logger output for asset requests.
+  config.assets.quiet = true
```
davidgisbey, line 20: and this?

Gemfile
```diff
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem "rails", "6.1.4.4"
```
ollietreend, line 3: Should we also be adding `sprockets-rails` as an explicit dependency here?

From the [upgrade guide](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#sprockets-is-now-an-optional-dependency):
> **Sprockets is now an optional dependency**
> The gem rails doesn't depend on sprockets-rails anymore. If your application still needs to use Sprockets, make sure to add sprockets-rails to your Gemfile.

It looks like `sprockets` is still pulled in as a dependency of the `sassc-rails` gem, so maybe that's good enough..? 🤔

kevindew: That's a good shout. As it's an explicit dependency of the app, we should call it out - especially as we use it for more than sassc.

What's also interesting is that this app uses Sprockets 4, whereas most of our apps don't because govuk_publishing_components requires sprockets < 4 and nearly all GOV.UK apps use that: https://github.com/alphagov/govuk_publishing_components/pull/1172.

config/application.rb
```diff
@@ -14,7 +14,7 @@
 module ContentTagger
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 6.1
+    config.load_defaults 7.0
```
kevindew, line 17: Was there an equivalent of the new_framework_defaults file for 7? I have recollections in the past that some (particularly those related to cookies) sometimes need you to run the newer version before changing the defaults.

Be good if this upgrade didn't have as it's a bit of a pain doing the upgrade in two steps.

config/environments/development.rb
```diff
@@ -14,6 +14,11 @@
   # Show full error reports.
   config.consider_all_requests_local = true
 
+  config.active_record.verbose_query_logs = true
```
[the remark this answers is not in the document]
kevindew: This looks a nice feature: https://edgeguides.rubyonrails.org/debugging_rails_applications.html#verbose-query-logs 

as it's only on in dev the impact of this is also low for any problems.

config/environments/development.rb
kevindew: This looks like something that is set in most of all ours apps already (Rails default). If I remember right you just don't get asset requests in your Rails development server log which cuts out a lot of noise.

config/initializers/new_framework_defaults_6_0.rb
```diff
@@ -1,45 +0,0 @@
...
-# will be removed in Rails 6.1. This setting is not backwards compatible with earlier Rails versions.
-# If you send mail in the background, job workers need to have a copy of
-# MailDeliveryJob to ensure all delivery jobs are processed properly.
-# Make sure your entire app is migrated and stable on 6.0 before using this setting.
-# Rails.application.config.action_mailer.delivery_job = "ActionMailer::MailDeliveryJob"
-
-# Enable the same cache key to be reused when the object being cached of type
-# `ActiveRecord::Relation` changes by moving the volatile information (max updated at and count)
-# of the relation's cache key into the cache version to support recycling cache key.
-# Rails.application.config.active_record.collection_cache_versioning = true
```
kevindew, line 45: Nice one deleting this.

app/views/project_content_items/index.html.erb
```diff
@@ -14,7 +14,7 @@
     <tr>
       <td><%= project_content_item.suggested_tags %></td>
       <td><%= link_to project_content_item.title, project_content_item.url %></td>
-      <td><%= project_content_item.updated_at.to_s(:govuk_date) %></td>
+      <td><%= project_content_item.updated_at.to_formatted_s(:govuk_date) %></td>
```
kevindew, line 17: Nice that this is the only deprecation.

kevindew (commented):
This looks good. It looks like it might be our least painful Rails major version upgrade (touch wood).

I did find that when I tried `rails app:update` locally there were a few other minor changes Rails applied to config files and bin's that'd be nice to haves.

PR #1292, opened 2022-03-28
Relax URL override regex

sihugh:
We were using the same regex as taxons use, which is limited to two path
segements:
- /foo
- /foo/bar

The URL override path shouldn't be limited in this way, as we should be
allowed deeper - many Whitehall URLs are several layers deep. This change (switching the last `?` to a `*`)
should allow:
- /foo
- /foo/bar
- /foo/bar/something/else

I indented the regexes to emphasise how similar they are as they only differ by one character.

### DIFF
--- a/app/models/taxon.rb
+++ b/app/models/taxon.rb
@@ -1,5 +1,6 @@
 class Taxon
-  PATH_COMPONENTS_REGEX = %r{\A/(?<prefix>[A-z0-9\-]+)(?:/(?<slug>[A-z0-9\-]+))?\z}.freeze
+  PATH_COMPONENTS_REGEX          = %r{\A/(?<prefix>[A-z0-9\-]+)(?:/(?<slug>[A-z0-9\-]+))?\z}.freeze
+  OVERRIDE_PATH_COMPONENTS_REGEX = %r{\A/(?<prefix>[A-z0-9\-]+)(?:/(?<slug>[A-z0-9\-]+))*\z}.freeze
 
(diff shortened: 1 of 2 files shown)


Tetrino (approved):

spec/lib/tasks/taxonomy/health_checks_spec.rb
```diff
@@ -1,20 +1,18 @@
...
 
   it "queues the expected jobs" do
     ActiveJob::Base.queue_adapter = :test
 
-    expect {
-      rake "taxonomy:health_checks"
-    }.to change(TaxonomyHealth::MaximumDepthMetric.jobs, :size).by(1)
-    .and change(TaxonomyHealth::ContentCountMetric.jobs, :size).by(1)
-    .and change(TaxonomyHealth::ChildTaxonCountMetric.jobs, :size).by(1)
+    Sidekiq::Testing.fake! do
```
hannako, line 10: Ah this is cool, I didn't know about this API!

hannako (approved):
Lovely!

spec/lib/tasks/taxonomy/health_checks_spec.rb
sihugh: 💯 
I'd started off by running the test with `Sidekiq::Testing::inline!` and adding stubs, but what this needed was for things to _not_ be inline so they could go on the queue (which is what the tests want to measure).

PR #1293, opened 2022-03-28
Remove logic around updating /brexit.cy

sihugh:
This is the only Welsh taxon. It's not even a real taxon, it's just here to provide a route. 

It has been manually unpublished and redirected because we don't need it anymore, but we need to unpick this link so that we don't resurrect it accidentally by updating the English Brexit taxon.


hannako (approved):
Wow I had no idea this was how we were doing it! ✂️

PR #1307, opened 2022-04-13
Use consistent language to describe tagging and topic pages

AgaDufrat:
## Context 
Clear consistent language in the publishing apps will help publishers tag content better, and understand the implications of this. Better tagging will make their content more findable.

Edit taxonomy flow is out of scope

Trello: https://trello.com/c/8urAxFrT/927-content-tagger-update-copy-based-on-jenns-doc

## Changes proposed 

Update the copy as per this doc: https://docs.google.com/spreadsheets/d/1OLw4B46ZqTgsuo762Kjxb-M6_rfX_dW23qp8bNhZrMo/edit#gid=994922232

| Before  | After |
| ------------- | ------------- |
| Screenshot 2022-04-26 at 17 35 03 | Screenshot 2022-04-26 at 17 27 30 |
| Screenshot 2022-04-26 at 17 35 16 | Screenshot 2022-04-26 at 17 28 03 |

## How to review this

- visit `/tagging/:content_id` path

### DIFF
--- a/app/views/taggings/_form_for_topics.html.erb
+++ b/app/views/taggings/_form_for_topics.html.erb
@@ -1,16 +1,15 @@
-<h3>Topics</h3>
+<h3>Specialist topic pages</h3>
 
 <%= f.input :topics,
     collection: linkables.topics, as: :grouped_select, group_method: :last,
-    placeholder: "Choose Topics...",
+    placeholder: "Choose specialist topic pages...",
     input_html: { multiple: true, include_blank: true, class: :select2 } %>
 
 <p class='explain'>
-  Topic pages live under <a href="https://www.gov.uk/topic">/topic</a>.
-  They contain collections of pages on GOV.UK. Example:
+  Specialist topic pages live under <a href="https://www.gov.uk/topic">/topic</a>.
+  They contain pages on GOV.UK that have been tagged to the specialist topic. Example:
   <a href="https://www.gov.uk/topic/business-tax">Business Tax</a> or
   <a href="https://www.gov.uk/topic/animal-welfare/pets">Animal Welfare / Pets</a>.
-  Publishing a page with a topic may send out an
+  Publishing a page with a specialist topic tag may send out an
   <a href="https://www.gov.uk/topic/business-tax/vat/email-signup">email alert to subscribers</a>.
-  Formerly called <i>specialist sectors</i>.
 </p>
--- a/app/services/linkables.rb
+++ b/app/services/linkables.rb
@@ -39,7 +39,7 @@ private
   end
 
   def for_nested_document_type(document_type)
-    # In Topics and Browse pages, the "internal name" is generated in the
+    # In Specialist topic pages and Mainstream Browse pages, the "internal name" is generated in the
     # form: "Parent title / Child title". Because currently we only show
     # documents on child-topic pages (like /topic/animal-welfare/pets), we
     # only allow tagging to those tags in this application. That's why we
--- a/app/views/taggings/_form_for_mainstream_browse_pages.html.erb
+++ b/app/views/taggings/_form_for_mainstream_browse_pages.html.erb
@@ -9,6 +9,5 @@
   Mainstream browse pages live under <a href="https://www.gov.uk/browse">/browse</a>.
   Examples: <a href="https://www.gov.uk/browse/working">Working</a> or
   <a href="https://www.gov.uk/browse/business">Business</a>.
-  They are used to organise mainstream content. Formerly called
-  <i>sections and subsections</i>.
+  They are used to organise curated content.
 </p>
--- a/app/views/taggings/_form_for_organisations.html.erb
+++ b/app/views/taggings/_form_for_organisations.html.erb
@@ -6,6 +6,6 @@
     input_html: { multiple: true, class: :select2 } %>
 
 <p class='explain'>
-  Tagging a page to an organisation will make it show up in search.
+  Tagging a page to an organisation will make it show up in search when filtered to the organisation(s). 
   Example: <a href="https://www.gov.uk/search?q=taxes&filter_organisations%5B%5D=hm-revenue-customs">search for documents published by HMRC</a>.
 </p>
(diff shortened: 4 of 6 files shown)


app/views/taggings/_form_for_topics.html.erb
```diff
@@ -1,16 +1,15 @@
-<h3>Topics</h3>
+<h3>Specialist topic pages</h3>
 
 <%= f.input :topics,
     collection: linkables.topics, as: :grouped_select, group_method: :last,
-    placeholder: "Choose Topics...",
+    placeholder: "Choose Specialist topic pages...",
```
hannako, line 5: I'm sure this is totally fine, but just flagging that there's no note that this placeholder text should be updated in Jen's document. https://docs.google.com/spreadsheets/d/1OLw4B46ZqTgsuo762Kjxb-M6_rfX_dW23qp8bNhZrMo/edit#gid=994922232 Maybe just triple check!

hannako (commented):
This looks good, but it's hard to be sure there's nothing missing without clicking around in the UI a bit. Did you put this branch on integration? If you have time, it might be nice to add some screen shots to the PR description if poss.

hannako (approved):
Lovely job!

AgaDufrat:
I understand your point about the old terminology appearing in other places in the app. We discussed in the team but decided it's out of scope for this piece of work. Please see Jenn's comment justification:
> For this piece of work, we're really only concerned with the terminology that affects the tagging of mainstream and Whitehall content to topic pages. In Content Tagger, this falls under the 'Edit a page' screen
 `https://content-tagger.integration.publishing.service.gov.uk/taggings/lookup` and 
 `https://content-tagger.integration.publishing.service.gov.uk/taggings/:content_id`

> Other taxonomy-related screens in Content Tagger are not visible to most publishers, and therefore don't impact their work directly. The core audience for Bulk Tag, Edit Taxonomy, Tagging History, Taxonomy health warnings and Projects is content strategists/taxonomy folks/data folks. That means it's not as important to change 'taxon' related terminology in these contexts, until we can directly address the future of the taxonomy.

> For now, let's exclude:
/taxons
/bulk_tag/new
/taxonomy/health_warnings
/tagging_history
/projects
These need to be reviewed in future with those folks most likely to use them; we might not need all of this functionality in Content Tagger at all.

But we are happy to discuss.

kevindew:
> I understand your point about the old terminology appearing in other places in the app. We discussed in the team but decided it's out of scope for this piece of work. Please see Jenn's comment justification:
> 
> > For this piece of work, we're really only concerned with the terminology that affects the tagging of mainstream and Whitehall content to topic pages. In Content Tagger, this falls under the 'Edit a page' screen
> > `https://content-tagger.integration.publishing.service.gov.uk/taggings/lookup` and
> > `https://content-tagger.integration.publishing.service.gov.uk/taggings/:content_id`
> 
> > Other taxonomy-related screens in Content Tagger are not visible to most publishers, and therefore don't impact their work directly. The core audience for Bulk Tag, Edit Taxonomy, Tagging History, Taxonomy health warnings and Projects is content strategists/taxonomy folks/data folks. That means it's not as important to change 'taxon' related terminology in these contexts, until we can directly address the future of the taxonomy.
> 
> > For now, let's exclude:
> > /taxons
> > /bulk_tag/new
> > /taxonomy/health_warnings
> > /tagging_history
> > /projects
> > These need to be reviewed in future with those folks most likely to use them; we might not need all of this functionality in Content Tagger at all.
> 
> But we are happy to discuss.

Sure, makes sense that it's out of scope. Were all the areas I found out of scope?

Is resolving these inconsistencies something that you want to own as part of this piece of work or is this something that Publishing will need to capture?

kevindew:
> > Were all the areas I found out of scope?
> 
> When picking this work I did consider colsolidating the terminology in the code as well i.e. rename URL/routes, params/keys, models, method/class names, view and partials files, when appropriate.
> 
> However the content schema for the “Specialist topic” is still called “[topic](https://docs.publishing.service.gov.uk/content-schemas/topic.html)” so I decided to keep those references as they are. There will be an opportunity to change URLs (and references in the code) in Phase 2 in Collections Publisher and I flagged it with Jenn since URLs are “user facing”.
> 
> I double checked your list (thanks for pulling it out).  I can confirm that
> 
> * we are intentionally not changing:
>   
>   * app/views/project_content_items/*
>   * app/views/bulk_tags/*
>   * app/views/projects/*
>   * app/views/taxons/*
>   * config/locales/en.yml (“Edit taxonomy” copy)
> * I made the necessary changes in (was the grep on main? 🤔 ):   
>   
>   * [app/views/taggings/_form_for_topics.html.erb](https://github.com/alphagov/content-tagger/blob/0cdc3ba1bd46cc0906fdc13648d7d07a11fa3efc/app/views/taggings/_form_for_topics.html.erb)
>   * https://github.com/alphagov/content-tagger/blob/0cdc3ba1bd46cc0906fdc13648d7d07a11fa3efc/app/services/linkables.rb#L42
>   * https://github.com/alphagov/content-tagger/blob/0cdc3ba1bd46cc0906fdc13648d7d07a11fa3efc/config/blacklisted_tag_types.yml#L19
> 
> > Is resolving these inconsistencies something that you want to own as part of this piece of work or is this something that Publishing will need to capture?
> 
> I'd suggest moving the conversation about this to Slack, so we can get Sam and Jenn looped in, but in short — it's something that we think "Find and view" should own.

Argh very sorry if I included ones you'd already changed. Not too sure how it happened but you're probably right about main. I think I did a grep against main then one against your branch and may have got muddled in the pasting process 🤦 

Cool it seems good if Find & View are keeping track of the inconsistencies. There probably needs to be a larger discussion about the future of content tagger too since it's a rather fishy app.

jennpb:
I'm happy with Aga's changes and exclusions. 

> Cool it seems good if Find & View are keeping track of the inconsistencies. There probably needs to be a larger discussion about the future of content tagger too since it's a rather fishy app.

I will keep track of this and the need for a broader discussion as a tie-in with future taxonomy review.

PR #1333, opened 2022-07-12
Improve test coverage for continuous deployment

benjamineskola:
A lot of these are for rake tasks; ideally the logic should probably be moved out of the take task entirely but for the most part I've tested them as-is (mainly just testing that the library functions are called as appropriate, because most of the logic is already tested).

Other changes:

- RakeTaskHelper was breaking test coverage by reinitialising the rake environment for every test, this has now been simplified;
- One of the rake tasks was calling `return`, which is invalid since a rake task is not a function. It did have the desired effect of aborting the task early, but only because it was throwing an exception about the invalid `return`. This is now fixed to use `abort`.

https://trello.com/c/3xZQq6rw/499-enable-continuous-deployment-for-content-tagger

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -40,6 +40,7 @@ end
 
 group :test do
   gem "database_cleaner"
+  gem "fakefs", require: "fakefs/safe"
   gem "govuk_test"
   gem "timecop"
   gem "webmock"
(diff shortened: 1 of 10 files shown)


beccapearce (approved):
LGTM

PR #1344, opened 2022-07-28
Rework test for taxonomy:untag to avoid intermittent failure

benjamineskola:
This task exits early which works oddly in Rake. Previously used `return` which is not valid (but worked by accident, because it caused a runtime error). `abort` appeared to work but fails intermittently in CI, perhaps due to test ordering? Stack Overflow recommends `next` instead. Then the test can just ensure that the untagger is never called rather than checking for a `SystemExit`.


benjamineskola:
Superseded by #1345 because I accidentally created this one in a fork(thanks GitHub cli)

PR #1372, opened 2022-09-29
Update rake task to handle different situations

chao-xian:
Running on staging found different data states that meant things errored and the rake task would bomb. Hopefully these changes will handle things ok.

### DIFF
--- a/lib/tasks/travel_advice/tmp_swap_travel_advice_link.rake
+++ b/lib/tasks/travel_advice/tmp_swap_travel_advice_link.rake
@@ -15,15 +15,23 @@ namespace :travel_advice do
       links = Services.publishing_api.get_links(edition["content_id"])["links"]
       next unless links && links["ordered_related_items"]
 
-      index_of_link = links["ordered_related_items"].index(take_out)
+      p "Updating links for item #{edition['content_id']}"
 
-      links["ordered_related_items"][index_of_link] = swap_in if index_of_link
+      begin
+        index_of_link = links["ordered_related_items"].index(take_out)
 
-      Services.publishing_api.patch_links(
-        edition["content_id"],
-        links: links,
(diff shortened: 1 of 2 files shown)


spec/lib/tasks/tmp_swap_travel_advice_link_spec.rb
```diff
@@ -71,4 +71,46 @@
...
   end
+
+  context "another item doesn't have any links" do
+    let(:no_links_edition) do
+      {
+        content_id: SecureRandom.uuid,
+        document_type: "travel_advice",
+        publishing_app: "travel-advice-publisher",
+      }
+    end
```
leenagupte, line 82: Don't you also have to stub the publishing-api call too `get_editions` for `no_links_edition` as well?

lib/tasks/travel_advice/tmp_swap_travel_advice_link.rake
```diff
@@ -15,15 +15,20 @@ namespace :travel_advice do
...
-        edition["content_id"],
-        links: links,
-        bulk_publishing: true,
-      )
+        Services.publishing_api.patch_links(
+          edition["content_id"],
+          links: links,
+          bulk_publishing: true,
+        )
+      rescue StandardError
```
leenagupte, line 28: Should this be more specific? i.e. catching the GdsApi error rather than everything?

leenagupte (commented):
Two questions, but the plan looks good.

lib/tasks/travel_advice/tmp_swap_travel_advice_link.rake
chao-xian: Oh yeah - thanks

leenagupte (approved):
👍

PR #1378, opened 2022-10-05
Revert "Update rake task to handle different situations"

chao-xian:
Reverts alphagov/content-tagger#1372

We've run this temporary rake task and no longer need it.


chao-xian:
This actually needs to undo 2 PRs

PR #1389, opened 2022-10-20
Some TLC to the test suite

kevindew:
Trello: https://trello.com/c/QrxjshEm/270-move-ruby-gems-from-jenkins-to-github-actions

This resolves a number of undesirable aspects of this test suite that I encountered when trying to port some functionality ([govuk_taxonomy_helpers](https://github.com/alphagov/govuk_taxonomy_helpers)) into this app. It resolves things such as: deprecation warnings, tests writing to stdout/stderr, unusual RSpec configuration for GOV.UK, unnecessary rubygems and unnecessary RSpec declarations. For full details see the individual commit messages.

It is recommended to review this commit by commit, it's much less scary that way.

I also noticed there wasn't rspec linting applied and will resolve that in a subsequent pull request that builds off this one.


spec/spec_helper.rb
```diff
@@ -42,22 +40,6 @@
...
 
   config.fixture_path = "#{::Rails.root}/spec/fixtures"
-  config.use_transactional_fixtures = false
+  config.use_transactional_fixtures = true
   config.infer_spec_type_from_file_location!
-
-  config.before(:each) do
-    User.create!(permissions: ["signin", "GDS Editor"])
-    DatabaseCleaner.start
-    WebMock.reset!
```
ollietreend: Shouldn't `User.create!` and `WebMock.reset!` still exist here? I'm trying to work out if/why they're not needed any more.

I was expecting for this commit to only remove things relating to `DatabaseCleaner`.

kevindew: Ah yes, sorry I should have covered those details.

I've amended this commit now to explain. 

Essentially we don't need the user because the db:seeds create it (and we certainly didn't need one for every test 😢).

Webmock.reset! is done automatically by Webmock so we don't need to call it ourselves.

ollietreend (approved):
Really nice ✨🚀 thanks for this!

PR #1465, opened 2023-02-16
Upgrade Yarn to v3.4.1

deborahchua:
Version 1 of Yarn is no longer reliably supported, and is only receiving occasional hot fixes on a ‘best efforts’ basis.

Version 3 has been recommended as it has better backwards compatability. 
See [Yarn's migration steps](https://yarnpkg.com/getting-started/migration) and individual commits for more details.

Trello card: https://trello.com/c/Mn4SWQwj/3093-investigate-whats-required-to-move-off-yarn-1x


.yarnrc.yml
```diff
@@ -1,2 +1,13 @@
...
+packageExtensions:
+  stylelint-config-recommended-scss@*:
+    peerDependencies:
+      postcss: '8'
+  stylelint-config-standard-scss@*:
+    peerDependencies:
+      postcss: '8'
+  stylelint-config-gds@*:
+    peerDependencies:
+      postcss: '8'
```
ChrisBAshton, line 13: The `.yarnrc.yml` change is one of the suggested fixes from the linked GitHub issue.

The other suggestion is to [define `peerDependencies` in package.json](https://github.com/stylelint-scss/stylelint-config-recommended-scss/pull/112/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R70-R76). We [already have `postcss` defined as a dev dependency](https://github.com/alphagov/content-tagger/blob/main/package.json#L29-L30) so it would be a case of either duplicating or moving that value.

Did you look into that option too? I'm not really sure which is best, but it certainly feels neater to me that all the dependency related definitions are all contained in the same file (`package.json`).

.yarnrc.yml
```diff
@@ -1 +1,2 @@
 yarnPath: .yarn/releases/yarn-3.4.1.cjs
+nodeLinker: node-modules
```
ChrisBAshton, line 2: After the `yarn install` step, looks like we need to update the `.gitignore` file as per [Which files should be gitignored?](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored) (the "not using Zero-Installs" option, [since we're not using Plug'n'Play](https://yarnpkg.com/features/zero-installs#how-do-you-reach-this-zero-install-state-youre-advocating-for))

ChrisBAshton (commented):
Nice one - looks good.

I wasn't sure if the `.yarn/releases/` file should remain in git history, but [it should](https://yarnpkg.com/getting-started/qa):

> `.yarn/plugins` and `.yarn/releases` contain the Yarn releases used in the current repository (as defined by [yarn set version](https://yarnpkg.com/cli/set/version)). You will want to keep them versioned (this prevents potential issues if, say, two engineers use different Yarn versions with different features).

So just a couple of comments below.

.yarnrc.yml
deborahchua: Thanks @ChrisBAshton I've updated this in https://github.com/alphagov/content-tagger/pull/1465/commits/d1ef8dc99416df113816a61aada035a2696d4d02

.yarnrc.yml
deborahchua: Yep I did try the `peerDependencies` solution first but for some reason it didn't work!

PR #1587, opened 2023-11-07
Revert "Bump bootsnap from 1.16.0 to 1.17.0"

andy-collon:
Reverts alphagov/content-tagger#1584
We've been getting Failed to open TCP connection to localhost:9091 errors in integration, staging and production since yesterday just after this was released to staging and production, though integration was released on the 1st November. Reverting to rule it out as root cause


anatron (approved):

andy-collon:
The issue was identified as due to missing configuration which has been added to the helm charts

PR #1601, opened 2023-12-06
Update to Ruby 3.2.2

mtaylorgds:
There has been a change in how Ruby prints errors, which seems to have impacted a test that was checking stdout. The Ruby change discussion talks about printing unhandled errors, and in our case we're printing the results of an exception's `#inspect` method in a `rescue` block, but it seems too closely-related to be a coincidence.

Aside from making the output go all on a single line (so the newlines in the HEREDOC were being removed), the escape sequences were being double-escaped (so every `\"` became `\\"`). Putting single-quotes around the opening identifier avoids this issue.

Ruby language change discussion:
https://bugs.ruby-lang.org/issues/18367

[Trello card](https://trello.com/c/vmM1yVzB)


cynthia-anya (approved):
LGTM

PR #1768, opened 2024-08-08
Fix broken icons

andysellick:

## What / why
- fixes broken icons across the site
- icons are glyphicons from bootstrap, which are included as part of that dependency and can't be directly modified
- problem is caused by the compiled Sass not being able to find and download the icon font files e.g. https://content-tagger.publishing.service.gov.uk/assets/content-tagger/bootstrap/glyphicons-halflings-regular.eot
- this appears to be because the syntax in bootstrap for defining these files uses `url(font-path(path))` which doesn't seem to work anymore, could be related to the recent dart-sass migration but unclear
- as a workaround have copied the font-face declaration from the bootstrap CSS into our own, and replaced the above with `url(path)`, which works
- this occurs after the bootstrap declaration, overriding it and allowing the icon font file to be found and loaded correctly

## Visual changes

Before
Screenshot 2024-08-08 at 16 12 33

After
Screenshot 2024-08-08 at 16 12 40

Trello card: https://trello.com/c/gW2NW1sB/239-fix-sass-compilation-warnings

### DIFF
--- a/app/assets/javascripts/views/related-content-tagger.js
+++ b/app/assets/javascripts/views/related-content-tagger.js
@@ -10,7 +10,7 @@
       var $templateTag = $fieldset.find('.related-item-template > li').first()
       var $fieldErrors = $fieldset.find('.related-item-error-message')
 
-      $fieldset.on('click', '.select2-search-choice-close', removeItem)
+      $fieldset.on('click', '.js-remove-related', removeItem)
       $basePathInput.on('keypress', lookUpBasePathOnEnterPress)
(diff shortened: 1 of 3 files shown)


andysellick:
Think I've fixed all the icons now. Unfortunately they're mostly in external dependencies, which aren't fully compatible with the new dart-sass compilation, meaning we can't access that code to fix it. I've had to add some unfortunately quite specific overrides in our our Sass, the problem is around background images originally written using `image-url` instead of `url`.

mtaylorgds (approved):
LGTM—great work figuring this out!

PR #1900, opened 2025-01-20
World taxon rake update code extracted to class

ellohez:
Implements the testing for the Rake task which will check all of the [World taxonomy titles]
(https://docs.publishing.service.gov.uk/manual/world-taxonomy.html) adding the name of the country they relate to where necessary. [Trello card](https://trello.com/c/24XeROPv). All of the world taxons are descendants of [/world/all content item](https://www.gov.uk/api/content/world/all)

For accessibility reasons, taxon titles need to be more unique. Having multiple versions of pages for different
countries all titled e.g. 'Trade and invest' goes against the following WCAG guidelines.
https://www.w3.org/WAI/WCAG22/Understanding/page-titled.html
https://www.w3.org/WAI/WCAG22/Techniques/failures/F25 

Essentially, world location pages such as:[ gov.uk/world/japan](https://www.gov.uk/world/japan) include links to sections such as 'Coming to the UK' or 'Birth, death and marriage abroad'. These sections can however also appear as pages in their own right, for example, they can appear as a link under 'Explore the topic' where the title makes the content sound generic but actually links to specific information relating to that country. E.g. the 'Coming to the UK' link at the bottom of this page: https://www.gov.uk/government/publications/tuberculosis-test-for-a-uk-visa-clinics-in-cameroon actually links to the dedicated page -[ https://www.gov.uk/world/coming-to-the-uk-cameroon](https://www.gov.uk/world/coming-to-the-uk-cameroon)

Screenshot 2025-01-21 at 09 59 19

Screenshot 2025-01-21 at 10 01 05

For titles which need to be updated, this task takes the name of the country from the internal name and then combines it with a relevant word or symbol to create a full title:
E.g. 

> Birth, death and marriage abroad in Cameroon
`+ in COUNTRY_NAME`
or

> Coming to the UK from Cameroon
`+ from COUNTRY_NAME`

All template versions of pages, the root and any taxon that already includes the country name in the title, are skipped and not processed. 


spec/lib/tasks/world_taxon_update_helper_spec.rb
```diff
@@ -0,0 +1,268 @@
...
+
+    let(:retitled_child_taxon_coming_to) do
+      FactoryBot.build(:taxon, content_id: "id3", title: "Coming to the UK from Porpoise Spit")
+    end
+    let(:retitled_child_taxon_trade_invest) do
+      FactoryBot.build(:taxon, content_id: "id4", title: "Trade and invest: Porpoise Spit")
+    end
+    let(:retitled_child_taxon_birth_death) do
+      FactoryBot.build(:taxon, content_id: "id5", title: "Birth, death and marriage abroad in Porpoise Spit")
+    end
```
mtaylorgds, line 227: Similar to the tests for add, I think these might as well be moved to the one test that really uses them ("correctly updates the taxon title, removing the country name and any added words etc").

See comment below for dealing with the one other case on line 238.

spec/lib/tasks/world_taxon_update_helper_spec.rb
```diff
@@ -0,0 +1,268 @@
...
+    it "logs an error to stderr if the taxon is invalid" do
+      allow_item_double(retitled_multi_level_linked_items)
+      allow(Taxonomy::BuildTaxon).to receive(:call).and_raise(Taxonomy::UpdateTaxon::InvalidTaxonError.new("Invalid taxon"))
+
+      expect { described_class.new.remove_country_names("tmp/rake_log") }.to output(/An error occurred while processing taxon/).to_stderr
+    end
+
+    it "logs an error to stderr if BulkPublishTaxon fails" do
+      allow_item_double(retitled_multi_level_linked_items)
+      allow(Taxonomy::BuildTaxon).to receive(:call).and_return(retitled_child_taxon_coming_to)
```
mtaylorgds, line 238: Similar to the test for add, this test doesn't really care what the taxon returned by BuildTaxon is, so we could change this line (which then allows the 3 variables at the top of the describe block to be moved down into the remaining test that uses them).
```suggestion
      allow(Taxonomy::BuildTaxon).to receive(:call).and_return(FactoryBot.build(:taxon))
```

spec/lib/tasks/world_taxon_update_helper_spec.rb
```diff
@@ -0,0 +1,268 @@
...
+
+      expect { described_class.new.remove_country_names("tmp/rake_log") }.to output(/An unexpected error occurred/).to_stderr
+    end
+
+    it "correctly updates the taxon title, removing the country name and any added words etc" do
+      allow_item_double(retitled_multi_level_linked_items)
+      allow(Taxonomy::BuildTaxon).to receive(:call).with(content_id: retitled_child_item_coming_to.content_id).and_return(retitled_child_taxon_coming_to)
+      allow(Taxonomy::BuildTaxon).to receive(:call).with(content_id: retitled_child_item_trade_invest.content_id).and_return(retitled_child_taxon_trade_invest)
+      allow(Taxonomy::BuildTaxon).to receive(:call).with(content_id: retitled_child_item_birth_death.content_id).and_return(retitled_child_taxon_birth_death)
+      allow(Taxonomy::BulkPublishTaxon).to receive(:call)
```
mtaylorgds, line 257: This allow is covered in the before block, so it doesn't need doing again here.
```suggestion
```

spec/lib/tasks/world_taxon_update_helper_spec.rb
```diff
@@ -0,0 +1,268 @@
...
+      FactoryBot.build(:taxon, content_id: "id3", title: "Coming to the UK from Porpoise Spit")
+    end
+    let(:retitled_child_taxon_trade_invest) do
+      FactoryBot.build(:taxon, content_id: "id4", title: "Trade and invest: Porpoise Spit")
+    end
+    let(:retitled_child_taxon_birth_death) do
+      FactoryBot.build(:taxon, content_id: "id5", title: "Birth, death and marriage abroad in Porpoise Spit")
+    end
+
+    it "logs an error to stderr if the taxon is invalid" do
```
mtaylorgds, line 229: Similar to the tests for add, I think we're maybe missing a test to check that when this error occurs, it goes onto the next taxon (the difference between being caught by the inner rescue block and the outer one).

mtaylorgds (commented):
Good work on this—I know I've added quite a few comments, but they are mostly pretty minor!

spec/lib/tasks/world_taxon_update_helper_spec.rb
```diff
@@ -0,0 +1,268 @@
...
+    parent_item_uk_help << child_item_coming_to << child_item_trade_invest << child_item_birth_death
+
+    root_item
+  end
+
+  before do
+    allow(Taxonomy::BulkPublishTaxon).to receive(:call)
+  end
+
+  def allow_item_double(item)
```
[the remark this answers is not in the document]
ellohez: Good point!

spec/lib/tasks/world_taxon_update_helper_spec.rb
```diff
@@ -0,0 +1,268 @@
...
+    child_double = instance_double(Taxonomy::ExpandedTaxonomy, child_expansion: item)
+    item_double = instance_double(Taxonomy::ExpandedTaxonomy, build: child_double)
```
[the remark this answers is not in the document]
ellohez: Seems beneficial to model the real behaviour more closely.

lib/world_taxon_update_helper.rb
```diff
@@ -0,0 +1,162 @@
...
+  rescue StandardError => e
+    log_rake_error(log_file, "An error occurred while publishing taxons: #{e.full_message}")
+  ensure
+    log_file&.close
+  end
+
+private
+
+  def create_new_taxon_title(internal_name)
+    # -------------
```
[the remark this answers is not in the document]
ellohez: This is confusing but at the start of the process the title (aka external name) doesn't actually include the country name at all but the internal name does (in brackets).  So the code takes:

internal_name: Coming to the UK (Armenia)
title: Coming to the UK

and turns it into:

internal_name: Coming to the UK (Armenia)
title: Coming to the UK from Armenia

So I clearly need to explain that better in the comments!

spec/lib/tasks/world_taxon_update_helper_spec.rb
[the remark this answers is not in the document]
ellohez: Changing this causes 11 of the 14 tests to error with WebMock::NetConnectNotAllowedError so have left this as was for now.

spec/lib/tasks/world_taxon_update_helper_spec.rb
```diff
@@ -0,0 +1,268 @@
...
+  describe "add_country_name_to_title" do
+    let(:child_taxon_coming_to) do
+      FactoryBot.build(:taxon, content_id: "id3", title: "Coming to the UK")
+    end
+    let(:child_taxon_trade_invest) do
+      FactoryBot.build(:taxon, content_id: "id4", title: "Trade and invest")
+    end
+    let(:child_taxon_birth_death) do
+      FactoryBot.build(:taxon, content_id: "id5", title: "Birth, death and marriage abroad")
+    end
```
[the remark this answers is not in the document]
ellohez: Moving them to inside the `it` block causes the following error: 

>   `let` is not available from within an example (e.g. an `it` block) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc). It is only available on an example group (e.g. a `describe` or `context` block).

So, I have moved them to just before the test they are used in, so outside of the `it` to sit within the `describe` but not at the top.

spec/lib/tasks/world_taxon_update_helper_spec.rb
ellohez: See comment above re. `let` not being allowed in `it` blocks.

spec/lib/tasks/world_taxon_update_helper_spec.rb
```diff
@@ -0,0 +1,268 @@
...
+    it "logs an error to stderr if the taxon is invalid" do
+      allow_item_double(multi_level_linked_items)
+      allow(Taxonomy::BuildTaxon).to receive(:call).and_raise(Taxonomy::UpdateTaxon::InvalidTaxonError.new("Invalid taxon"))
+
+      expect { described_class.new.add_country_names("tmp/rake_log") }.to output(/An error occurred while processing taxon/).to_stderr
+    end
+
+    it "logs an error to stderr if BulkPublishTaxon fails" do
+      allow_item_double(multi_level_linked_items)
+      allow(Taxonomy::BuildTaxon).to receive(:call).and_return(child_taxon_coming_to)
```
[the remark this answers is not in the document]
ellohez: Have made this change but only moved the 3 taxon variables to just above the `it` block where they are used, see previous comment about `let` not being available from within `it` blocks.

spec/lib/tasks/world_taxon_update_helper_spec.rb
```diff
@@ -0,0 +1,268 @@
...
+
+      # Tests all three title change possibilities
+      expect(Taxonomy::UpdateTaxon).to have_received(:call).with(taxon: having_attributes(content_id: child_taxon_coming_to.content_id, title: "Coming to the UK from Porpoise Spit"))
+      expect(Taxonomy::UpdateTaxon).to have_received(:call).with(taxon: having_attributes(content_id: child_taxon_trade_invest.content_id, title: "Trade and invest: Porpoise Spit"))
+      expect(Taxonomy::UpdateTaxon).to have_received(:call).with(taxon: having_attributes(content_id: child_taxon_birth_death.content_id, title: "Birth, death and marriage abroad in Porpoise Spit"))
+    end
+  end
+
+  describe "skip_tree_item?" do
+    let(:linked_item_living_in) do
```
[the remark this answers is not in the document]
ellohez: All done!

spec/lib/tasks/world_taxon_update_helper_spec.rb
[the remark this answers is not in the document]
mtaylorgds: We could just use regular variables inside the test.

spec/lib/tasks/world_taxon_update_helper_spec.rb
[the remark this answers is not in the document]
mtaylorgds: Did you just replace the two indicated lines (and leave the`allow(Taxonomy::ExpandedTaxonomy).to receive(:new).and_return(item_double)` in there? They all pass for me when I do that (and 11 fail if I also remove the last line).

spec/lib/tasks/world_taxon_update_helper_spec.rb
mtaylorgds: Similarly, these could just be variables in that test now.

lib/world_taxon_update_helper.rb
```diff
@@ -0,0 +1,185 @@
...
+  end
+
+  def skip_tree_item?(linked_item)
+    # Tree includes root (world/all) - skip that
+    if linked_item.content_id == WORLD_ROOT_CONTENT_ID
+      log_rake_progress("Skipping world root taxon")
+      return true
+    end
+
+    # Skip titles where the country name is already included at the end
```
mtaylorgds, line 174: This comment says it "skips titles where…", but the code is checking the internal name, not the title. We should update the text to be clearer here.

lib/world_taxon_update_helper.rb
```diff
@@ -0,0 +1,185 @@
...
+      suffix_index = title.index(": ")
+    elsif title.include?(" in ")
+      message = "removing - ...in COUNTRY_NAME from #{title}"
+      suffix_index = title.index(" in ")
+    end
+    log_rake_progress(message)
+
+    if suffix_index.nil?
+      log_rake_progress("No change to title: #{title}")
+      return ["No change to title: #{title}", ""]
```
mtaylorgds, line 159: It's not a good idea to mix return types (here it's returning an array of strings, but on line 164 it's returning a string scalar value).

I think returning the current title here would make the most sense (and would more closely match what the add method is doing).

lib/world_taxon_update_helper.rb
ellohez: Good point!

lib/world_taxon_update_helper.rb
ellohez: Well spotted! I think that was a hangover from when the method was returning message AND the new title.

mtaylorgds (commented):
Just a couple more comments.

mtaylorgds (approved):
LGTM

PR #1918, opened 2025-02-05
Add additional logging to catch edge cases

ellohez:
- If the newly created title is the same as the existing title as this indicates no change even though there should be
- If the internal name does not include the country name as the new title relies on this

Building on the existing work for this [Trello Card](https://trello.com/c/24XeROPv). 

### DIFF
--- a/lib/world_taxon_update_helper.rb
+++ b/lib/world_taxon_update_helper.rb
@@ -59,7 +59,10 @@ class WorldTaxonUpdateHelper
       title = linked_item.title
       new_title = create_title_removing_country_name(title)
 
-      next if title == new_title
+      if new_title == title
+        log_rake_progress("No change to title: #{title}")
+        next
+      end
 
       new_taxon = Taxonomy::BuildTaxon.call(content_id: linked_item.content_id)
       new_taxon.title = new_title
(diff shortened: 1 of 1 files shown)


lib/world_taxon_update_helper.rb
```diff
@@ -180,6 +183,11 @@ def skip_tree_item?(linked_item)
       return true
     end
 
+    if linked_item.internal_name.include?("(") != true
+      log_rake_progress("Error - no country name in internal name: #{internal_name}")
```
mtaylorgds, line 187: Just wondering whether we should be using `log_rake_error` here (or not say that it's an error)?

Either way, it may be worth including the fact that the item is being skipped in the log text.

lib/world_taxon_update_helper.rb
```diff
@@ -180,6 +183,11 @@ def skip_tree_item?(linked_item)
       return true
     end
 
+    if linked_item.internal_name.include?("(") != true
```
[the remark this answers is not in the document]
ellohez: I will log this as an error as I would prefer it to stand out in the logs rather then getting lost in all of the 'Skipping...' messages. You are right, at the moment it is a little confused.

lib/world_taxon_update_helper.rb
ellohez: Hmm, good point this is a bit confused at the moment. I would prefer that this didn't get lost in all of the 'Skipping...' messages so I think I will log it as an error.

mtaylorgds (approved):
LGTM

PR #2290, opened 2026-06-09
Fix brakeman warning

young-rob:
### Changes

We are getting a Brakeman warning for "Potentially unsafe model attribute in link_to href" in some of our views which use URL on models as hrefs. These URLs come from user input to columns in spreadsheets imported into content items. We generally trust these inputs, but it is worth adding a helper method to sanitise the urls for defence in depth.

The brakeman.ignore file is also regenerated.


AlexRyanShep (approved):
LGTM!

PR #2315, opened 2026-07-15
Update dependency to solve security alert with `js-yaml`

Tetrino:
Unlike other tools where we have been able to solve the dependency
by running a blanket update of `stylelint` and `standardx`, the
move to a much newer version of `yarn` means we have to be
more targeted in our approach.

This commit updates the problematic package directly with
`yarn up js-yaml -R`.


young-rob (approved):
LGTM 👍

